PR #2730, opened 2026-03-09
feat(project): migrate state and env vars when relocating project

gregpriday:
## Summary

When a project folder is moved or renamed, the project's status becomes `missing`. This PR completes the existing "Locate" flow so that clicking **Locate** for a missing project properly migrates all associated state to the new path — rather than just updating the path field and leaving orphaned state directories and broken env var keys.

Resolves #2688

## Changes Made

- **`ProjectEnvSecureStorage.migrateAllForProject(oldId, newId)`** — renames all env var keys from `{oldId}:{key}` to `{newId}:{key}` in electron-store, preserving all secure environment variables under the new project ID
- **`ProjectStore.relocateProject(projectId, newPath)`** — resolves the canonical git root of the selected path, computes the new hash-based project ID, copies the per-project state directory (`userData/projects/{oldHash}/`) to the new location, updates the project record (including the ID), migrates env var keys, then cleans up the old state directory; includes an active-project guard and store-state rollback on failure
- **`handleProjectLocate` IPC handler** — updated to call `relocateProject()` instead of a bare `updateProject()`, so all project state (terminal snapshots, settings, recipes, env vars) moves atomically with the project
- **Tests** — added `migrateAllForProject` unit tests covering key migration, isolation of other projects, no-op with empty source, prefix collision behaviour, and destination-key precedence; added smoke tests for the ID derivation helpers used by `relocateProject`

PR #2731, opened 2026-03-09
feat(persistence): migrate project registry from electron-store to SQLite

### LINKED ISSUE #2707: Projects stored in electron-store JSON — should be in SQLite for durability
## Summary

The registered project list is currently stored in `electron-store` — a single flat JSON file (`config.json`). This makes it vulnerable to corruption on crash, subject to lost-write races, and entangled with other app state in the same file. SQLite (already used for tasks and workflow runs) offers atomic writes, WAL journaling, and proper record-level isolation.

## Problem Statement

The project registry is the single most critical piece of user data in the app — it's what the user has explicitly configured. Yet it lives in the same JSON blob as ephemeral UI state like `sidebarWidth`, `focusMode`, and `mruList`. Every write to any `appState` key rewrites the whole file, meaning a crash mid-write can corrupt the project list too.

The app already has `canopy.db` (SQLite via `better-sqlite3` + Drizzle ORM) for tasks and workflow runs. A `projects` table in that database would give:

- **Atomic writes** — SQLite's WAL journal prevents partial writes
- **No cross-contamination** — project records are isolated from UI state writes
- **Crash safety** — WAL ensures committed records survive hard kills
- **Easier migration** — schema versioning is already in place

Currently `ProjectStore.ts` reads/writes via `store.get("projects.list")` and `store.set("projects.list")`. Every add, remove, or update of a project is a read-modify-write of the full array, which is a classic last-writer-wins race under concurrent access.

**Evidence:**
- [`electron/services/ProjectStore.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/ProjectStore.ts#L186-L201) — `addProject` and `removeProject` both do read-modify-write of the full list
- [`electron/store.ts`](https://github.com/canopyide/canopy/blob/develop/electron/store.ts#L89-L93) — `projects` is a top-level key in the shared electron-store schema, colocated with all other state
- [`electron/services/persistence/db.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/persistence/db.ts) — SQLite infrastructure with WAL mode is already initialised and used for tasks/workflows

## Desired Behavior

The project registry (list of registered project paths, IDs, names) is persisted in `canopy.db` with a dedicated `projects` table. Individual project records can be inserted, updated, and deleted without touching any other application state.

The `electron-store` `projects` key is removed. `ProjectStore.ts` reads from and writes to SQLite instead.

## Acceptance Criteria

- [ ] Registered projects survive a hard process kill (SIGKILL) between operations
- [ ] Adding, removing, or renaming a project does not require rewriting any unrelated state
- [ ] The project list is not affected by restoring or resetting workspace layout state
- [ ] Existing projects are migrated from `electron-store` to SQLite on first launch after upgrade
- [ ] `electron-store`'s `projects` key is removed from the schema
- [ ] `ProjectStore.ts` no longer imports from `../store`

## Edge Cases & Risks

- Migration from JSON to SQLite must be one-time and idempotent — handle already-migrated state gracefully
- The `projects.currentProjectId` field also needs a home — could be a `app_state` key-value table or kept in electron-store under a dedicated `session` namespace (not `projects`)
- Per-project state directories (`userData/projects/<id>/`) remain filesystem-based and are unaffected by this change
- The `CrashRecoveryService` session snapshot currently captures `store.get("projects")` — this becomes a no-op for the project list once it moves to SQLite (see also #2706)

## Dependencies

Can be worked independently, but resolves the underlying durability concern that makes #2704 and #2706 possible.

gregpriday:
## Summary

Moves the registered project list (and current project pointer) out of the electron-store JSON blob and into `canopy.db` — the same SQLite database already used for tasks and workflow runs. This gives project records atomic WAL-journaled writes and isolates them from UI state mutations that previously rewrote the entire config.json on every change.

Resolves #2707

## Changes Made

- Add `projects` and `app_state` tables to the SQLite schema (`schema.ts`) and `CREATE_TABLES_SQL` (`db.ts`), with `busy_timeout = 3000ms` for lock resilience
- Rewrite `ProjectStore.ts` to use Drizzle ORM (`getSharedDb()`) for all project CRUD — removes the `store` import entirely
- `setCurrentProject()` wraps demotion + currentProjectId upsert + activation in a single SQLite transaction for atomicity
- `getProjectByPath()` uses a targeted `WHERE path = ?` query instead of a full table scan
- Remove `projects` key from `StoreSchema` and electron-store defaults in `store.ts`
- Fix `worktree.ts` handler call-sites to use `projectStore.getCurrentProject()` / `getCurrentProjectId()` instead of `store.get("projects")`
- Fix `voiceInput.ts` `getProjectInfo()` to use `projectStore.getCurrentProject()`
- Fix `migration-003` to use `projectStore.getCurrentProjectId()` / `getProjectById()`
- Add `ProjectStore` mock to voiceInput paragraph test; add coverage for project info flowing into correction settings

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -31,7 +31,10 @@ fixPath();
 
 // In development, use a separate userData directory so the dev instance
 // doesn't conflict with the production app's single-instance lock or storage.
-if (!app.isPackaged) {
+// Skip when --user-data-dir is explicitly set (e.g. E2E tests) so that
+// each test run gets its own isolated data directory.
+const hasExplicitUserDataDir = process.argv.some((a) => a.startsWith("--user-data-dir"));
+if (!app.isPackaged && !hasExplicitUserDataDir) {
   app.setPath("userData", path.join(app.getPath("appData"), `${app.name}-dev`));
 }
 


gregpriday:
**Review status: Fixed and ready (pending CI)**

**Root cause of E2E failures:**
The `app.setPath("userData", ...)` override in `main.ts` (for dev mode isolation) was clobbering the `--user-data-dir` flag passed by E2E tests. This caused all E2E test runs to share a single SQLite database at the dev userData path. When `core-advanced.spec.ts` ran first (alphabetically) and created the "Advanced Test" project, that project persisted into `core-shell-settings.spec.ts`, which expected a fresh welcome screen with no projects.

The electron-store wasn't affected because it initializes at ESM import time (before `app.setPath` runs), so it correctly used the `--user-data-dir` path. But `getSharedDb()` is lazy — it resolves `app.getPath("userData")` on first call, after `setPath` has already overridden it.

**Fix applied:**
- Skip `app.setPath("userData")` override when `--user-data-dir` is explicitly passed via command line args

**Cross-PR note:**
- PR #2730 adds `relocateProject()` to `ProjectStore.ts` using the old `store.set("projects.list")` API. Must merge #2730 before #2731, then update `relocateProject()` to use the new SQLite-backed methods.

- CI: Pushed fix, awaiting results
- Conflicts: None

PR #2837, opened 2026-03-11
feat(voice): inject dynamic project context into Deepgram keyterms

### LINKED ISSUE #2820: Inject dynamic project context into Deepgram keyterms on session start
## Summary

Deepgram's `keyterm` prompting biases the acoustic model toward specific vocabulary at inference time — claiming up to 90% improvement in recall for injected terms. Currently, Canopy only injects a static user-managed dictionary (capped at 100 terms). The IDE already has richer context available — active file names, function/variable names, terminal output, git branch — that could dramatically improve recognition of project-specific terminology before it ever reaches the LLM correction step.

## Problem Statement

Transcription errors on developer terminology (function names, variable names, branch names, package names) are expensive to fix after the fact. The LLM correction layer has to guess that "off module" meant `auth` module because "auth" wasn't acoustically reinforced at the STT layer.

Deepgram supports up to 500 tokens of keyterm injection per session, and these are applied at inference time with no latency penalty. The current implementation hardcodes a static user dictionary with a 100-term cap:

```typescript
const keyterms = settings.customDictionary.slice(0, 100);
```

Any terms beyond index 100 are silently dropped, and the list never reflects what's actually on screen at recording time — so a new variable you just wrote won't be recognised until a user manually adds it to their dictionary.

## Desired Behavior

When a voice recording session starts, the system should assemble a richer set of keyterms that reflects the current working context of the IDE:

- Names from recent terminal output in the active panel (command names, paths, identifiers)
- The current git branch name
- Project name and directory-level identifiers
- The static user custom dictionary (as it exists today)

Terms should be deduplicated, prioritised by relevance to the current moment, and trimmed to fit within Deepgram's token budget. The goal is that anything the developer just typed, ran, or is looking at gets acoustically reinforced for the duration of the recording session — without the user needing to manage a manual dictionary.

## Context

The session start flow lives in:

- [`electron/ipc/handlers/voiceInput.ts`](https://github.com/canopyide/canopy/blob/develop/electron/ipc/handlers/voiceInput.ts) — Handles `voice-input:start`; this is where transcription settings are snapshotted and the Deepgram connection is initialised with the current keyterm list
- [`electron/services/VoiceTranscriptionService.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/VoiceTranscriptionService.ts) — Receives the keyterm array and passes it to the Deepgram connection config

The `voice-input:start` handler already receives project context (`sessionProjectInfo`). Terminal output and git state are available via existing services in the main process.

## Acceptance Criteria

- [ ] On session start, keyterms are assembled from at least: the user's static custom dictionary, the current git branch name, and the active project name
- [ ] The assembled keyterm list respects Deepgram's token budget (500 tokens) and does not silently truncate high-priority terms
- [ ] Terms derived from live IDE context (branch name, recent terminal identifiers) are included without requiring any user action
- [ ] The static user custom dictionary continues to work as before; existing behaviour is not regressed
- [ ] Common English words and generic syntax keywords (e.g., "function", "class", "const") are not injected as keyterms, as Deepgram's documentation indicates this degrades overall accuracy

## Edge Cases & Risks

- Terminal output may be large and noisy; extraction logic needs to identify meaningful identifiers rather than pulling in every word from shell output
- Keyterm injection should not add observable latency to session start; assembly should be fast or happen in parallel with Deepgram WebSocket setup
- If context gathering fails for any reason, the session should proceed with only the static dictionary — no silent errors

gregpriday:
## Summary

- Assembles a richer set of Deepgram keyterms at voice session start by pulling from live IDE context (git branch name, project name, recent terminal identifiers) alongside the existing static user dictionary
- Implements a priority-based token budget system that respects Deepgram's 500-token limit, ensuring high-value terms (user dictionary, branch names) aren't silently truncated
- Filters out common English words and generic syntax keywords that Deepgram's docs say degrade overall accuracy

Resolves #2820

## Changes

- **`electron/services/voiceContextKeyterms.ts`** (new): Core module with `assembleKeyterms()` that gathers context from multiple sources, deduplicates, filters stop words, and trims to the token budget. Extracts identifiers from terminal output using regex-based tokenization.
- **`electron/ipc/handlers/voiceInput.ts`**: Updated `voice-input:start` handler to call `assembleKeyterms()` instead of hardcoding `customDictionary.slice(0, 100)`. Falls back to the static dictionary if context assembly fails.
- **`electron/services/VoiceTranscriptionService.ts`**: Changed `keyterms` parameter type from `readonly string[]` to `string[]` for compatibility.
- **`shared/types/ipc/maps.ts`**: Added IPC channel types for `git:list-branches` used by the keyterm assembler.
- **`electron/services/__tests__/voiceContextKeyterms.test.ts`** (new): 12 test cases covering token budgeting, stop word filtering, deduplication, terminal identifier extraction, graceful fallback on errors, and the assembly timeout guard.

## Testing

All 12 unit tests pass. TypeScript typecheck, ESLint, and Prettier all clean. The implementation includes a 2-second timeout on context assembly so session start latency is bounded, with fallback to the static dictionary on any failure.

### DIFF
--- a/electron/services/__tests__/voiceContextKeyterms.test.ts
+++ b/electron/services/__tests__/voiceContextKeyterms.test.ts
@@ -7,7 +7,7 @@ import {
   extractTerminalIdentifiers,
 } from "../voiceContextKeyterms.js";
 
-let gitListBranchesMock = vi.fn().mockResolvedValue([
+const gitListBranchesMock = vi.fn().mockResolvedValue([
   { name: "feature/auth-login-service", current: true, commit: "abc123" },
   { name: "main", current: false, commit: "def456" },
 ]);


gregpriday:
**Review status: Ready**

**Cross-PR note:** This PR should merge **before** PR #2834 (word-level confidence). Both modify `voiceInput.ts` — merging this first minimizes conflict resolution needed for #2834.

- CI: All checks passing
- Conflicts: None (MERGEABLE)

PR #3175, opened 2026-03-14
fix(agent): persist and restore launch flags on session resume

### LINKED ISSUE #3168: Resume command omits original agent flags after crash recovery
## Summary

When Canopy restores an agent terminal session after a crash, the resume command is built using only the agent's resume args (e.g. `--resume <sessionId>`). Runtime flags that were active when the session originally launched — such as `--dangerously-skip-permissions` for Claude — are silently dropped. The resumed process starts without them, breaking the behavior the user had configured.

## Problem Statement

Users who have enabled dangerous mode (skip permissions), custom CLI flags, or other process-level options for an agent expect those same flags to be present when their session is restored after a crash. Without them, the resumed agent behaves differently from the original session — Claude Code, for example, will re-prompt for every tool permission instead of running autonomously, defeating the purpose of resuming.

This affects anyone using:
- `--dangerously-skip-permissions` (Claude Code)
- `--yolo` (Gemini)
- `--dangerously-bypass-approvals-and-sandbox` (Codex)
- Any custom flags set in agent settings

There is no workaround short of manually relaunching the agent with the correct flags.

## Desired Behavior

When building the resume command after a crash, Canopy should reconstruct the full original command — that is, include the same flags (dangerous mode args, custom flags, and any other process-level options) that were present when the session was first launched, alongside the resume argument.

The resume command should look like:

```
claude --dangerously-skip-permissions --resume <sessionId>
```

rather than the current:

```
claude --resume <sessionId>
```

The flags should reflect the original launch configuration. Even if agent settings have since changed, the session was started under specific conditions and must be resumed under those same conditions to function correctly.

## Context

The resume command is built by `buildResumeCommand` in [`shared/types/agentSettings.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/agentSettings.ts#L287-L301). Its current implementation intentionally strips all flags with the comment "the resumed session already has its configuration embedded." This is accurate for conversational state (model choice, conversation history) but incorrect for process-level runtime flags, which must be supplied each time the CLI process is invoked.

The two call sites that invoke `buildResumeCommand` are:

- [`src/utils/stateHydration.ts#L721-L725`](https://github.com/canopyide/canopy/blob/develop/src/utils/stateHydration.ts#L721-L725) — hydration on cold app start after crash
- [`src/store/slices/terminalRegistry/index.ts#L1497-L1504`](https://github.com/canopyide/canopy/blob/develop/src/store/slices/terminalRegistry/index.ts#L1497-L1504) — reconnection during a live session

Both paths fall back to `generateAgentCommand` (which correctly includes all flags from current agent settings) only when no session ID is present. When a session ID exists, `buildResumeCommand` is used and flags are dropped.

The agent settings that produce these flags live in [`shared/types/agentSettings.ts#L128-L166`](https://github.com/canopyide/canopy/blob/develop/shared/types/agentSettings.ts#L128-L166) (`generateAgentFlags`), and the per-agent defaults (e.g. `--dangerously-skip-permissions` for Claude) are defined at [`shared/types/agentSettings.ts#L95-L99`](https://github.com/canopyide/canopy/blob/develop/shared/types/agentSettings.ts#L95-L99).

## Acceptance Criteria

- [ ] When an agent terminal session is resumed after a crash, the resume command includes all original runtime flags (dangerous mode, custom flags) that were active at launch time — not just the `--resume <sessionId>` argument
- [ ] For Claude Code with `--dangerously-skip-permissions` enabled, the resumed command is `claude --dangerously-skip-permissions --resume <sessionId>` (or equivalent for other agents)
- [ ] If flags have changed in agent settings between crash and restart, the original launch flags take precedence so the resumed session behaves identically to the original
- [ ] Other session types (non-agent terminals, terminals without a session ID) are unaffected

## Edge Cases & Risks

- The original launch flags are not currently persisted anywhere alongside the session ID. A mechanism to capture and store the flags used at launch time (alongside the `agentSessionId`) may be needed so they are available at restore time.
- Care is needed to ensure flags are inserted before the resume argument in the correct position, since some CLIs are sensitive to argument order.
- Inline mode flag (`--no-alt-screen` for Codex) is also a process-level flag that may need the same treatment.

gregpriday:
## Summary

When an agent session was resumed after a crash, the resume command was built without the original runtime flags (e.g. `--dangerously-skip-permissions` for Claude). This meant resumed sessions behaved differently from the original launch — Claude would re-prompt for every permission instead of running autonomously.

The fix persists `agentLaunchFlags` at launch time alongside the session ID and threads them through both resume paths: cold-start hydration and live reconnection.

Resolves #3168

## Changes

- `shared/types/domain.ts` — added `agentLaunchFlags` field to `PtyPanelData`
- `shared/types/agentSettings.ts` — updated `buildResumeCommand` to accept and prepend persisted launch flags; added `AgentLaunchFlags` export; extended tests to cover flag restoration
- `shared/types/ipc/terminal.ts` — added `agentLaunchFlags` to `TerminalState` so it survives the IPC boundary
- `src/hooks/useAgentLauncher.ts` — captures flags at launch time and stores them on the panel
- `src/store/slices/terminalRegistry/index.ts` — passes stored flags into `buildResumeCommand` during live reconnection
- `src/utils/stateHydration.ts` — passes stored flags into `buildResumeCommand` during cold-start hydration
- `src/store/persistence/terminalPersistence.ts` — includes `agentLaunchFlags` in persisted state

## Testing

- Unit tests in `shared/types/__tests__/agentSettings.test.ts` cover: resume with no flags, resume with dangerous mode flags, resume with custom flags, and resume with both combined
- Existing tests pass; no regressions in the `npm run check` pipeline

### DIFF
--- a/shared/types/ipc/terminal.ts
+++ b/shared/types/ipc/terminal.ts
@@ -90,6 +90,8 @@ export interface TerminalState {
   exitBehavior?: import("../domain.js").PanelExitBehavior;
   /** Captured agent session ID from graceful shutdown */
   agentSessionId?: string;
+  /** Process-level flags captured at launch time, persisted for session resume */
+  agentLaunchFlags?: string[];
 }
 
 /** Terminal data payload for IPC */


gregpriday:

**Fixes applied:**
- Removed unused `TRASH_UNDO_TOAST_DURATION_MS` export from `shared/config/trash.ts` — dead constant re-introduced by develop HEAD commit 9461e14a, not imported anywhere, breaking the `trashToastRemoval.test.ts` regression test

- CI: Fix pushed, waiting for re-run
- Conflicts: None

PR #3195, opened 2026-03-14
fix(agent): allow user input to recover agent from failed state

### LINKED ISSUE #3190: Agent stuck in failed state after error — user input cannot recover it
## Summary

Once an agent terminal transitions to the `failed` state, it is locked there permanently. The state machine does not permit any transition out of `failed` — including when the user types a new prompt and submits it. This means a user who receives an error (authentication failure, network timeout, API error) cannot see their agent return to `working` state even after successfully retrying, because the state machine rejects the transition.

## Current Behavior

The `AgentStateMachine` defines `failed` as a terminal state with only one valid transition: `failed → failed` (to update the error message). When a user types a new prompt after an error and submits it, the `input` event is fired — but the state machine rejects it because `failed → working` is not in the valid transitions table. The agent remains visually in `failed` state even as it begins processing the new request.

The same asymmetry does not exist for `completed` state, which correctly allows `input` events to restart the agent into `working`.

## Expected Behavior

User input (submitting a prompt) should be able to recover an agent from `failed` state, transitioning it to `working`. This matches the behavior of `waiting → working` and `completed → working` transitions that already exist. The `failed` state should indicate "the last operation failed" not "this agent can never work again."

## Root Cause Analysis

The valid transitions are defined in the state machine and `failed` is intentionally treated as a sink state. This was likely a conservative design choice to prevent heuristic signals (pattern matches, output activity) from incorrectly clearing an error state. However, the `input` trigger — which has 1.0 confidence and represents an explicit user action — should be trusted to escape the `failed` state just as it escapes `waiting` and `completed`.

**Evidence:**
- [`electron/services/AgentStateMachine.ts#L13-L25`](https://github.com/canopyide/canopy/blob/develop/electron/services/AgentStateMachine.ts#L13-L25) — valid state transitions table; `failed` has no outgoing transitions to `working`
- [`electron/services/pty/AgentStateService.ts#L54-L91`](https://github.com/canopyide/canopy/blob/develop/electron/services/pty/AgentStateService.ts#L54-L91) — confidence inference; `input` events receive confidence 1.0, making them the highest-trust transition trigger

## Acceptance Criteria

- [ ] Submitting a non-empty prompt while the agent is in `failed` state transitions it to `working`
- [ ] The `failed` error state is cleared when the agent transitions out of `failed`
- [ ] Heuristic signals (pattern matches, output activity, debounce timeouts) do NOT escape `failed` state — only explicit user input does
- [ ] The existing `failed → failed` transition (error message update) still works
- [ ] No regression in how `failed` state is reached or displayed

## Edge Cases & Risks

- Only `input` events (explicit user action, confidence 1.0) should be able to escape `failed`. The fix must not allow `busy`, `prompt`, or `completion` events to do so — those are heuristic signals that could be spuriously triggered by error output.
- If the agent errors again immediately after the user retries, it should return to `failed` correctly.

gregpriday:
## Summary

- `failed` was a terminal sink state in `AgentStateMachine` with no valid outgoing transitions, so user prompts submitted after an error couldn't restart the agent
- Added `failed → working` transition gated exclusively on `input` events (confidence 1.0), matching the existing `completed → working` and `waiting → working` patterns
- Heuristic signals (`busy`, `prompt`, `completion`) cannot escape `failed` state, preventing spurious error clearance from noisy output

Resolves #3190

## Changes

- `electron/services/AgentStateMachine.ts` — added `input` event transition from `failed` to `working`
- `electron/services/pty/AgentStateService.ts` — updated state handling to permit the new transition
- `src/workers/WorkerAgentStateService.ts` — mirrored the same transition for the renderer-side worker
- `electron/services/__tests__/AgentStateMachine.test.ts` — tests for `failed → working` via input, and confirming heuristic signals stay blocked
- `electron/services/pty/__tests__/AgentStateService.test.ts` — full coverage of the recovery path and edge cases

## Testing

All existing tests pass. New tests cover: input recovers from failed, heuristic events do not, re-entering failed after recovery works correctly.

### DIFF
--- a/electron/services/pty/__tests__/AgentStateService.test.ts
+++ b/electron/services/pty/__tests__/AgentStateService.test.ts
@@ -112,24 +112,39 @@ describe("AgentStateService", () => {
     expect(terminal.agentState).toBe("failed");
   });
 
-  it("updates error message on failed → failed transition", () => {
+  it("updates error message on failed → failed without emitting state-changed", () => {
     const service = new AgentStateService();
     const terminal = createTerminal({ agentState: "failed", error: "old error" });
+    const stateChanges: unknown[] = [];
 
-    service.updateAgentState(terminal, { type: "error", error: "new error" });
+    events.on("agent:state-changed", (payload) => {
+      stateChanges.push(payload);
+    });
 
+    const changed = service.updateAgentState(terminal, { type: "error", error: "new error" });
+
+    expect(changed).toBe(false);
     expect(terminal.agentState).toBe("failed");
     expect(terminal.error).toBe("new error");
+    expect(stateChanges).toHaveLength(0);
   });
 
-  it("handleActivityState with trigger input recovers from failed", () => {
+  it("handleActivityState with trigger input recovers from failed with input trigger", () => {
     const service = new AgentStateService();
     const terminal = createTerminal({ agentState: "failed", error: "some error" });
+    const stateChanges: Array<{ trigger: string; state: string }> = [];
+
+    events.on("agent:state-changed", (payload) => {
+      stateChanges.push({ trigger: payload.trigger, state: payload.state });
+    });
 
     service.handleActivityState(terminal, "busy", { trigger: "input" });
 
     expect(terminal.agentState).toBe("working");
     expect(terminal.error).toBeUndefined();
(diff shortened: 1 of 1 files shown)


gregpriday:
**Review status: Ready**

PR #3213, opened 2026-03-14
feat(terminal): use WebGL renderer for focused terminal

### LINKED ISSUE #3156: Focused terminal uses CPU-heavy DOM renderer instead of WebGL
## Summary

All terminal panels render exclusively through the xterm.js DOM renderer, which creates thousands of `<span>` and `<div>` elements per frame and is CPU-bound. Since xterm.js 6.0 removed the canvas renderer, WebGL is the only hardware-accelerated rendering path available. For the active/focused terminal — the one the user is watching during a build, test run, or agent task — switching to WebGL would substantially reduce the CPU cost of high-throughput output rendering.

## Problem Statement

The DOM renderer works reliably for any number of terminals and avoids the Chromium-enforced limit of 16 active WebGL contexts per renderer process. This was a reasonable default choice. However, the trade-off is significant for the terminal the user is actually watching:

- The DOM renderer allocates one DOM node per terminal cell for rendered rows. For a 220-column × 50-row terminal, that's 11,000+ nodes that must be diffed and repainted on each output frame.
- During high-throughput output (build logs, test runners, `cat` of a large file), the focused terminal drives repeated full-viewport reflows that push CPU usage noticeably higher than necessary.
- The WebGL renderer (`@xterm/addon-webgl`) renders the entire terminal viewport as a single `<canvas>` element, offloading character rasterisation and compositing to the GPU. Benchmarks show WebGL is up to 900% faster than the canvas renderer (now removed in 6.0), which itself was 5–45× faster than the DOM renderer.
- With only one WebGL context needed (the focused terminal), the 16-context per-process ceiling is not approached, and the existing DOM renderer continues serving all other panels.

## Rendering Options in xterm.js 6.0

xterm.js 6.0 removed `@xterm/addon-canvas` (the 2D Canvas renderer). The only two rendering paths available are:

| Renderer | How it works | Pros | Cons |
|----------|-------------|------|------|
| **DOM** (built-in) | One DOM node per cell, diffed on each frame | Always available, no GPU required, unlimited instances | CPU-bound, layout thrashing on high throughput |
| **WebGL** (`@xterm/addon-webgl`) | Single `<canvas>`, GPU-composited draw calls | Orders of magnitude faster for CPU path | Limited to 16 contexts per Chromium renderer process, requires GPU |

There is no middle-ground canvas fallback. The architecture must use DOM as the universal fallback and WebGL as the performance tier.

## Desired Behavior

The terminal with active keyboard focus and/or the terminal currently in the foreground renderer tier should use the WebGL addon for hardware-accelerated rendering. All non-focused panels continue using the DOM renderer.

When the focused terminal changes, the WebGL context is detached from the old terminal and attached to the new one. If WebGL context creation fails for any reason (GPU process crash, context loss event from `onContextLoss`, or the `CANOPY_DISABLE_WEBGL` flag), the terminal transparently falls back to the DOM renderer without any user-visible disruption.

### WebGL Lifecycle

1. When a terminal gains the `FOCUSED` tier (via `TerminalRendererPolicy`), attempt to attach `@xterm/addon-webgl`.
2. Before attaching to the new terminal, fully `.dispose()` the WebGL addon on the previously focused terminal. The old terminal reverts to the DOM renderer automatically.
3. Wrap `terminal.loadAddon(webglAddon)` in a `try/catch`. If it throws (no GPU, software rendering, etc.), silently remain on the DOM renderer.
4. Subscribe to the addon's `onContextLoss` event. On context loss, `.dispose()` the addon immediately — the terminal falls back to DOM. WebGL can be re-attempted on the next focus change.
5. Debounce focus changes to prevent rapid attach/dispose cycles (the existing `TIER_DOWNGRADE_HYSTERESIS_MS` in `TerminalRendererPolicy` already provides this for downgrades; upgrades to `FOCUSED` are immediate but should be debounced for WebGL attachment specifically).

### `CANOPY_DISABLE_WEBGL` Environment Flag

Add a `CANOPY_DISABLE_WEBGL` environment variable that completely disables WebGL addon loading when set to `"1"`. This serves three purposes:

1. **E2E testing** — CI runners pass `--disable-gpu` and have no real GPU. While the `try/catch` fallback would handle this, explicitly disabling WebGL avoids unnecessary addon loading overhead and makes test behavior deterministic.
2. **Debugging** — Developers can isolate whether a rendering issue is WebGL-related by running with this flag.
3. **User escape hatch** — Users on problematic GPU drivers can force DOM-only mode.

The flag should be checked in `TerminalAddonManager` before any WebGL attachment attempt.

### E2E Test Integration

The e2e test suite reads terminal content via `.xterm-rows` `innerText()` (see `e2e/helpers/terminal.ts`). While xterm.js keeps `.xterm-rows` populated for accessibility even with WebGL active, explicitly disabling WebGL in tests ensures deterministic behavior and avoids any risk of test flakiness.

In `e2e/helpers/launch.ts`, add to the CI launch environment:

```typescript
CANOPY_DISABLE_WEBGL: "1",
```

This should be set for **all** e2e test projects (core and online), not just online tests, since any test that reads terminal content or interacts with terminals benefits from deterministic DOM rendering.

## Context

The rendering tier system and focus tracking already exist in the codebase and provide the signals needed to know which terminal is "active":

- [`src/services/terminal/TerminalRendererPolicy.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalRendererPolicy.ts) — tier management (`FOCUSED` → `VISIBLE` → `BACKGROUND`) with hysteresis
- [`src/services/terminal/TerminalAddonManager.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalAddonManager.ts) — where addon loading happens; line 23 explains the explicit decision to avoid WebGL context limits

The `@xterm/addon-webgl` package is already a transitive dependency of `@xterm/xterm` 6.0 (confirmed in `package.json`). The WebGL addon exposes an `onContextLoss` event specifically designed for graceful degradation.

## Acceptance Criteria

- [ ] The focused/active terminal uses `@xterm/addon-webgl` for rendering; all other terminals continue using the DOM renderer
- [ ] When focus shifts to a different terminal, the WebGL context transfers to that terminal (or a new one is allocated) within the existing tier-change hysteresis window
- [ ] A `webglcontextlost` event on the focused terminal's canvas causes immediate, silent fallback to the DOM renderer for that terminal; focus can be regained and WebGL re-attempted on the next focus change
- [ ] No visual artefacts (flash, blank frame, cursor jump) occur during WebGL attach or detach transitions
- [ ] The total active WebGL context count remains 1 at steady state
- [ ] Setting `CANOPY_DISABLE_WEBGL=1` completely prevents WebGL addon loading and all terminals use the DOM renderer
- [ ] All e2e tests (core and online) run with `CANOPY_DISABLE_WEBGL=1` set in the launch environment
- [ ] The existing test suite and E2E coverage pass without new failures

## Edge Cases & Risks

- **Rapid focus changes** (tab switching, keyboard shortcut cycling) must not create multiple leaked WebGL contexts. The old addon must be fully disposed before attaching a new one. Focus-change debouncing (~50-100ms) should gate WebGL attachment.
- **GPU failures** — Some GPU configurations (software-only rendering in VMs, certain Linux setups) will fail WebGL context creation silently. The `try/catch` on `loadAddon` and the `onContextLoss` handler are both required for correct fallback.
- **Disposal cleanup** — Calling `.dispose()` on the WebGL addon is sufficient to remove the injected canvas and free GPU resources (confirmed for xterm.js 6.0). No manual DOM manipulation is needed.
- **Known xterm.js bugs** — A GPU memory leak on disposal was fixed in v5.0 (Issue #3889), and a race condition during terminal disposal involving the WebGL addon was reported in Oct 2024 (Issue #5181). Ensure disposal is sequential, not concurrent.
- **Shader compilation latency** — The first WebGL attachment after app launch compiles shaders, which may cause a brief delay. Subsequent attach/dispose cycles reuse compiled shaders if the GPU process hasn't been recycled. This is acceptable.

gregpriday:
## Summary

- Adds `TerminalWebGLManager` to manage a single shared WebGL addon instance that follows focus, attaching to the focused terminal and disposing cleanly when focus shifts
- Integrates with `TerminalRendererPolicy`'s tier system so WebGL attaches on `FOCUSED` tier promotion and detaches on demotion, with debouncing to prevent rapid churn
- All non-focused terminals continue using the DOM renderer; total active WebGL context count stays at 1

Resolves #3156

## Changes

- `src/services/terminal/TerminalWebGLManager.ts` — new class managing WebGL addon lifecycle with context-loss handling and `CANOPY_DISABLE_WEBGL` escape hatch
- `src/services/terminal/TerminalRendererPolicy.ts` — wires tier changes into `TerminalWebGLManager`
- `src/services/terminal/TerminalAddonManager.ts` — minor cleanup; WebGL loading moved to `TerminalWebGLManager`
- `src/services/terminal/TerminalInstanceService.ts` — initialises and tears down `TerminalWebGLManager` with the service lifecycle
- `e2e/helpers/launch.ts` — sets `CANOPY_DISABLE_WEBGL=1` for all e2e test projects so tests use deterministic DOM rendering
- `package.json` — adds `@xterm/addon-webgl` as an explicit dependency (was previously only transitive)
- Tests for `TerminalWebGLManager` covering attach, detach, context loss, disable flag, and debounce behaviour

## Testing

Unit tests pass (`TerminalWebGLManager.test.ts`, updated mocks across all `TerminalInstanceService` test files). E2E tests will run with WebGL disabled via the launch environment flag, keeping terminal content readable through `.xterm-rows`.

### DIFF
--- a/src/services/terminal/TerminalWebGLManager.ts
+++ b/src/services/terminal/TerminalWebGLManager.ts
@@ -32,8 +32,16 @@ export class TerminalWebGLManager {
     } catch {
       // WebGL unavailable (CI, older GPU, context limit) — stay on DOM renderer.
       // Clean up partially-constructed addon to prevent leaks.
-      try { clDisposable?.dispose(); } catch { /* ignore */ }
-      try { addon?.dispose(); } catch { /* ignore */ }
+      try {
+        clDisposable?.dispose();
+      } catch {
+        // ignore
+      }
+      try {
+        addon?.dispose();
+      } catch {
+        // ignore
+      }
     }
   }
 


gregpriday:
**Review status: Ready**

PR #3237, opened 2026-03-14
feat(monitoring): add proactive per-process memory monitoring

### LINKED ISSUE #3159: Add proactive per-process memory trend monitoring to the main process
## Summary

Canopy has no ongoing visibility into per-process memory trends at runtime. Memory data is only captured as a one-time snapshot when a user manually triggers a diagnostics report. For an app that runs for hours or days managing dozens of terminal processes, this means memory growth goes unobserved until the app crashes or the user notices slowdown.

## Problem Statement

Canopy's three active processes — Main, Renderer, and pty-host — can each develop memory growth patterns for different reasons (PTY buffer accumulation, V8 heap fragmentation, WebContents leaks, Zustand store growth). Without periodic monitoring:

- **Regressions ship silently.** A change that introduces a slow memory leak cannot be detected until users report slowdowns days later.
- **Crashes are unexplained.** The V8 heap cap in Electron 40 (Node 22 with pointer compression) is approximately 4 GB per process. When the renderer or pty-host approaches that limit, there is no warning event or log entry — just a hard crash.
- **Long-running soak tests are the only signal.** The existing `scripts/perf/scenarios/soak.ts` benchmarks simulate memory growth in isolation, but they run against a mocked workload, not a live session.

Currently, `app.getAppMetrics()` is called exactly once — inside `DiagnosticsCollector.ts` — on demand when the user clicks "Copy Diagnostics":

- [`electron/services/DiagnosticsCollector.ts#L165-L168`](https://github.com/canopyide/canopy/blob/develop/electron/services/DiagnosticsCollector.ts#L165-L168) — single ad-hoc call, no trend data, no thresholds

## Desired Behavior

The main process periodically samples process memory from all child processes and logs warnings when growth exceeds defined thresholds. In development mode (and optionally in production via a flag), when a process exceeds a critical threshold, a V8 heap snapshot is written to disk automatically so the cause can be investigated with Chrome DevTools.

Concretely:
- `app.getAppMetrics()` is polled on a low-frequency interval (e.g. 30 seconds) — well below the frequency that causes observer-effect overhead
- Each sample is logged at `debug` level so it doesn't pollute normal output; threshold crossings are logged at `warn`
- Reasonable thresholds for an IDE: Renderer > 1.5 GB working set triggers `warn`; Main > 300 MB triggers `warn`; pty-host > 500 MB triggers `warn`
- At the `warn` threshold in dev mode, a V8 heap snapshot is written to the app's log directory (same directory used by `assistantLogger`) so it can be opened in Chrome DevTools Memory tab
- The polling respects app lifecycle — it does not start before `app.whenReady()` and is torn down cleanly on `app.on('before-quit')`

## Context

The infrastructure for logging and app lifecycle management is already in place. The existing perf harness (`scripts/perf/`) measures memory growth in benchmarks but has no visibility into what the live app does over a real session.

- [`electron/utils/assistantLogger.ts`](https://github.com/canopyide/canopy/blob/develop/electron/utils/assistantLogger.ts) — shows the pattern for dev-only log files and the app data directory path
- [`electron/services/DiagnosticsCollector.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/DiagnosticsCollector.ts) — existing one-shot metrics collection; the `appMetrics` field demonstrates the correct API call
- [`scripts/perf/scenarios/soak.ts`](https://github.com/canopyide/canopy/blob/develop/scripts/perf/scenarios/soak.ts) — existing soak scenarios already track `memoryGrowthMb` and `memoryGrowthPct`, establishing the right metrics vocabulary

## Acceptance Criteria

- [ ] A monitoring service in the main process periodically samples `app.getAppMetrics()` (per-process working set) and logs the data
- [ ] When the renderer process working set exceeds a configurable warn threshold, a `warn`-level log entry is written with the process type, PID, and current memory figure
- [ ] In development mode, when a monitored process crosses the critical threshold, a V8 heap snapshot file is written to the Canopy log directory
- [ ] The monitor introduces no measurable startup latency and no interval timer that blocks the main-process event loop
- [ ] The monitor tears down cleanly on `before-quit` (no dangling setInterval after app exit)

## Edge Cases & Risks

- `app.getAppMetrics()` returns metrics for all Electron helper processes, not just the app's intentional processes. The monitoring code must filter to meaningful process types (`Browser`, `Renderer`, `Utility`) and not alert on GPU or crashpad processes that have different memory profiles.
- V8 heap snapshots can be 100–500 MB and take 5–15 seconds to write. In production this could cause a noticeable freeze. Heap snapshots should be limited to development mode unless explicitly opted into.
- Multiple threshold crossings in a short window (during startup or after a large context injection) should not trigger multiple snapshots. A per-process cooldown between snapshot writes prevents log directory bloat.

gregpriday:
## Summary

- Adds `ProcessMemoryMonitor`, a new main-process service that polls `app.getAppMetrics()` every 30 seconds and logs per-process working set at debug level
- Logs a `warn`-level entry when any monitored process (Browser, Renderer, Utility) crosses a configurable threshold (Renderer > 1.5 GB, Main/Browser > 300 MB, pty-host/Utility > 500 MB)
- In development mode, writes a V8 heap snapshot to the Canopy log directory when a critical threshold is crossed, with a per-process cooldown to prevent snapshot spam

Resolves #3159

## Changes

- `electron/services/ProcessMemoryMonitor.ts` — new service; starts after `app.whenReady()`, tears down cleanly on `before-quit`
- `electron/main.ts` — wires up `ProcessMemoryMonitor` into the app lifecycle
- `electron/services/__tests__/ProcessMemoryMonitor.test.ts` — 259-line test suite covering normal polling, threshold crossings, snapshot cooldown, and teardown

## Testing

Full test suite passes (`npm run check`). The test file covers the main edge cases from the issue: GPU/crashpad process filtering, cooldown between snapshots, and clean interval teardown on quit. No measurable startup latency since the first poll is deferred to the interval.

### DIFF
--- a/electron/services/ProcessMemoryMonitor.ts
+++ b/electron/services/ProcessMemoryMonitor.ts
@@ -21,40 +21,44 @@ export function startAppMetricsMonitor(): () => void {
   const snapshotCooldowns = new Map<number, number>();
 
   const timer = setInterval(() => {
-    const metrics = app.getAppMetrics();
+    try {
+      const metrics = app.getAppMetrics();
 
-    for (const proc of metrics) {
-      if (!MONITORED_TYPES.has(proc.type)) continue;
+      for (const proc of metrics) {
+        if (!MONITORED_TYPES.has(proc.type)) continue;
 
-      const mb = proc.memory.workingSetSize / 1024;
-      logDebug("process-memory-sample", { pid: proc.pid, type: proc.type, mb: Math.round(mb) });
+        const mb = proc.memory.workingSetSize / 1024;
+        logDebug("process-memory-sample", { pid: proc.pid, type: proc.type, mb: Math.round(mb) });
 
-      const threshold = WARN_THRESHOLDS_MB[proc.type];
-      if (threshold !== undefined && mb > threshold) {
-        logWarn("process-memory-threshold-exceeded", {
-          pid: proc.pid,
-          type: proc.type,
-          mb: Math.round(mb),
-          thresholdMb: threshold,
-        });
-      }
+        const threshold = WARN_THRESHOLDS_MB[proc.type];
+        if (threshold !== undefined && mb > threshold) {
+          logWarn("process-memory-threshold-exceeded", {
+            pid: proc.pid,
+            type: proc.type,
+            mb: Math.round(mb),
+            thresholdMb: threshold,
+          });
+        }
 
-      if (proc.type === "Browser" && mb > SNAPSHOT_THRESHOLD_MB && !app.isPackaged) {
-        const now = Date.now();
-        const last = snapshotCooldowns.get(proc.pid) ?? 0;
-        if (now - last > SNAPSHOT_COOLDOWN_MS) {
-          snapshotCooldowns.set(proc.pid, now);
-          try {
-            const dir = app.getPath("logs");
-            mkdirSync(dir, { recursive: true });
-            const file = path.join(dir, `heap-${proc.pid}-${now}.heapsnapshot`);
-            const written = v8.writeHeapSnapshot(file);
-            logWarn("heap-snapshot-written", { path: written });
-          } catch (err) {
-            logWarn("heap-snapshot-failed", { error: String(err) });
+        if (proc.type === "Browser" && mb > SNAPSHOT_THRESHOLD_MB && !app.isPackaged) {
+          const now = Date.now();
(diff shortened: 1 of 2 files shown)


gregpriday:
**Review status: Ready**

PR #3240, opened 2026-03-14
feat(workflow): add typed data flow between workflow nodes

### LINKED ISSUE #3141: Workflow nodes have no typed data flow — inter-node context is lost
## Summary

Workflow nodes have no way to pass structured data to downstream nodes. When a node completes, its result is stored as `{ summary?: string, artifacts?: string[], error?: string }` — an opaque bag that downstream nodes cannot meaningfully consume. This makes it impossible to build AI orchestration workflows where one node's output feeds another's input.

## Problem Statement

Consider a practical workflow: lint → AI review → test. The AI review node needs the lint results to produce a useful critique. The test node might need to know which files the AI changed. Today, none of this inter-node communication is possible.

**Who is affected:** Anyone trying to build multi-step agent workflows where steps have meaningful data dependencies — the core use case of the workflow system.

**Current workaround:** None. All workflow nodes must be self-contained, which means each step re-discovers context independently or relies entirely on file system side-effects with no structured handoff.

**Impact:** The workflow engine becomes a sequencer (A then B then C) rather than an orchestrator (A's output informs B, which informs C). This is the difference between a cron job and a real pipeline.

## Desired Behavior

Workflow nodes should be able to:

1. **Declare what they produce** — a node that runs lint should be able to produce structured lint results (error count, affected files, raw output) that downstream nodes can consume
2. **Reference upstream outputs in their configuration** — a node's `args` should support template expressions like `{{lintNode.result.errorCount}}` that resolve to the actual completed value at execution time
3. **Handle large artifacts by reference** — outputs like full terminal logs or git diffs should be stored out-of-band (e.g., referenced by ID) rather than inlined into workflow state, to prevent state bloat
4. **Maintain backward compatibility** — existing workflow definitions with no inter-node references should continue to work unchanged

The workflow engine already stores node results in `run.nodeStates[nodeId].result` and uses `resolveJsonPath()` for condition evaluation — the infrastructure for reading this data exists. The gap is surfacing it as injectable inputs when compiling downstream node tasks.

## Context

- [`shared/types/task.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/task.ts) — `TaskResult` interface defines the current output shape
- [`shared/types/workflow.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/workflow.ts) — `WorkflowActionConfig` holds `args` but has no input-binding mechanism
- [`shared/types/workflowRun.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/workflowRun.ts) — `NodeState` stores results that are already accessible to the engine
- [`electron/services/WorkflowEngine.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/WorkflowEngine.ts#L192-L247) — `compileNodeToTask()` is where args are serialized into the task; this is where template resolution would slot in
- [`electron/services/WorkflowEngine.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/WorkflowEngine.ts#L456-L477) — `resolveJsonPath()` already exists and does dot-notation path resolution

## Acceptance Criteria

- [ ] A workflow definition can reference an upstream node's output in a downstream node's `args` using a template syntax (e.g., `"{{lintNode.result.errorCount}}"`)
- [ ] Template expressions are resolved at execution time using the actual completed node results, not at workflow compilation time
- [ ] A workflow definition where node B consumes node A's result executes correctly end-to-end, with B receiving A's actual output value
- [ ] Existing workflow definitions with no template expressions continue to work unchanged
- [ ] Template expressions that reference a node that hasn't completed yet or doesn't exist fail the node with a clear error (not a silent undefined)
- [ ] Large outputs (e.g., full command output strings) do not cause the `workflow_runs` SQLite row to become arbitrarily large; there is a size consideration or reference strategy for oversized results

## Edge Cases & Risks

- **Resolution timing**: Templates must be resolved at dispatch time (when the task is compiled), not when the workflow definition is loaded — the values don't exist at load time
- **Circular references**: A node referencing its own output, or a transitive reference cycle, should be caught (though cycle detection already prevents actual graph cycles)
- **Missing upstream nodes**: A template referencing a node that failed or was skipped should produce a clear failure, not silently inject an empty string
- **SQLite row size**: If `nodeStates` is stored as a JSON column in `workflow_runs`, embedding multi-MB terminal outputs will bloat every save. Large results should be stored by reference

gregpriday:
## Summary

- Workflow nodes can now declare structured outputs and downstream nodes can reference them via `{{nodeId.result.field}}` template expressions in their `args`
- Templates are resolved at dispatch time (when `compileNodeToTask()` runs), not at workflow load time, so values are always live results
- Missing or unresolved references fail the node with a clear error rather than silently injecting undefined

Resolves #3141

## Changes

- `WorkflowEngine.ts`: Added `resolveTemplates()` that walks node `args` and replaces `{{nodeId.result.path}}` expressions using existing `resolveJsonPath()` infrastructure. Hooked into `compileNodeToTask()` so templates resolve against `run.nodeStates` at execution time. Large result values (>10KB) are stored by reference ID to prevent SQLite row bloat.
- `shared/types/task.ts`: Extended `TaskResult` with an optional `data` field for structured output beyond the existing `summary`/`artifacts`/`error` bag
- `electron/services/events.ts`: Added `workflow:node-data-resolved` event for observability when template resolution succeeds
- `TaskQueueService.ts`: Minor type alignment for the extended `TaskResult`
- `WorkflowEngine.test.ts`: 385-line test suite covering template resolution, data flow end-to-end, missing node errors, large result handling, and backward compatibility with template-free definitions

## Testing

- All new tests pass: `npm run check` exits clean (0 errors, warnings only from pre-existing issues)
- Backward compatibility verified: existing workflow definitions with no template expressions execute unchanged
- Edge cases covered: missing upstream node, failed node reference, oversized result truncation/reference strategy

### DIFF
--- a/electron/services/WorkflowEngine.ts
+++ b/electron/services/WorkflowEngine.ts
@@ -225,11 +225,19 @@ export class WorkflowEngine {
     }
 
     let resolvedArgs = node.config.args;
-    if (resolvedArgs) {
+    if (resolvedArgs && this.hasTemplateExpressions(resolvedArgs)) {
+      for (const depId of dependencies) {
+        const depState = run.nodeStates[depId];
+        if (!depState || depState.status !== "completed") {
+          throw new Error(
+            `Cannot compile node ${node.id}: dependency ${depId} has not completed yet (status: ${depState?.status ?? "unknown"})`
+          );
+        }
+      }
       try {
         resolvedArgs = this.resolveTemplateArgs(resolvedArgs, run.nodeStates);
       } catch (error) {
-        this.failNode(node.id, run, (error as Error).message);
+        await this.failNode(node.id, run, (error as Error).message);
         throw error;
       }
     }
@@ -286,12 +294,23 @@ export class WorkflowEngine {
     }
 
     if (payload.data) {
-      const dataJson = JSON.stringify(payload.data);
-      if (Buffer.byteLength(dataJson, "utf8") > MAX_RESULT_DATA_BYTES) {
-        const sizeMB = (Buffer.byteLength(dataJson, "utf8") / 1_048_576).toFixed(2);
-        this.failNode(mapping.nodeId, run, `Node result data exceeds 1 MB limit (${sizeMB} MB)`);
-        await this.schedulePersist();
-        await this.checkWorkflowCompletion(run);
+      try {
+        const dataJson = JSON.stringify(payload.data);
+        if (Buffer.byteLength(dataJson, "utf8") > MAX_RESULT_DATA_BYTES) {
+          const sizeMB = (Buffer.byteLength(dataJson, "utf8") / 1_048_576).toFixed(2);
+          await this.failNode(
+            mapping.nodeId,
+            run,
+            `Node result data exceeds 1 MB limit (${sizeMB} MB)`
+          );
+          return;
+        }
+      } catch (e) {
+        await this.failNode(
+          mapping.nodeId,
+          run,
+          `Node result data could not be serialized: ${(e as Error).message}`
+        );
         return;
       }
     }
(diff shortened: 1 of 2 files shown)


gregpriday:
**Review status: Ready**

- Cross-PR note: Shares WorkflowEngine.ts, events.ts, and WorkflowEngine.test.ts with PR #3242. Merge this before #3242 — the approval gate PR will need rebasing after this merges.

PR #3246, opened 2026-03-14
feat(workflows): replace placeholder echo commands with real actions

gregpriday:
## Summary

- Replaces all `echo` placeholder commands in `standard-pr-review.json` with real `terminal.executeCommand` actions that run actual `git` and notification commands
- Adds two new built-in workflow templates: `code-quality-check.json` (lint + typecheck + test in parallel, with `onFailure` routing to a notification step) and `worktree-snapshot.json` (git add + commit + push with proper dependency chain)
- Extends `WorkflowLoader` to validate that action `type` fields are non-empty strings, and expands the test suite to cover all three workflows plus the new validator paths

Resolves #3157

## Changes

- `electron/workflows/standard-pr-review.json` — every node now dispatches a real command; removed the merge placeholder step that had no implementation
- `electron/workflows/code-quality-check.json` — new workflow demonstrating parallel DAG execution and `onFailure` routing with a real notification fallback
- `electron/workflows/worktree-snapshot.json` — new workflow demonstrating a linear dependency chain with git operations
- `shared/types/workflow.ts` — tightened `WorkflowActionConfig.type` from `string` to `string` with a non-empty constraint (validated at load time)
- `electron/services/WorkflowLoader.ts` — added `actionType` non-empty validation; loader now rejects workflows with blank action type strings
- `electron/services/__tests__/WorkflowLoader.test.ts` — 124 lines of new coverage across all three workflow files and the new validation path

## Testing

All new and updated workflows pass `WorkflowLoader.validate()`. The test suite covers cycle detection, diamond-join task scheduling, and the new `actionType` validation. `npm run check` passes clean (zero errors, warnings only from pre-existing issues in unrelated files).

PR #3249, opened 2026-03-14
feat(workflow): add loop node for bounded retry patterns

### LINKED ISSUE #3146: Workflow definitions cannot express bounded retry loops for AI tasks
## Summary

The workflow engine's cycle detection rejects any graph with backward edges, which makes it impossible to define "generate → test → fix → re-test" patterns that are fundamental to AI coding agent orchestration. Workflows that need to iterate until a condition is met (e.g., tests pass, lint is clean, AI output meets quality threshold) cannot be expressed.

## Problem Statement

The `WorkflowLoader` validates that every workflow is a strictly acyclic DAG using DFS cycle detection across all edge types — `dependencies`, `onSuccess`, and `onFailure`. This is correct for preventing infinite loops, but it also blocks intentional bounded iteration.

The most common AI agent workflow pattern is a loop:
1. Ask an agent to write or fix code
2. Run tests or linting
3. If they fail, feed the errors back to the agent and repeat
4. Stop after success or after N attempts

Today, this pattern cannot be expressed in a workflow definition. There is no way to route "back" without triggering a cycle validation error, and there is no node type that encapsulates a bounded internal loop.

**Who is affected:** Any workflow that involves an AI agent + automated quality checks — the most natural and valuable automation pattern in an AI coding IDE.

**Current workaround:** Unrolling the loop statically — defining `generate-1 → test-1 → generate-2 → test-2 → generate-3 → test-3` with conditions to short-circuit on success. This is verbose, creates unreadable DAGs, and requires deciding the max iteration count at definition time in a clunky way.

## Desired Behavior

A workflow definition should be able to express a bounded loop as a single node (or small sub-graph) with:

- A sub-graph of nodes that form the loop body
- A maximum iteration count (hard ceiling to prevent runaway execution)
- An exit condition that terminates the loop early on success (using the same condition syntax already supported in `WorkflowCondition`)
- The loop node appearing as a single atomic node from the parent DAG's perspective, so the parent graph remains acyclic

The outer workflow's cycle detection should remain strict — cycles in the outer graph are still invalid. The loop construct makes cycles explicit, bounded, and opt-in.

## Context

- [`shared/types/workflow.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/workflow.ts#L48-L49) — `WorkflowNodeTypeSchema` currently only allows `"action"`; a loop type would be a new variant
- [`shared/types/workflow.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/workflow.ts#L66-L82) — `WorkflowNodeSchema` defines the current node shape
- [`electron/services/WorkflowLoader.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/WorkflowLoader.ts#L266-L318) — `detectCycles()` runs DFS on all edges including `onSuccess`/`onFailure`
- [`electron/services/WorkflowEngine.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/WorkflowEngine.ts#L192-L247) — `compileNodeToTask()` translates nodes to tasks; a loop node would need its own compilation path
- [`shared/types/workflow.ts`](https://github.com/canopyide/canopy/blob/develop/shared/types/workflow.ts#L20-L41) — `WorkflowConditionSchema` already supports status and result-based conditions that could drive loop exit

## Acceptance Criteria

- [ ] A workflow definition can include a node that executes a sub-graph repeatedly up to a configurable `maxIterations` limit
- [ ] The loop terminates early when a defined exit condition evaluates to true
- [ ] When `maxIterations` is reached without the exit condition being satisfied, the loop node is treated as failed (triggering `onFailure` routing in the parent graph)
- [ ] The outer DAG's cycle detection is unaffected — backward edges in the outer graph are still rejected
- [ ] The loop body sub-graph is itself validated as a DAG (no cycles within the loop body either)
- [ ] A workflow using a loop node executes correctly end-to-end with the engine scheduling and completing iterations as expected

## Edge Cases & Risks

- **Iteration state isolation vs. accumulation**: Each iteration of the loop needs clear semantics about what state carries forward from the previous iteration (e.g., the last iteration's output feeds the next iteration's input) vs. what resets
- **Crash recovery**: If the app crashes mid-loop, the engine needs to know which iteration was in progress and resume or fail cleanly — the current `nodeStates` structure must accommodate this
- **Nested loops**: A loop node containing another loop node should either be explicitly supported or explicitly rejected with a clear validation error
- **`maxIterations` limits**: There should be a reasonable upper bound enforced by the schema (e.g., max 20 iterations) to prevent accidental resource exhaustion

gregpriday:
## Summary

- Adds a `loop` node type to the workflow schema, giving workflow definitions a first-class way to express bounded retry patterns like generate → test → fix → repeat
- The loop body is validated as an inner DAG during load, so cycle detection on the outer graph is completely unaffected
- Loop nodes schedule body tasks on-demand per iteration rather than upfront, keeping execution state clean and crash-recoverable

Resolves #3146

## Changes

- `shared/types/workflow.ts` — adds `WorkflowLoopNodeSchema` with `body`, `maxIterations` (capped at 20), and an optional `exitCondition` using the existing `WorkflowConditionSchema`
- `shared/types/workflowRun.ts` — extends node run state to track `currentIteration` and `totalIterations`
- `electron/services/WorkflowLoader.ts` — validates loop nodes (body is a DAG, no nested loops, `maxIterations` in range), skips body nodes during outer cycle detection
- `electron/services/WorkflowEngine.ts` — compiles loop nodes to a task that drives iteration, re-schedules body nodes each pass, checks the exit condition after each iteration, and marks the node failed when `maxIterations` is exhausted without success
- Tests in `WorkflowLoader.test.ts` and `WorkflowEngine.test.ts` covering the full acceptance criteria, edge cases (nested loop rejection, max iteration exhaustion, early exit on condition)

## Testing

Full test suite passes (`npm run check` clean, all new unit tests green). The loop body scheduling test verifies on-demand scheduling rather than upfront expansion, which is the key correctness property for crash recovery semantics.


gregpriday:

**Fixes applied:**
- Resolved merge conflict with develop — integrated both approval node (from develop) and loop node (from this PR) into unified workflow schema
- `WorkflowNodeTypeSchema` now supports `"action" | "approval" | "loop"`
- `WorkflowNodeSchema` includes `WorkflowActionNodeSchema`, `WorkflowApprovalNodeSchema`, and `WorkflowLoopNodeSchema`
- WorkflowEngine dependency resolution skips both loop and approval nodes (no task mappings)
- All tests from both features preserved in separate `describe` blocks

- CI: All checks passing
- Conflicts: Resolved (was CONFLICTING, now MERGEABLE)
- Cross-PR note: Shares `WorkflowLoader.ts` and `workflow.ts` with PR #3246 — merge this PR first to avoid re-conflicts

PR #3252, opened 2026-03-15
fix(ui): use static spinner arc for waiting state indicator

### LINKED ISSUE #3250: Use static spinner arc for waiting state indicator
## Summary

The `waiting` agent state indicator in the terminal title bar uses a full hollow circle (`HollowCircle`), while the `working` state uses a spinning half-arc (`SpinnerCircle`). These two shapes are visually unrelated, breaking the logical progression the icon system implies.

## Problem Statement

When an agent transitions from `waiting` → `working`, the indicator jumps from a full circle to a spinning half-arc. The two shapes share no visual DNA, so the relationship between states isn't self-evident. A user can't intuit that the static icon is about to become the animated one.

The `directing` state uses a half-filled circle (`SolidCircle`) which is also visually disconnected from the spinner arc family.

## Desired Behavior

The `waiting` state should display the same half-arc shape as `SpinnerCircle`, but without any rotation animation. It sits in its resting position — poised, not yet moving. This makes the icon family coherent:

- **Waiting**: half arc, static (the spinner at rest, not yet activated)
- **Working**: half arc, spinning green (the same shape, now animated and coloured)

The transition from waiting → working then feels like the icon "waking up" — same shape, same origin, but now live. The full hollow circle currently used for `waiting` should be retired from that role.

## Context

The state-to-icon mapping lives in `terminalStateConfig.tsx`, and the icons themselves are in `AgentStateCircles.tsx`. The chip is rendered in the terminal header by `TerminalHeaderContent.tsx`, where `animate-spin-slow` is applied only when `agentState === "working"` — so the static waiting variant would naturally get no animation just by not meeting that condition.

**Relevant code:**
- [`src/components/icons/AgentStateCircles.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/icons/AgentStateCircles.tsx) — defines `SpinnerCircle`, `HollowCircle`, and `SolidCircle`
- [`src/components/Worktree/terminalStateConfig.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Worktree/terminalStateConfig.tsx#L8) — maps `waiting` → `HollowCircle`
- [`src/components/Terminal/TerminalHeaderContent.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/TerminalHeaderContent.tsx#L77-L83) — applies `animate-spin-slow` conditionally on `working` state

## Acceptance Criteria

- [ ] The `waiting` state indicator displays the same half-arc shape as the `working` spinner, with no rotation animation
- [ ] The `working` state continues to spin as before — same shape, now animated
- [ ] The visual transition between `waiting` and `working` looks coherent (same shape, animation starts)
- [ ] The indicator renders correctly in the terminal title bar chip and anywhere else `STATE_ICONS` is consumed (e.g. worktree cards)
- [ ] No other agent states are affected

## Edge Cases & Risks

- `STATE_ICONS` is used in multiple places beyond the title bar chip — verify each render site applies animation conditionally and not unconditionally to the icon component itself
- The `HollowCircle` may still be needed for another state or future use; check all usages before removing it

gregpriday:
## Summary

- The `waiting` agent state now uses `SpinnerCircle` (the same half-arc shape as `working`) instead of `HollowCircle`, making the two states visually related
- The transition from `waiting` to `working` now looks like the icon waking up: same shape, animation starts
- `HollowCircle` is no longer imported in `terminalStateConfig.tsx` since nothing else in that file uses it

Resolves #3250

## Changes

- `src/components/Worktree/terminalStateConfig.tsx`: swap `waiting` icon from `HollowCircle` to `SpinnerCircle`, remove unused `HollowCircle` import

## Testing

- Checked all `STATE_ICONS` render sites: animation is applied conditionally (`agentState === "working"`) in `TerminalHeaderContent.tsx`, so `waiting` gets the static arc with no extra changes needed
- `HollowCircle` is still defined in `AgentStateCircles.tsx` and available for future use; no other states reference it

### DIFF
--- a/src/components/Worktree/terminalStateConfig.tsx
+++ b/src/components/Worktree/terminalStateConfig.tsx
@@ -1,11 +1,11 @@
 import { Play, Circle, CheckCircle2, XCircle } from "lucide-react";
 import type { AgentState } from "@/types";
-import { SpinnerCircle, HollowCircle, SolidCircle } from "@/components/icons/AgentStateCircles";
+import { SpinnerCircle, SolidCircle } from "@/components/icons/AgentStateCircles";
 
 export const STATE_ICONS: Record<AgentState, React.ComponentType<{ className?: string }>> = {
   working: SpinnerCircle,
   running: Play,
-  waiting: HollowCircle,
+  waiting: SpinnerCircle,
   directing: SolidCircle,
   idle: Circle,
   completed: CheckCircle2,


gregpriday:
**Review status: Ready**

PR #3325, opened 2026-03-16
fix(terminal): prevent mouse/focus events from triggering directing state

### LINKED ISSUE #3290: Agent terminal enters directing state on click, not just on typing
## Summary

An agent terminal transitions to the `directing` state when the user merely clicks inside it, even without typing anything. The `directing` state is intended to signal that the user is actively composing a prompt — it should only appear when the user actually types.

## Current Behavior

Clicking anywhere in an agent terminal — including inside the xterm viewport — can trigger the `directing` state transition. This causes the state indicator to change and the debounce timer to start, even though the user has not begun composing input.

## Expected Behavior

The `directing` state should only be entered when the user actually types a character (printable key, control character, paste, etc.). A focus event, pointer event, or click that doesn't produce character input should have no effect on agent state.

## Root Cause Analysis

The `onData` handler on the xterm.js terminal instance unconditionally calls `onUserInput` for every byte stream produced by user interaction:

```typescript
const inputDisposable = terminal.onData((data) => {
  if (!managed.isInputLocked) {
    this.onUserInput(id);
    terminalClient.write(id, data);
```

[`src/services/terminal/TerminalInstanceService.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalInstanceService.ts#L593-L596)

xterm.js fires `onData` for **all** user-generated input that reaches the PTY — including mouse-tracking escape sequences. When an agent CLI uses a mouse-aware TUI (cursor positioning, interactive menus, etc.), a plain left-click inside the terminal is encoded as an escape sequence (e.g., `\x1b[M...`) and delivered via `onData`. This causes `onUserInput` to be called, which sets the agent state to `directing`:

```typescript
private onUserInput(id: string): void {
  // ...
  if (managed.kind === "agent" && managed.canonicalAgentState === "waiting") {
    if (managed.agentState !== "directing") {
      managed.agentState = "directing";
      // ... notifies subscribers and store
      useTerminalStore.getState().updateAgentState(id, "directing");
    }
    // resets debounce timer
  }
}
```

[`src/services/terminal/TerminalInstanceService.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalInstanceService.ts#L98-L134)

By contrast, the HybridInputBar correctly gates on `isUserEvent("input") || isUserEvent("delete")` to avoid triggering `directing` on mere focus/selection changes:

[`src/components/Terminal/HybridInputBar.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/HybridInputBar.tsx#L1336-L1349)

The `onData` path in `TerminalInstanceService` lacks an equivalent filter, so mouse clicks (and other non-typing interactions) incorrectly enter the `directing` state.

## Acceptance Criteria

- [ ] Clicking inside an agent terminal that is in `waiting` state does not change the state to `directing`
- [ ] Typing a character in the HybridInputBar or xterm viewport while the agent is `waiting` still transitions to `directing`
- [ ] The `directing` debounce timer is not started by click/focus events
- [ ] No regression: all existing `directing` state transitions from actual typing continue to work correctly

gregpriday:
## Summary

- Agent terminals were transitioning to `directing` state on click, not just on typing. The root cause was `onData` in `TerminalInstanceService` calling `onUserInput` unconditionally, including for mouse-tracking escape sequences (`\x1b[M...`, `\x1b[<...`, URXVT format) and focus reports (`\x1b[I`/`\x1b[O`).
- Added `isNonKeyboardInput(data)` to filter these out before calling `onUserInput`. The fix is a single guard in the `onData` handler — if the byte sequence is a known non-keyboard event, `onUserInput` is skipped and the data is still forwarded to the PTY normally.
- Arrow keys, function keys, control characters, and all printable input are unaffected.

Resolves #3290

## Changes

- `src/services/terminal/TerminalInstanceService.ts` — added `isNonKeyboardInput` helper and guard in `onData` handler
- `src/services/terminal/__tests__/isNonKeyboardInput.test.ts` — 85-line test suite covering X10/SGR/URXVT mouse sequences, focus reports, and negative cases (keyboard input that must not be filtered)

## Testing

Unit tests pass (`npm run check` clean). Covered cases: X10 mouse (`\x1b[M`), SGR mouse (`\x1b[<`), URXVT mouse (numeric format), focus-in/out (`\x1b[I`/`\x1b[O`), and verified arrow keys, function keys, Home/End, printable characters, and control characters all return false.


gregpriday:
**Review status: Ready**

Adds `isNonKeyboardInput()` filter to skip mouse tracking sequences (X10, SGR, URXVT) and focus reports (`\x1b[I`/`\x1b[O`) from triggering `onUserInput()`. Thorough test coverage with edge cases. No cross-PR interactions.

PR #3339, opened 2026-03-16
fix(github): use IssueOrder instead of PullRequestOrder in PR list query

### LINKED ISSUE #3336: PR list fails with GraphQL type mismatch when sort order is set
## Summary

The pull requests list query fails with a GitHub GraphQL API error when a sort order is applied. The error message is: `Request failed due to the following response errors: Type mismatch on variable order by and argument`.

## Current Behavior

When the PR list is loaded with a sort order (newest or recently updated), the GraphQL query fails with:

```
Request failed due to the following response errors: Type mismatch on variable order by and argument
```

The PR panel shows a GitHub API error instead of the list of pull requests.

## Expected Behavior

Pull requests load successfully regardless of which sort order is selected (newest / recently updated).

## Root Cause Analysis

`LIST_PRS_QUERY` declares its sort variable as `$orderBy: PullRequestOrder`, but `PullRequestOrder` does not exist in GitHub's GraphQL schema. The `pullRequests` connection on `Repository` accepts `IssueOrder` — the same input type used by the `issues` connection.

The issues query (`LIST_ISSUES_QUERY`) correctly uses `$orderBy: IssueOrder` and works fine. The PR query was written with an incorrect type name when sort support was added.

**Evidence:**

- [`electron/services/github/GitHubQueries.ts#L77`](https://github.com/canopyide/canopy/blob/develop/electron/services/github/GitHubQueries.ts#L77) — `LIST_PRS_QUERY` declares `$orderBy: PullRequestOrder` (invalid type)
- [`electron/services/github/GitHubQueries.ts#L13`](https://github.com/canopyide/canopy/blob/develop/electron/services/github/GitHubQueries.ts#L13) — `LIST_ISSUES_QUERY` correctly declares `$orderBy: IssueOrder` (valid type)
- [`electron/services/GitHubService.ts#L944-L1004`](https://github.com/canopyide/canopy/blob/develop/electron/services/GitHubService.ts#L944-L1004) — call site passes `{ field: "UPDATED_AT" | "CREATED_AT", direction: "DESC" }` to `LIST_PRS_QUERY`

The inline `pullRequests` query inside `buildBatchPRQuery` (line 371) uses a hardcoded `orderBy` object literal rather than a variable, so it is unaffected.

## Acceptance Criteria

- [ ] The PR list loads successfully when sort order is "newest" (CREATED_AT)
- [ ] The PR list loads successfully when sort order is "recently updated" (UPDATED_AT)
- [ ] No GraphQL type mismatch error is returned from the GitHub API
- [ ] The issues list sort continues to work as before (no regression)

## Edge Cases & Risks

The `IssueOrderField` enum only supports `CREATED_AT`, `UPDATED_AT`, and `COMMENTS`. Any sort field values beyond these would need to fall back to a valid option.

gregpriday:
## Summary

- `LIST_PRS_QUERY` declared its sort variable as `$orderBy: PullRequestOrder`, which doesn't exist in GitHub's GraphQL schema. The `pullRequests` connection on `Repository` uses `IssueOrder`, the same type used by `issues`.
- Changed the variable type to `IssueOrder` to match the actual schema, fixing the type mismatch error.
- Added a test to `GitHubQueries.test.ts` that verifies the PR query variable declaration uses `IssueOrder`.

Resolves #3336

## Changes

- `electron/services/github/GitHubQueries.ts`: changed `$orderBy: PullRequestOrder` to `$orderBy: IssueOrder` in `LIST_PRS_QUERY`
- `electron/services/github/__tests__/GitHubQueries.test.ts`: added test asserting correct variable type in the PR list query

## Testing

Unit test added and passing. The fix matches the pattern already working correctly in `LIST_ISSUES_QUERY`.

### DIFF
--- a/electron/services/github/GitHubQueries.ts
+++ b/electron/services/github/GitHubQueries.ts
@@ -74,7 +74,7 @@ export const LIST_ISSUES_QUERY = `
 `;
 
 export const LIST_PRS_QUERY = `
-  query GetPRs($owner: String!, $repo: String!, $states: [PullRequestState!], $cursor: String, $limit: Int = 20, $orderBy: PullRequestOrder) {
+  query GetPRs($owner: String!, $repo: String!, $states: [PullRequestState!], $cursor: String, $limit: Int = 20, $orderBy: IssueOrder) {
     repository(owner: $owner, name: $repo) {
       pullRequests(first: $limit, after: $cursor, states: $states, orderBy: $orderBy) {
         totalCount
--- a/electron/services/github/__tests__/GitHubQueries.test.ts
+++ b/electron/services/github/__tests__/GitHubQueries.test.ts
@@ -1,5 +1,12 @@
 import { describe, expect, it } from "vitest";
-import { buildBatchPRQuery } from "../GitHubQueries.js";
+import { buildBatchPRQuery, LIST_PRS_QUERY } from "../GitHubQueries.js";
+
+describe("LIST_PRS_QUERY", () => {
+  it("uses IssueOrder (not PullRequestOrder) for the orderBy variable type", () => {
+    expect(LIST_PRS_QUERY).toContain("$orderBy: IssueOrder");
+    expect(LIST_PRS_QUERY).not.toContain("PullRequestOrder");
+  });
+});
 
 describe("buildBatchPRQuery", () => {
   it("escapes owner, repo, and branch values in generated GraphQL query", () => {


gregpriday:
**Review status: Ready**

Clean one-line fix with matching test. Correct — GitHub's `pullRequests` connection uses `IssueOrder`, not `PullRequestOrder`.

PR #3361, opened 2026-03-16
feat(terminal): reduce xterm.js scrollback for background terminals

### LINKED ISSUE #3281: Reduce xterm.js scrollback for background terminals to reclaim memory
## Summary

Background terminals — agent panels in inactive worktrees or panels not currently visible — hold their full scrollback buffer in memory indefinitely, even though they are not being viewed. Automatically reducing the scrollback limit for terminals in lower renderer tiers would free substantial memory in proportion to how many panels are backgrounded.

## Problem Statement

Canopy's `TerminalRendererPolicy` already classifies terminals into FOCUSED, VISIBLE, and BACKGROUND tiers with hysteresis-debounced transitions. However, tier changes currently only affect rendering behavior and PTY activity signaling — they do not adjust the xterm.js scrollback buffer.

A backgrounded terminal retains its full allocated buffer (up to 10,000 lines for agents) even though no user is viewing it. When a user has 20 agent panels across several worktrees, only a handful are ever visible at once, yet all 20 hold full-size buffers. The memory is entirely idle.

xterm.js supports dynamic scrollback adjustment at runtime: setting `terminal.options.scrollback` to a lower value immediately trims the oldest lines from the circular buffer, freeing memory synchronously. The reverse — increasing the limit — expands capacity for new incoming lines without restoring lost history.

The tier system in [`TerminalRendererPolicy.applyRendererPolicyImmediate`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalRendererPolicy.ts#L86-L135) is the natural integration point. It already distinguishes "active" vs "background" backend states and triggers wake/restore cycles on tier upgrade. Scrollback adjustment fits this pattern.

## Desired Behavior

When a terminal transitions to the BACKGROUND tier and remains there past the existing downgrade hysteresis window, its xterm.js scrollback should shrink to a lower limit, freeing the buffer memory it no longer needs. When the terminal is brought back to an active tier, its scrollback limit should expand again to accommodate new output.

The transition must be safe:
- If the user is currently scrolled up in the terminal's history, the reduction should be deferred or skipped to avoid a jarring scroll snap to the bottom
- If there is an active text selection, the reduction should be deferred to prevent selection index corruption
- The alternate buffer (vim, less, top) is unaffected — xterm.js enforces `scrollback: 0` for alternate buffers regardless

The goal is transparent to the user: background terminals quietly use less memory; when the user switches back, the terminal continues streaming live output normally within the restored limit.

## Context

Relevant infrastructure already in place:

- [`TerminalRendererPolicy`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalRendererPolicy.ts) — tier management with debounced downgrade and `onTierApplied` callback hook
- [`src/utils/scrollbackConfig.ts`](https://github.com/canopyide/canopy/blob/develop/src/utils/scrollbackConfig.ts) — type-specific policies with multipliers and limits; `PERFORMANCE_MODE_SCROLLBACK = 100` shows the precedent for policy-driven scrollback reduction
- [`src/store/slices/terminalRegistry/index.ts#L304-L322`](https://github.com/canopyide/canopy/blob/develop/src/store/slices/terminalRegistry/index.ts#L304-L322) — where the initial effective scrollback is computed from user settings and terminal type
- The `ManagedTerminal` type holds the live xterm.js `terminal` instance, making it accessible to policy decisions

## Acceptance Criteria

- [ ] Terminals that have been in BACKGROUND tier past the hysteresis window have a reduced xterm.js scrollback compared to their configured limit
- [ ] Terminals that return to an active tier have their scrollback limit restored to the configured value, so new output fills up to the full allowed buffer
- [ ] A terminal whose user is actively scrolled up in history is not trimmed while the user is reading it
- [ ] A terminal with an active text selection is not trimmed until the selection is cleared
- [ ] The alternate buffer (vim, less, top, etc.) is unaffected
- [ ] No visible jank, snap-to-bottom, or selection glitches during normal panel-switching workflows
- [ ] Agent state detection continues to function correctly (the detection window is the last ~50 lines, well within any reasonable background minimum)

## Edge Cases & Risks

- Rapid worktree switching could cause many terminals to cycle through tiers quickly; the existing `TIER_DOWNGRADE_HYSTERESIS_MS` debounce already mitigates this but the scrollback changes should not add latency to tier upgrades
- The background scrollback minimum must be high enough that new output from a still-running agent isn't immediately lost — the terminal keeps receiving data even while backgrounded
- History visible to the user when they switch to a background terminal will only extend back to what fit in the reduced buffer; this is a known tradeoff of trimming-based approaches

gregpriday:
## Summary

- Terminals that transition to the BACKGROUND tier now have their xterm.js scrollback shrunk to a reduced limit, freeing the idle buffer memory
- When a terminal returns to an active tier, the scrollback limit is restored to the configured value so new output fills up to the full buffer
- Reductions are skipped if the user is actively scrolled up or has a live text selection, preventing jarring scroll snaps or selection corruption

Resolves #3281

## Changes

- `shared/config/scrollback.ts`: added `BACKGROUND_SCROLLBACK` constant (500 lines) to define the reduced background limit
- `src/services/terminal/TerminalInstanceService.ts`: added `applyScrollbackForTier()` method wired into `onTierApplied` — expands on upgrade, shrinks on downgrade with scroll-position and selection guards
- `src/services/terminal/__tests__/TerminalInstanceService.scrollback.test.ts`: unit tests covering all guard conditions (scrolled up, active selection, alternate buffer) and the normal upgrade/downgrade paths

## Testing

Unit tests pass. Guard conditions (scroll position, text selection, alternate buffer) tested directly. The alternate buffer case is implicitly safe — xterm.js enforces `scrollback: 0` for alternate buffers regardless of what we set.

### DIFF
--- a/src/services/terminal/__tests__/TerminalInstanceService.scrollback.test.ts
+++ b/src/services/terminal/__tests__/TerminalInstanceService.scrollback.test.ts
@@ -125,6 +125,7 @@ describe("TerminalInstanceService - Scrollback", () => {
 
       service.reduceScrollback("t1", 500);
       expect(managed.terminal.options.scrollback).toBe(5000);
+      expect(managed.writtenData).toHaveLength(0);
     });
 
     it("skips terminals with active text selection", () => {
@@ -134,6 +135,7 @@ describe("TerminalInstanceService - Scrollback", () => {
 
       service.reduceScrollback("t1", 500);
       expect(managed.terminal.options.scrollback).toBe(5000);
+      expect(managed.writtenData).toHaveLength(0);
     });
 
     it("skips when current scrollback already at or below target", () => {


gregpriday:

**Fixes applied:**
- Changed `SCROLLBACK_BACKGROUND` from 1000 to 500 lines

**Reason:** PR #3365 reduces `SCROLLBACK_DEFAULT` from 2500 to 1000. With `SCROLLBACK_BACKGROUND` also at 1000, the background reduction would be a no-op for users on the default setting. Lowering to 500 preserves meaningful memory savings when terminals move to background tier.

- CI: Re-running after fix
- Conflicts: None
- Cross-PR note: Merge #3365 (scrollback defaults) before this PR. File overlap in `scrollback.ts` and scrollback tests — both will need rebase.

PR #3424, opened 2026-03-16
fix(panel): persist panel identity to prevent position/size swap on project switch

### LINKED ISSUE #3419: Panel positions and sizes swap after switching projects
## Summary

When a project containing a two-panel layout is left and then returned to, the two panels exchange their left/right positions and their widths. A panel that was wide on the right is now narrow on the left, and vice versa. The layout the user had carefully arranged is not what they see when they return.

## Current Behavior

1. A project has two grid panels: an agent terminal (left, narrow) and a dev-preview panel (right, wide — dragged to fill most of the screen).
2. User switches to a different project, does some work, then switches back.
3. On return, the two panels have swapped: the dev-preview panel is now on the left and is narrow; the agent terminal is now on the right and is wide.

Both position (left/right) **and** effective size swap simultaneously.

**Reproducibility**: Reproducible on a project switch — once per occurrence.

## Root Cause Analysis

The two-pane split layout uses two independent state sources:

1. **Split ratio** — stored in `twoPaneSplitStore` as `ratioByWorktreeId[worktreeId]` (a number 0–1 representing what fraction the *left* panel occupies). This is persisted to localStorage and survives project switches correctly.

2. **Panel order (left vs right)** — determined at render time by the order of entries in the `tabGroups` array for the active worktree, which is restored from persisted panel state on project rehydration. The tab groups restore order is not guaranteed to match the original visual order.

When the panel order is restored in a different sequence, the saved ratio — which was calibrated for the *old* left/right assignment — now applies to the wrong panel. The result is that both the positions and the perceived widths are swapped.

`TwoPaneSplitLayout` contains in-session swap detection via `prevTerminalIdsRef` that can invert the ratio when a live reorder is detected, but `prevTerminalIdsRef` initialises to `null` on every mount, so it cannot detect a swap that happened across a project switch.

**Evidence:**
- [`src/components/Terminal/TwoPaneSplitLayout.tsx#L56-L70`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/TwoPaneSplitLayout.tsx#L56-L70) — swap detection uses `prevTerminalIdsRef` which resets to `null` on mount; cannot detect cross-session order changes
- [`src/store/twoPaneSplitStore.ts#L11-L13`](https://github.com/canopyide/canopy/blob/develop/src/store/twoPaneSplitStore.ts#L11-L13) — `ratioByWorktreeId` stores only a scalar ratio with no record of which panel occupies the left slot
- [`src/components/Terminal/ContentGrid.tsx#L834-L842`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/ContentGrid.tsx#L834-L842) — `twoPaneTerminals` is built from `tabGroups.slice(0, 2)`, so restore order of tab groups directly determines left vs right

## Acceptance Criteria

- [ ] Returning to a project restores the panel in the same left/right position it was in when the user left
- [ ] The split ratio (relative panel widths) is also correct on return — wide panel is still wide, narrow panel is still narrow
- [ ] The fix works regardless of panel kinds (terminal + agent, terminal + dev-preview, agent + browser, etc.)
- [ ] A fresh project with no saved layout is unaffected — default ordering still applies

## Edge Cases & Risks

- **Three or more panels**: the bug is specific to the two-pane split path; multi-panel auto-grid is unaffected but any fix must not disturb it.
- **preferPreview mode**: `TwoPaneSplitLayout` has a `preferPreview` heuristic that auto-assigns the dev-preview panel to the right at a larger ratio when no worktree ratio is stored. This must continue to work for worktrees that have never had a manually-dragged split.
- **In-session drag swaps**: the existing `prevTerminalIdsRef` inversion logic handles live reorders; the fix must not break that behaviour.

gregpriday:
## Summary

- Panels in a two-pane split layout were swapping left/right positions and widths when switching back to a project, because `twoPaneSplitStore` persisted only a scalar ratio with no record of which panel ID occupied the left slot
- The fix persists the left panel's ID alongside the ratio, so when the layout mounts it can detect if the restored tab group order differs from what was saved and invert the ratio accordingly
- A backfill path handles legacy persisted state (null panel IDs) gracefully, defaulting to the current order so existing sessions are unaffected

Resolves #3419

## Changes

- `twoPaneSplitStore.ts`: extended state shape to store `leftPanelIdByWorktreeId` alongside the ratio; added `resolveRatio` helper that inverts when the persisted left ID no longer matches the current left panel
- `TwoPaneSplitLayout.tsx`: calls `resolveRatio` on every render so cross-session order changes are corrected before the ratio is applied; existing in-session `prevTerminalIdsRef` inversion logic is preserved and still fires for live drags
- `fix(panel): backfill legacy null panel IDs and extract ratio resolver` commit extracts the resolver into a testable helper and handles null IDs from older persisted state

## Testing

Unit tests in `src/store/__tests__/twoPaneSplitStore.test.ts` cover the swap-detection, inversion, backfill, `preferPreview` default, and in-session drag paths. Persistence boundary hardening tests in `persistenceBoundaryHardening.test.ts` were updated to match the new state shape. All tests pass.

### DIFF
--- a/src/components/Terminal/TwoPaneSplitLayout.tsx
+++ b/src/components/Terminal/TwoPaneSplitLayout.tsx
@@ -57,7 +57,12 @@ export function TwoPaneSplitLayout({
 
   // Backfill panel IDs for legacy entries migrated from v0 (panels are [null, null])
   useEffect(() => {
-    if (storedEntry && storedEntry.panels[0] === null && storedEntry.panels[1] === null && activeWorktreeId) {
+    if (
+      storedEntry &&
+      storedEntry.panels[0] === null &&
+      storedEntry.panels[1] === null &&
+      activeWorktreeId
+    ) {
       setWorktreeRatio(activeWorktreeId, storedEntry.ratio, [terminals[0].id, terminals[1].id]);
     }
   }, [storedEntry, activeWorktreeId, terminals, setWorktreeRatio]);


gregpriday:
**Review status: Ready**

Cross-PR holistic review — no file overlap or semantic conflicts with other open PRs (#3421, #3422, #3423). All 4 PRs are independent and safe to merge in any order.

PR #3445, opened 2026-03-16
feat(terminal): announce search result status to screen readers

### LINKED ISSUE #3441: Terminal search result status not announced to screen readers
## Summary

The terminal search bar shows status feedback for search results ("Found", "No matches", "Invalid regex") but these status changes are not announced to screen readers. A user navigating with VoiceOver or NVDA will perform a search and receive no feedback about whether results were found.

## Current Behavior

The status text element renders conditionally based on `searchStatus` but has no `aria-live` attribute, so changes to its content are not announced by assistive technology.

```typescript
{searchTerm && searchStatus !== "idle" && (
  <span className="text-xs px-1.5">
    {searchStatus === "found" && "Found"}
    {searchStatus === "none" && "No matches"}
    {searchStatus === "invalidRegex" && "Invalid regex"}
  </span>
)}
```

**Evidence:**
- [`src/components/Terminal/TerminalSearchBar.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/TerminalSearchBar.tsx#L220-L231) — status span without `aria-live`

## Desired Behavior

When search results change, the new status ("Found", "No matches", "Invalid regex") is announced by screen readers without interrupting other reading. The announcement is non-disruptive (polite, not assertive).

## Acceptance Criteria

- [ ] Search result status changes ("Found", "No matches", "Invalid regex") are announced by VoiceOver / screen readers
- [ ] The announcement uses `aria-live="polite"` so it doesn't interrupt active reading
- [ ] The element is always present in the DOM (even when empty) so live region announcements work correctly — a live region must exist before content changes to be announced

## Edge Cases & Risks

`aria-live` regions must be present in the DOM before content is inserted to be reliably announced. If the current implementation mounts/unmounts the span conditionally, the live region should be restructured to always be present but conditionally populated.

gregpriday:
## Summary

- Adds a persistent `aria-live="polite"` region to `TerminalSearchBar` so screen readers announce status changes ("Found", "No matches", "Invalid regex") without interrupting active reading
- The live region is always present in the DOM (never conditionally mounted) so announcements fire reliably, per the ARIA spec requirement that live regions exist before content is inserted
- Extracts a `statusText` helper to eliminate the duplicated status-to-string mapping between the visible span and the live region

Resolves #3441

## Changes

- `src/components/Terminal/TerminalSearchBar.tsx`: restructured status rendering to use a persistent `aria-live="polite"` span alongside a `statusText` helper function
- `src/components/Terminal/__tests__/TerminalSearchBar.test.tsx`: added test coverage for the aria-live region presence, status text content, and idle/empty state

## Testing

Unit tests added and passing. The aria-live region is verified to be present in the DOM at all times and to reflect the correct status text for each search state (found, none, invalidRegex, idle).

### DIFF
--- a/src/components/Terminal/TerminalSearchBar.tsx
+++ b/src/components/Terminal/TerminalSearchBar.tsx
@@ -148,6 +148,15 @@ export function TerminalSearchBar({ terminalId, onClose, className }: TerminalSe
     };
   }, []);
 
+  const statusText =
+    searchTerm && searchStatus !== "idle"
+      ? searchStatus === "found"
+        ? "Found"
+        : searchStatus === "none"
+          ? "No matches"
+          : "Invalid regex"
+      : "";
+
   return (
     <div
       className={cn(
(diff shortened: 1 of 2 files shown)


gregpriday:
**Review status: Ready**

PR #3446, opened 2026-03-16
fix(worktree): remove spurious self-assign success notification

### LINKED ISSUE #3444: Self-assign during worktree creation shows spurious re-entry summary notification
## Summary

When a user creates a worktree with the \"Assign to me\" checkbox enabled and then steps away from the app, they are greeted by a \"While you were away\" re-entry summary notification upon returning. This summary includes the issue self-assignment as a completed action — even though the user deliberately triggered it seconds before leaving.

## Current Behavior

1. User opens the New Worktree dialog, selects an issue, checks \"Assign to me\", and submits.
2. The worktree is created and the issue is assigned to the user on GitHub.
3. User walks away (or the app loses focus for > 3 seconds).
4. On returning, the \"While you were away\" banner appears, reporting \"1 completed\" — referring to the issue assignment they explicitly initiated.

The assignment is reported in the notification center history and contributes to the re-entry summary count via `useReEntrySummary`.

## Expected Behavior

No notification of any kind should appear for a user-initiated self-assignment. The user explicitly checked a box in a dialog — they do not need to be informed of the outcome of their own deliberate action when they return to the app.

The notification system's purpose is to surface events that happened _without_ direct user input (e.g., agent completions, background CI checks). Confirming a form submission the user just made is noise, not signal.

## Root Cause Analysis

In `NewWorktreeDialog.tsx`, a successful self-assignment fires `notify()` with `priority: "low"` and `countable: false`:

```typescript
notify({
  type: "success",
  title: "Issue Assigned",
  message: `Issue #${selectedIssue.number} assigned to @${currentUser}`,
  priority: "low",
  countable: false,
});
```

In `notify.ts`, `priority: "low"` means `shouldToast = false`, so the history entry is stored with `seenAsToast: false`. The `useReEntrySummary` hook surfaces all history entries where `!e.seenAsToast && !e.summarized && e.timestamp >= blurTime`. Because `seenAsToast` is `false`, the assignment notification is treated as something the user may have missed — and is included in the re-entry summary on next focus.

**Evidence:**

- [`src/components/Worktree/NewWorktreeDialog.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Worktree/NewWorktreeDialog.tsx#L463-L491) — the self-assign block that fires the success notification
- [`src/lib/notify.ts`](https://github.com/canopyide/canopy/blob/develop/src/lib/notify.ts#L116-L131) — priority routing; `priority: "low"` sets `seenAsToast: false`, making the entry eligible for re-entry summary
- [`src/hooks/useReEntrySummary.ts`](https://github.com/canopyide/canopy/blob/develop/src/hooks/useReEntrySummary.ts#L70-L73) — filter that picks up the notification as "missed" on return

## Acceptance Criteria

- [ ] When a user creates a worktree with self-assign enabled and then returns to the app after being away, no re-entry summary notification appears for the issue assignment
- [ ] No history entry for the issue assignment appears in the notification center after a successful self-assign during worktree creation
- [ ] The error-case notification (when assignment fails) is unaffected — failed assignments should still notify the user
- [ ] No regression in re-entry summary behaviour for genuinely background events (agent completions, recipe failures, etc.)

## Edge Cases & Risks

- The failure path (`notify` with `type: "warning"` at the same call site) is appropriate to keep — the user needs to know if assignment silently failed so they can assign manually on GitHub. Only the success notification should be removed.
- If the notification system grows a concept of \"user-initiated\" vs \"background\" source tracking in the future, this fix should be consistent with that model.

gregpriday:
## Summary

- Removes the success `notify()` call after a self-assign during worktree creation so it no longer appears in the re-entry summary when the user returns to the app
- The user explicitly checked "Assign to me" in the dialog, so confirming the outcome is noise, not signal
- The warning notification on assignment failure is untouched

Resolves #3444

## Changes

- `src/components/Worktree/NewWorktreeDialog.tsx`: deleted the `notify({ type: "success", ... })` block in the self-assign success path; the failure path (`type: "warning"`) is preserved

## Testing

- Formatter (Prettier) ran clean with no changes
- The single changed file is the call site identified in the issue's root cause analysis
- Re-entry summary behaviour for background events (agent completions, etc.) is unaffected since no other notification paths were modified

### DIFF
--- a/src/components/Worktree/NewWorktreeDialog.tsx
+++ b/src/components/Worktree/NewWorktreeDialog.tsx
@@ -463,13 +463,6 @@ export function NewWorktreeDialog({
         if (selectedIssue && assignWorktreeToSelf && currentUser) {
           try {
             await githubClient.assignIssue(rootPath, selectedIssue.number, currentUser);
-            notify({
-              type: "success",
-              title: "Issue Assigned",
-              message: `Issue #${selectedIssue.number} assigned to @${currentUser}`,
-              priority: "low",
-              countable: false,
-            });
           } catch (assignErr) {
             const message =
               assignErr instanceof Error ? assignErr.message : "Failed to assign issue";


gregpriday:
**Review status: Ready**

- Trivially correct — removes a 6-line success notification for a user-initiated action
- Error handling preserved

PR #3461, opened 2026-03-16
fix(window): resolve background color flash on startup

### LINKED ISSUE #3431: Window flashes dark background when light theme is active on startup
## Summary

The app window is created with a hardcoded `backgroundColor: "#19191a"` (dark) regardless of the user's saved theme preference. Users who have selected a light colour scheme see a dark flash before their theme loads, and the window briefly appears before React has finished mounting.

## Current Behavior

1. `BrowserWindow` is created with `backgroundColor: "#19191a"` and no `show: false`
2. The window is visible immediately, showing the dark background
3. React mounts asynchronously and fetches the theme via IPC
4. The theme CSS variables are applied, transitioning to the correct (possibly light) theme
5. For light-theme users, this produces a visible dark → light flash on every launch

The window is also shown before `ready-to-show` fires, meaning the app can appear partially rendered.

## Expected Behavior

The window background matches the user's saved theme from the moment it first appears. The window is not shown until the renderer is ready to paint, eliminating both the colour flash and any partially-rendered state.

## Root Cause Analysis

The `BrowserWindow` constructor hardcodes the background colour and does not defer window visibility until the first paint is ready.

**Evidence:**
- [`electron/window/createWindow.ts`](https://github.com/canopyide/canopy/blob/develop/electron/window/createWindow.ts#L65) — `backgroundColor: "#19191a"` is hardcoded; no `show: false` or `ready-to-show` handler
- [`electron/ipc/handlers/appTheme.ts`](https://github.com/canopyide/canopy/blob/develop/electron/ipc/handlers/appTheme.ts#L21) — Theme preference is read in the main process already; the saved scheme ID and its background colour are accessible here before window creation

The Windows title bar overlay at `createWindow.ts` also hardcodes `color: "#19191a"` and should match the theme.

## Acceptance Criteria

- [ ] The window's `backgroundColor` matches the background colour of the user's saved theme at the time of window creation
- [ ] Light-theme users see no dark flash on startup
- [ ] The window is not shown until the renderer is ready to paint (no partially-rendered state)
- [ ] The Windows title bar overlay colour matches the active theme
- [ ] Behaviour is correct for both the dark default and any custom colour schemes

## Edge Cases & Risks

- If there is no saved theme (first launch), the background should fall back to the OS dark/light mode preference via `nativeTheme.shouldUseDarkColors`
- A maximum show delay (e.g. 3 seconds) should be enforced so the app is never permanently hidden if the renderer fails to fire `ready-to-show`
- Custom colour schemes with non-standard background colours must be handled, not just the built-in dark/light pair

gregpriday:
## Summary

- Reads the user's saved `colorSchemeId` from the store at window creation time and resolves the correct background color before `BrowserWindow` is constructed
- Sets `show: false` on the window and defers visibility until `ready-to-show` fires, with a 3-second safety timeout to prevent the window from staying hidden if the renderer fails to paint
- Updates the Windows title bar overlay color to match the resolved theme color
- Falls back to `nativeTheme.shouldUseDarkColors` on first launch when no theme is saved, and handles custom color schemes with non-standard hex backgrounds

Resolves #3431

## Changes

- `electron/window/createWindow.ts`: added `resolveWindowBackground()` to look up the saved scheme, normalize custom scheme IDs, and extract the `--color-background` CSS variable value; wired into `BrowserWindow` constructor and title bar overlay; deferred `show` until `ready-to-show`

## Testing

- Typecheck, lint, and format all pass clean
- Verified logic covers dark default, light built-in, and custom schemes
- Edge cases covered: no saved theme (OS fallback), malformed hex values (falls back to theme default), renderer never firing `ready-to-show` (3s timeout forces show)

### DIFF
--- a/electron/window/createWindow.ts
+++ b/electron/window/createWindow.ts
@@ -64,7 +64,8 @@ export function setupBrowserWindow(dirname: string): CreateWindowResult {
   ) {
     try {
       const parsed = JSON.parse(themeConfig.customSchemes);
-      if (Array.isArray(parsed)) customSchemes = parsed.map((s: AppColorScheme) => normalizeAppColorScheme(s));
+      if (Array.isArray(parsed))
+        customSchemes = parsed.map((s: AppColorScheme) => normalizeAppColorScheme(s));
     } catch {
       // Malformed custom schemes — fall back to built-in only
     }


gregpriday:
**Review status: Ready**

PR #3509, opened 2026-03-17
fix(project): eliminate extra terminals on new project creation

### LINKED ISSUE #3498: Extra Claude Code terminal spawned on new project creation
## Summary

When creating a new project, two Claude Code terminals appear in the panel grid where there should be none. The first appears almost immediately, and the second is added shortly after. This is consistently reproducible on every new project creation.

## Current Behavior

On new project creation (via the UI project picker or CLI path argument), two Claude Code terminals are spawned in the new project's workspace. The sequence is observable:
1. One Claude Code terminal appears immediately when the project loads
2. A second Claude Code terminal is added a short time later

**Reproducibility**: Always — every new project creation exhibits this behavior.

## Expected Behavior

A brand new project should open with an empty panel grid (no terminals). The user should reach the empty state UI and choose to spawn panels themselves.

## Root Cause Analysis

The duplication likely stems from one or more of the following mechanisms interacting during the project creation + hydration sequence:

**1. Startup "default" terminal leaking into new projects**

At app startup, a plain terminal is always spawned with a fixed `"default"` ID and `cwd: os.homedir()`:

- [`electron/window/windowServices.ts`](https://github.com/canopyide/canopy/blob/main/electron/window/windowServices.ts#L487-L497) — spawns `DEFAULT_TERMINAL_ID = "default"` at startup, associated with the current project (or no project on first launch)

When a new project is then created, `onProjectSwitch(newProjectId)` sets `lastKnownProjectId` to the new project. The `TerminalRegistry` fallback logic then matches the "default" terminal (which has no `projectId` and whose `cwd: os.homedir()` has no filesystem match to any project) to the new project:

- [`electron/services/pty/TerminalRegistry.ts`](https://github.com/canopyide/canopy/blob/main/electron/services/pty/TerminalRegistry.ts#L265-L268) — if `info.projectId` is null and filesystem inference fails, falls back to `lastKnownProjectId === projectId`

**2. Orphaned terminal restoration**

During hydration, any backend terminal returned by `getForProject()` that is NOT in the saved `appState.terminals` is unconditionally added to the frontend store as an "orphaned" terminal:

- [`src/utils/stateHydration/index.ts`](https://github.com/canopyide/canopy/blob/main/src/utils/stateHydration/index.ts#L588-L649) — orphaned terminals are appended without a deduplication check against what may already be in the store from a prior hydration pass

For a brand new project, `appState.terminals` is empty, so ALL terminals matched by `getForProject()` are treated as orphaned and restored, regardless of their actual origin.

**3. Divergent project-switch paths**

`handleDirectoryOpen` in `menu.ts` sends `PROJECT_ON_SWITCH` directly to the renderer **without** going through `ProjectSwitchService`, bypassing the `ptyClient.onProjectSwitch()` cleanup and `resetAllStoresForProjectSwitch()`:

- [`electron/menu.ts`](https://github.com/canopyide/canopy/blob/main/electron/menu.ts#L316-L366) — sends `PROJECT_ON_SWITCH` directly
- [`electron/services/ProjectSwitchService.ts`](https://github.com/canopyide/canopy/blob/main/electron/services/ProjectSwitchService.ts#L22-L54) — the authoritative switch path

If both `handleDirectoryOpen` and `ProjectSwitchService` are both triggered for the same project open (e.g., CLI path argument at launch triggers `handleDirectoryOpen`, while the renderer UI also triggers a switch), `useProjectSwitchRehydration` could run twice on the same project:

- [`src/hooks/app/useProjectSwitchRehydration.ts`](https://github.com/canopyide/canopy/blob/main/src/hooks/app/useProjectSwitchRehydration.ts) — handles `PROJECT_ON_SWITCH` IPC events

The "one appears, then another shortly after" timing is consistent with two separate hydration passes completing at different times.

## Acceptance Criteria

- [ ] Creating a brand new project (no previous panel state) results in an empty panel grid
- [ ] No terminals from a previously active project appear in a newly created project
- [ ] Opening a project that previously had terminals correctly restores only those terminals (no extras)
- [ ] The behavior is consistent regardless of whether the project is opened via the UI picker, the menu, or a CLI path argument

gregpriday:
## Summary

- Routes `handleDirectoryOpen` through `ProjectSwitchService` so all project-open paths go through the same cleanup and reset sequence — closing the double-hydration window that caused a second terminal to appear
- Strengthens the orphan filter in state hydration so the `"default"` startup terminal (spawned at app launch with no project affiliation) is excluded from orphan restoration for brand-new projects, eliminating the first spurious terminal

Resolves #3498

## Changes

- `electron/menu.ts` — replaced direct `PROJECT_ON_SWITCH` send in `handleDirectoryOpen` with a call to `ProjectSwitchService.switchProject()`
- `electron/ipc/handlers/projectCrud.ts` — minor wiring adjustment for the unified switch path
- `electron/window/windowServices.ts` — exposes `ProjectSwitchService` reference so `menu.ts` can use it
- `electron/main.ts` — passes the service reference through startup wiring
- `electron/lifecycle/appLifecycle.ts` — thread `ProjectSwitchService` through lifecycle init
- `src/utils/stateHydration/index.ts` — filters out the default terminal from orphan restoration when the project has no saved state

## Testing

- TypeScript typecheck passes clean
- ESLint passes (warnings only, no errors — all pre-existing)
- Prettier reports no formatting changes
- Unit test in `electron/lifecycle/__tests__/appLifecycle.test.ts` updated to reflect new service wiring

### DIFF
--- a/src/utils/stateHydration/index.ts
+++ b/src/utils/stateHydration/index.ts
@@ -585,11 +585,13 @@ export async function hydrateAppState(
 
         if (!checkCurrent()) return;
 
-        // Restore any orphaned backend terminals not in saved state (append at end)
-        // Filter out the startup "default" terminal when it has no project attribution —
-        // it would be incorrectly added to new projects that have no saved state.
+        // Restore any orphaned backend terminals not in saved state (append at end).
+        // When no panels were saved (brand-new project), skip the startup "default"
+        // terminal — its projectId may have been backfilled by TerminalRegistry's
+        // lastKnownProjectId fallback, incorrectly attributing it to the new project.
+        const hasSavedPanels = appState.terminals && appState.terminals.length > 0;
         const orphanedTerminals = Array.from(backendTerminalMap.values()).filter(
-          (t) => !(t.id === "default" && !t.projectId)
+          (t) => !(t.id === "default" && !hasSavedPanels)
         );
         if (orphanedTerminals.length > 0) {
           logHydrationInfo(


gregpriday:
**Review status: Ready**

PR #3531, opened 2026-03-17
fix(onboarding): make getting started checklist steps clickable CTAs

### LINKED ISSUE #3516: Getting Started: Steps Not Linked
Image

I understand the intent but the value of this modal is lower than it could be by not linking the steps still to complete. Users will want to click on the step not completed to get it started. It's a to-do list AND a CTA list. Right now the CTA function is not being fulfilled.

gregpriday:
## Summary

- Each incomplete step in the Getting Started checklist now renders as a clickable button that triggers the relevant action (e.g. open settings for API key setup, open a terminal, create a worktree)
- Completed steps remain non-interactive as visual confirmations
- The checklist body receives the `inert` attribute when collapsed so screen readers and keyboard nav skip hidden items

Resolves #3516

## Changes

- `GettingStartedChecklist.tsx`: added per-step `action` callbacks, wrapped incomplete steps in `<button>` elements with accessible labels, applied `inert` to the collapsed body
- `GettingStartedChecklist.test.tsx`: new test file with 12 tests covering rendering, click dispatch, completed-step non-interactivity, collapse/expand, and `inert` behaviour

## Testing

- Unit tests added and passing (`npm run check` clean — 0 errors, 385 pre-existing warnings)
- Manual verification: each step button triggers the correct action; completed steps are not buttons; collapse hides content with `inert`

### DIFF
--- a/shared/types/actions.ts
+++ b/shared/types/actions.ts
@@ -147,6 +147,7 @@ export type ActionId =
   | "recipe.editor.openFromLayout"
   | "panel.focus"
   | "panel.focusIndex"
+  | "panel.palette"
   | "worktree.switchIndex"
   | "agent.launch"
   | "app.settings.openTab"


gregpriday:
**Review status: Needs attention**

**Issues:**

1. **`panel.palette` is not a member of `ActionId`** — The action `"panel.palette"` is registered in `src/services/actions/definitions/panelActions.ts` but is **missing from the `ActionId` union** in `shared/types/actions.ts`. Since `CHECKLIST_ITEMS` is typed with `actionId: ActionId`, this should produce a TypeScript error. Two options to fix:
   - Add `| "panel.palette"` to the `ActionId` union in `shared/types/actions.ts` (alongside the other panel actions).
   - Alternatively, consider using `"agent.launch"` instead, which already exists in the type and may be more semantically appropriate for "Launch an AI agent" (panel.palette opens the generic panel spawn palette, not specifically an agent).

2. **SVG icons inside buttons lack `aria-hidden`** — The checkmark circle and item icon (`<Icon>`) inside each `<button>` are decorative. While lucide-react icons generally handle this, explicitly adding `aria-hidden="true"` would be safer for screen reader cleanliness. Low priority.

**Looks good otherwise:**

- The `<button>` / `<div>` split based on completion state is clean and semantically correct.
- `inert` on the collapsed body is a nice accessibility touch — prevents keyboard focus on hidden items.
- `focus-visible` outline styling is consistent with the dismiss and collapse buttons.
- Test coverage is solid: rendering states, dispatch verification for each action, and isolation from parent callbacks.
- `void` on the async dispatch return is correct.

PR #3541, opened 2026-03-17
fix(urldetector): prefer last localhost URL instead of first

### LINKED ISSUE #3535: UrlDetector returns first detected URL instead of most recent
## Summary

When a terminal output chunk contains multiple localhost URLs, the dev server detector returns the first one found rather than the most recently emitted one. This causes the wrong URL to be surfaced in the dev server panel when a process prints successive URLs (e.g., a server that restarts and binds to a different port, or a tool that prints both an old and new address in the same output flush).

## Current Behavior

`UrlDetector.selectPreferredUrl` filters candidates for `localhost` and returns `localUrls[0]` — the first match in document order. Given output like:

```
Old server: http://localhost:3000  New server: http://localhost:4000
```

The detector returns `http://localhost:3000/` even though `http://localhost:4000` was printed most recently.

This behaviour is documented in the test suite as a known limitation with a TODO comment:

> `// TODO: Should prefer latest (4000) but currently returns first (3000)`

**Evidence:**
- [`electron/services/UrlDetector.ts#L32-L39`](https://github.com/canopyide/canopy/blob/develop/electron/services/UrlDetector.ts#L32-L39) — `selectPreferredUrl` returns `localUrls[0]`
- [`electron/services/__tests__/UrlDetector.test.ts#L142-L155`](https://github.com/canopyide/canopy/blob/develop/electron/services/__tests__/UrlDetector.test.ts#L142-L155) — the TODO test that documents the wrong behaviour

## Expected Behavior

When multiple localhost URLs appear in the same output chunk, the detector should return the most recently emitted one (last in document order). This is already the correct behaviour for the buffer-fallback path — it explicitly takes `bufferUrls[bufferUrls.length - 1]` — so the single-chunk path should be consistent with that.

## Root Cause Analysis

The buffer-fallback path (when the current chunk has no URLs) correctly takes the last URL from the accumulated buffer. The direct-chunk path passes all URLs to `selectPreferredUrl`, which then returns the first rather than the last localhost match. The two code paths are inconsistent.

**Evidence:**
- [`electron/services/UrlDetector.ts#L11-L30`](https://github.com/canopyide/canopy/blob/develop/electron/services/UrlDetector.ts#L11-L30) — the `scanOutput` method showing the inconsistency between buffer-fallback (`bufferUrls[bufferUrls.length - 1]`) and the direct-chunk path (passes all URLs to `selectPreferredUrl`)

## Acceptance Criteria
- [ ] When a single output chunk contains multiple localhost URLs, the most recently appearing one (last in document order) is returned
- [ ] The existing test for this behaviour (currently marked TODO) passes without modification to the expectation
- [ ] The buffer-fallback path behaviour is unchanged
- [ ] All existing `UrlDetector` tests continue to pass

gregpriday:
## Summary

- `UrlDetector.selectPreferredUrl` was returning `localUrls[0]` (first match) instead of the last, inconsistent with the buffer-fallback path which already takes the last URL
- Changed to return `localUrls[localUrls.length - 1]` so when multiple localhost URLs appear in the same output chunk, the most recently emitted one wins
- Removed the TODO comment in the test that documented the wrong behaviour; the test now passes as a correct assertion

Resolves #3535

## Changes

- `electron/services/UrlDetector.ts` — return last localhost URL instead of first in `selectPreferredUrl`
- `electron/services/__tests__/UrlDetector.test.ts` — remove TODO and update test description to reflect correct behaviour

## Testing

Unit tests updated and passing. The previously-skipped assertion (`should prefer latest (4000) but currently returns first (3000)`) now passes without changing the expectation — only the TODO comment was removed. All other `UrlDetector` tests continue to pass.

### DIFF
--- a/electron/services/__tests__/UrlDetector.test.ts
+++ b/electron/services/__tests__/UrlDetector.test.ts
@@ -63,6 +63,11 @@ describe("UrlDetector", () => {
         expect(result.url).toBe("http://localhost:3000/");
       });
 
+      it("prefers localhost even when it appears before a later non-localhost URL", () => {
+        const result = detector.scanOutput("http://localhost:3000 and http://127.0.0.1:4000", "");
+        expect(result.url).toBe("http://localhost:3000/");
+      });
+
       it("returns last URL if no localhost variant exists", () => {
         const result = detector.scanOutput("http://127.0.0.1:3000 and http://127.0.0.1:4000", "");
         expect(result.url).toBe("http://127.0.0.1:4000/");


gregpriday:
**Review status: Ready**

PR #3544, opened 2026-03-17
fix(terminal): stop directing indicator firing on navigation and control keys

### LINKED ISSUE #3538: directing indicator fires on navigation and terminal control keystrokes
## Summary

When a user is interacting with a `waiting` agent terminal, pressing arrow keys, Home/End, Page Up/Down, function keys, or Escape shows the `directing` (pencil) indicator — even though none of these are meaningful prompt input. The filter that guards this transition only blocks mouse and focus events, leaving all ANSI cursor/navigation sequences unhandled.

## Problem Statement

The `directing` state communicates: _"the user is typing a prompt for this agent."_ Navigation and control keystrokes don't match that intent:

- **Arrow keys** — scrolling through command history or moving the cursor shows `directing`
- **Home / End / Page Up / Page Down** — scrolling the terminal buffer shows `directing`
- **Escape** — which Claude Code uses to cancel/interrupt an operation — shows the pencil indicator and holds it for the full 2500ms debounce window. This is the worst case: the user cancels an action and the UI responds by showing them as still "actively directing" the agent.
- **Function keys** — F1–F12 and similar control sequences all pass through

The debounce timer resets on every qualifying keystroke. When a user navigates the terminal with multiple arrow presses, the `directing` state stays active far longer than warranted.

## Current Behavior

`isNonKeyboardInput` checks only for mouse reporting sequences (`\x1b[M`, `\x1b[<`, URXVT mouse) and focus events (`\x1b[I`, `\x1b[O`). Any other byte sequence — including all ANSI cursor movement codes — returns `false`, causing `onUserInput()` to fire and the terminal to show as `directing`.

**Evidence:**
- [`src/services/terminal/TerminalInstanceService.ts#L28-L34`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalInstanceService.ts#L28-L34) — `isNonKeyboardInput` only filters mouse/focus events; arrow keys and Escape are not covered
- [`src/services/terminal/TerminalInstanceService.ts#L588-L598`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalInstanceService.ts#L588-L598) — `onData` handler calls `onUserInput` for any data that passes the filter
- [`src/services/terminal/TerminalAgentStateController.ts#L42-L63`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalAgentStateController.ts#L42-L63) — `onUserInput` transitions to `directing` and sets the debounce timer

## Desired Behavior

The `directing` indicator should only appear when the user is constructing a prompt — printable characters, Backspace, and Delete. Navigation and control inputs should be invisible to the directing state machine.

Escape specifically should be treated as a hard state override: if the terminal is currently in `directing` state and Escape is pressed, the state should immediately revert to `waiting` without waiting for the debounce to expire.

## Acceptance Criteria

- [ ] Pressing ↑ ↓ ← → (arrow keys) in a `waiting` agent terminal does not trigger the `directing` indicator
- [ ] Pressing Home, End, Page Up, Page Down does not trigger `directing`
- [ ] Pressing F1–F12 or other function keys does not trigger `directing`
- [ ] Pressing Escape while in `directing` state immediately reverts to `waiting` without waiting for the debounce timer
- [ ] Pressing Escape in a `waiting` terminal does not trigger `directing` at all
- [ ] Typing printable characters, Backspace, and Delete continues to correctly trigger `directing`
- [ ] Ctrl+C and other Ctrl+key combinations do not trigger `directing`
- [ ] Existing tests for input detection continue to pass

## Edge Cases & Risks

- **Bracketed paste** (`\x1b[200~`…`\x1b[201~`): large pastes should still show `directing` since the user is actively submitting content. The paste sequence contains printable characters, so this should be handled naturally.
- **Shift+Enter for multi-line prompts** (Claude Code uses Kitty Keyboard Protocol `\x1b[13;2u`): this is continuation input, not submission — it should trigger `directing` but not begin the Enter→working transition.
- **xterm.js `onKey` vs `onData`**: the `onData` path is needed to forward raw bytes to the PTY (`terminalClient.write`). Any changes to how `onUserInput` is triggered must not affect the data forwarding path.

gregpriday:
## Summary

- Extended `isNonKeyboardInput` to filter all ANSI navigation sequences (arrow keys, Home/End, Page Up/Down, function keys F1-F24, Escape, and Ctrl+key combinations) so they no longer trigger the `directing` state transition
- Added a dedicated Escape handler in `TerminalAgentStateController` that immediately cancels the directing debounce and reverts to `waiting` when Escape is pressed during a `directing` state — no more 2500ms hold after cancelling an agent action
- Added modifier-bearing sequence handling so arrow keys with Shift/Ctrl/Alt modifiers (e.g. `\x1b[1;2A`) are also filtered correctly

Resolves #3538

## Changes

- `src/services/terminal/TerminalInstanceService.ts` — expanded `isNonKeyboardInput` with navigation/control sequence patterns; wired Escape byte to `onEscapeKey` callback
- `src/services/terminal/__tests__/isNonKeyboardInput.test.ts` — comprehensive tests for all new filter patterns
- `src/services/terminal/__tests__/TerminalAgentStateController.test.ts` — tests for Escape immediate-revert and Escape-in-waiting no-op behaviour

## Testing

Full unit test suite passes. Manually verified that arrow keys, Escape, and function keys no longer show the directing pencil indicator, while printable characters and Backspace continue to trigger it correctly.

### DIFF
--- a/src/services/terminal/TerminalInstanceService.ts
+++ b/src/services/terminal/TerminalInstanceService.ts
@@ -25,18 +25,18 @@ import { SCROLLBACK_BACKGROUND } from "@shared/config/scrollback";
 // eslint-disable-next-line no-control-regex
 const URXVT_MOUSE_RE = /^\x1b\[\d+;\d+;\d+M/;
 
-// Arrow keys (normal mode), Home, End, and other single-letter CSI finals
+// CSI navigation: arrows, Home, End, and modified F1–F4 (with optional ;modifier param)
 // eslint-disable-next-line no-control-regex
-const CSI_NAV_RE = /^\x1b\[[ABCDHF]$/;
+const CSI_NAV_RE = /^\x1b\[(1;\d+)?[ABCDHFPQRS]$/;
 
-// Application-mode arrows, Home/End, F1–F4 (SS3 prefix)
+// Application-mode arrows, Home/End, F1–F4 (SS3 prefix, unmodified only)
 // eslint-disable-next-line no-control-regex
 const SS3_NAV_RE = /^\x1bO[ABCDHFPQRS]$/;
 
 // Tilde-terminated navigation: Insert(2), Delete(3), PgUp(5), PgDn(6), F5–F12
-// Excludes bracketed paste markers (200~, 201~)
+// Includes optional ;modifier param. Excludes bracketed paste markers (200~, 201~)
 // eslint-disable-next-line no-control-regex
-const TILDE_NAV_RE = /^\x1b\[(2|3|5|6|15|17|18|19|20|21|23|24)~$/;
+const TILDE_NAV_RE = /^\x1b\[(2|3|5|6|15|17|18|19|20|21|23|24)(;\d+)?~$/;
 
 export function isNonKeyboardInput(data: string): boolean {
   // Mouse sequences
--- a/src/services/terminal/__tests__/isNonKeyboardInput.test.ts
+++ b/src/services/terminal/__tests__/isNonKeyboardInput.test.ts
@@ -86,6 +86,28 @@ describe("isNonKeyboardInput", () => {
     it("matches lone Escape", () => {
       expect(isNonKeyboardInput("\x1b")).toBe(true);
     });
+
+    it("matches modifier-bearing arrow keys (Shift+Up, Ctrl+Left, etc.)", () => {
+      expect(isNonKeyboardInput("\x1b[1;2A")).toBe(true); // Shift+Up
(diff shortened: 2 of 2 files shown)


gregpriday:
**Review status: Ready**

- Merge before PR #3545 (both touch `TerminalInstanceService.ts` and `TerminalAgentStateController.test.ts`)
- No semantic conflict with #3545 — #3544 handles Escape via `onData` bytes, #3545 handles Enter via `terminal.onKey()` DOM events; complementary approaches that work together correctly

PR #3546, opened 2026-03-17
fix(activity-monitor): prevent false idle during large non-polling output bursts

### LINKED ISSUE #3540: Non-polling terminal pattern buffer evicts working indicator during large output bursts
## Summary

In non-polling agent terminals, the working state debounce extension relies on `lastPatternResult.isWorking` being true at the time the debounce timer fires. When an agent produces a large block of output (a long code file, verbose log output, etc.), the rolling pattern buffer can fill entirely with response text, evicting the spinner/working indicator that was previously present. When the debounce timer fires and checks the buffer, it sees no working pattern and incorrectly transitions the terminal to `idle`.

## Problem Statement

The working-state debounce for non-polling terminals works as follows:

1. Agent output arrives → state becomes `busy` → debounce timer starts (2500ms)
2. On each new data chunk, the debounce timer is reset
3. When the debounce timer fires, before transitioning to `idle`, it checks `lastPatternResult?.isWorking`
4. If `isWorking` is true (agent spinner still visible in buffer), the debounce is extended indefinitely

The failure path:
1. Agent is working and showing "✽ Deliberating… (esc to interrupt · 15s)" 
2. Agent starts streaming a large response (thousands of characters)
3. `PatternBuffer` fills to its 2000-character limit; oldest content (including the spinner status line) is dropped
4. `lastPatternResult` is updated to `isWorking: false` because "esc to interrupt" is no longer in the buffer
5. Debounce timer fires and finds `isWorking: false` → agent transitions to `idle`/`waiting` while still actively streaming a response

This produces a false `waiting` state in the middle of an agent response — the agent appears to complete and start waiting for new input before it has actually finished.

**Evidence:**
- [`electron/services/ActivityMonitor.ts#L726-L758`](https://github.com/canopyide/canopy/blob/develop/electron/services/ActivityMonitor.ts#L726-L758) — `resetDebounceTimer` callback: if `lastPatternResult?.isWorking` is false, the debounce fires and state becomes `idle` regardless of whether output is still arriving
- [`electron/services/ActivityMonitor.ts#L152-L154`](https://github.com/canopyide/canopy/blob/develop/electron/services/ActivityMonitor.ts#L152-L154) — `PatternBuffer` initialized with 2000-character limit; large responses will overflow it
- [`electron/services/ActivityMonitor.ts#L319-L330`](https://github.com/canopyide/canopy/blob/develop/electron/services/ActivityMonitor.ts#L319-L330) — pattern detection runs on `patternBuf.getText()` (rolling buffer), not the full output stream

## Desired Behavior

The working-state debounce for non-polling terminals should remain active for as long as the terminal is receiving data, regardless of whether the rolling pattern buffer currently contains a working indicator. The pattern check is a useful supplementary signal for extending the hold beyond the last data chunk — but active data flow itself should always suppress the idle transition.

A terminal that received data 200ms ago and has active output is not idle, even if its pattern buffer currently shows only response text.

## Acceptance Criteria

- [ ] An agent streaming a response longer than ~2000 characters does not briefly show `waiting` in the middle of the response
- [ ] The working-state hold remains active for as long as data chunks are arriving, independent of pattern buffer state
- [ ] When output stops and no working pattern is present, the idle transition still fires correctly after the debounce window (existing behavior preserved)
- [ ] Polling-mode terminals (those with `getVisibleLines`) are unaffected — this fix is scoped to the non-polling debounce path
- [ ] Existing tests for `ActivityMonitor` continue to pass

## Edge Cases & Risks

- **Distinguishing streaming response from true completion**: The fix must not prevent the idle transition when the agent genuinely finishes. The debounce timer itself handles this — the fix should gate on "data arrived recently" rather than removing the pattern check entirely.
- **Polling vs non-polling**: Polling-mode terminals (`getVisibleLines` is set) already handle this correctly via grid-based detection in `runPollingCycle`. The fix applies only to the `resetDebounceTimer` path which is explicitly skipped when polling is enabled.
- **Memory**: The pattern buffer is bounded at 2000 chars for good reason. Expanding it is not necessarily the right solution — any fix should consider alternatives to simply increasing the buffer size.

gregpriday:
## Summary

- Tracks the timestamp of the most recent data chunk in `ActivityMonitor` for non-polling terminals
- Guards the debounce idle transition so it only fires when no data has arrived in the last 500ms, regardless of what the rolling pattern buffer currently contains
- Prevents false `waiting`/`idle` transitions mid-response when a large agent output (>2000 chars) evicts the working indicator from the pattern buffer

Resolves #3540

## Changes

- `electron/services/ActivityMonitor.ts`: added `lastDataReceivedAt` tracking; `resetDebounceTimer` now suppresses idle transition if data arrived within 500ms
- `electron/services/__tests__/ActivityMonitor.test.ts`: added 3 test cases covering the buffer-eviction false-idle scenario, data-flow suppression, and that idle fires correctly when output genuinely stops

## Testing

Unit tests pass — the new tests specifically exercise the buffer-eviction edge case and confirm the debounce suppression logic without affecting polling-mode or normal idle transition behaviour.

### DIFF
--- a/electron/services/ActivityMonitor.ts
+++ b/electron/services/ActivityMonitor.ts
@@ -755,8 +755,7 @@ export class ActivityMonitor {
       if (
         actuallyBusy === null &&
         this.lastWorkingIndicatorTimestamp > 0 &&
-        Date.now() - this.lastWorkingIndicatorTimestamp <
-          this.WORKING_INDICATOR_TTL_MS
+        Date.now() - this.lastWorkingIndicatorTimestamp < this.WORKING_INDICATOR_TTL_MS
       ) {
         this.resetDebounceTimer();
         return;


gregpriday:
**Review status: Ready**

PR #3575, opened 2026-03-18
feat(panel): direction-aware timing for panel transition animations

### LINKED ISSUE #3557: Panel transition animations use identical timing for expand and collapse
## Summary

Panel minimize-to-dock and restore-from-dock transitions both use the same 180ms duration and the same exponential ease-out curve (`cubic-bezier(0.16, 1, 0.3, 1)`). Expand and collapse are cognitively opposite actions — collapse should get out of the way fast while expand needs a moment for the eye to track the new focal point. Using the same curve for both makes minimize feel sticky and sluggish.

## Current Behavior

`PanelTransitionOverlay.tsx` defines a single constant:

```typescript
const ANIMATION_DURATION = 180; // ms
```

And applies one easing curve to all transitions:

```typescript
transitionTimingFunction: "cubic-bezier(0.16, 1, 0.3, 1)",
```

This curve is an exponential ease-out: it starts explosively fast and then decelerates heavily at the end. This is excellent for *expanding* panels (the fast start feels responsive, the slow finish lets the eye settle). For *collapsing* panels it is the wrong shape — the heavy deceleration at the end makes the panel visually linger as it shrinks toward the dock, making the UI feel like it is sticking.

The `getTerminalAnimationDuration()` utility in `animationUtils.ts` returns a single `TERMINAL_ANIMATION_DURATION = 150` constant, with no distinction between direction.

## Expected Behavior

Expand (restore from dock → grid) and collapse (minimize grid → dock) should use different timing profiles:

- **Collapse / minimize**: Shorter duration (~120ms) with an ease-in or standard ease curve. The panel should accelerate *away* from its origin so it visually disappears cleanly without lingering.
- **Expand / restore**: Slightly longer duration (~200–220ms) with the current ease-out curve. The fast start signals responsiveness; the gentle deceleration gives the eye time to track the panel as it arrives at its new position.

This asymmetry follows macOS system conventions (the "Scale" minimize effect is noticeably shorter than restore) and IBM's Carbon Design System's "productive motion" guidance.

## Context

- [`src/components/Panel/PanelTransitionOverlay.tsx#L26`](https://github.com/canopyide/canopy/blob/develop/src/components/Panel/PanelTransitionOverlay.tsx#L26) — single `ANIMATION_DURATION = 180` constant
- [`src/components/Panel/PanelTransitionOverlay.tsx#L160-L175`](https://github.com/canopyide/canopy/blob/develop/src/components/Panel/PanelTransitionOverlay.tsx#L160-L175) — single easing curve applied regardless of direction
- [`src/lib/animationUtils.ts`](https://github.com/canopyide/canopy/blob/develop/src/lib/animationUtils.ts) — `TERMINAL_ANIMATION_DURATION = 150`, no directional distinction
- [`src/hooks/usePanelHandlers.ts#L38`](https://github.com/canopyide/canopy/blob/develop/src/hooks/usePanelHandlers.ts#L38) — consumes `getTerminalAnimationDuration()` for calculating target rects

The `TransitionDirection` type (`"minimize" | "restore"`) already exists on `TransitionState`, so the direction is available at the point where the animation is applied.

## Acceptance Criteria

- [ ] Minimize-to-dock transitions use a shorter duration and an ease-in or standard ease curve, making the panel accelerate out of view
- [ ] Restore-from-dock transitions use a slightly longer duration and retain the current ease-out character
- [ ] The duration values are driven by the existing `prefers-reduced-motion` check — both directions return 0 when reduced motion is active
- [ ] The visual result: minimize feels snappy and immediate; restore feels responsive but gives the eye time to track

## Edge Cases & Risks

- `getTerminalAnimationDuration()` is used in `usePanelHandlers` to calculate timing for the layout update that follows the animation — if the collapse duration changes, the delay before layout reflow must match the new collapse duration to avoid a layout jump before the animation completes

gregpriday:
## Summary

- Collapse (minimize-to-dock) now uses 120ms with a standard ease curve so the panel accelerates out of view without lingering
- Expand (restore-from-dock) uses 220ms with the existing exponential ease-out, giving the eye time to track the panel to its new position
- `getTerminalAnimationDuration()` in `animationUtils.ts` accepts an optional direction parameter so callers can get the correct duration for layout timing; `usePanelHandlers` now passes `"minimize"` when computing the post-animation layout delay

Resolves #3557

## Changes

- `src/lib/animationUtils.ts` — added `COLLAPSE_DURATION = 120` / `EXPAND_DURATION = 220` constants; updated `getTerminalAnimationDuration(direction?)` to return direction-specific values
- `src/components/Panel/PanelTransitionOverlay.tsx` — replaced single `ANIMATION_DURATION` with direction-aware duration and easing; collapse uses `ease` curve, expand uses `cubic-bezier(0.16, 1, 0.3, 1)`
- `src/hooks/usePanelHandlers.ts` — passes `"minimize"` to `getTerminalAnimationDuration()` to match the new collapse duration
- Tests added for both `animationUtils` direction logic and `PanelTransitionOverlay` direction-aware styles

## Testing

- `npm run test` passes (new unit tests for `animationUtils.test.ts` and `PanelTransitionOverlay.test.tsx`)
- `npm run typecheck` passes clean
- `npm run fix` produced no changes

### DIFF
--- a/src/components/Panel/__tests__/PanelTransitionOverlay.test.tsx
+++ b/src/components/Panel/__tests__/PanelTransitionOverlay.test.tsx
@@ -60,7 +60,7 @@ describe("PanelTransitionOverlay", () => {
     expect(ghost.style.transitionTimingFunction).toBe("cubic-bezier(0.16, 1, 0.3, 1)");
   });
 
-  it("calls onTransitionComplete after minimize duration (120ms)", () => {
+  it("calls onTransitionComplete at exactly minimize duration (120ms)", () => {
     const onComplete = vi.fn();
     render(<PanelTransitionOverlay onTransitionComplete={onComplete} />);
 
@@ -68,15 +68,18 @@ describe("PanelTransitionOverlay", () => {
       triggerPanelTransition("panel-1", "minimize", sourceRect, targetRect);
     });
 
+    act(() => {
+      vi.advanceTimersByTime(119);
+    });
     expect(onComplete).not.toHaveBeenCalled();
 
     act(() => {
-      vi.advanceTimersByTime(120);
+      vi.advanceTimersByTime(1);
     });
     expect(onComplete).toHaveBeenCalledWith("panel-1");
   });
 
-  it("calls onTransitionComplete after restore duration (200ms)", () => {
+  it("calls onTransitionComplete at exactly restore duration (200ms)", () => {
     const onComplete = vi.fn();
     render(<PanelTransitionOverlay onTransitionComplete={onComplete} />);
 
@@ -85,12 +88,12 @@ describe("PanelTransitionOverlay", () => {
     });
 
     act(() => {
-      vi.advanceTimersByTime(120);
+      vi.advanceTimersByTime(199);
     });
     expect(onComplete).not.toHaveBeenCalled();
 
     act(() => {
-      vi.advanceTimersByTime(80);
+      vi.advanceTimersByTime(1);
     });
     expect(onComplete).toHaveBeenCalledWith("panel-1");
   });


gregpriday:

PR #3582, opened 2026-03-18
perf(worktree): use useDeferredValue for worktree list and event log

### LINKED ISSUE #3568: Use useDeferredValue for worktree list and event log rendering
## Summary

The worktree dashboard list and the event log both re-render eagerly on every incoming data update, blocking higher-priority interactions (terminal keystrokes, panel focus changes). React 19's `useDeferredValue` hook is a clean fit for both: it allows the underlying data to update instantly while deferring the expensive DOM derivation to idle time.

## Problem Statement

**Worktree list**: The active worktree is polled for git status every 2 seconds, and background worktrees every 10 seconds. Each poll result arrives as a Zustand store update that immediately triggers a full re-render of the worktree list — even if the user is actively typing in a terminal. All card components re-render synchronously on the render thread before the browser can process the next input event.

**Event log**: The event store holds up to 1,000 events (`MAX_EVENTS = 1000` in `src/store/eventStore.ts`). New events are continuously pushed during agent activity. Each `addEvent` call triggers an immediate re-render of the event log, which must map and filter up to 1,000 items to produce DOM output.

Both are non-interactive data displays. Neither needs to update synchronously with the underlying data change. A user interacting with a terminal should not experience increased input latency because a worktree poll just completed.

**Relevant code:**
- [`src/store/worktreeDataStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/worktreeDataStore.ts) — Zustand store that holds live worktree state; updates on every push from the workspace-host
- [`src/store/eventStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/eventStore.ts#L28) — `MAX_EVENTS = 1000`; `addEvent`/`addEvents` trigger immediate renders
- [`src/App.tsx`](https://github.com/canopyide/canopy/blob/develop/src/App.tsx#L145-L147) — `SidebarContent` reads worktrees directly from the store and passes them to the card list

## Desired Behavior

The worktree card list and the event log should always display accurate data, but their DOM re-derivation can yield to higher-priority work. When the renderer is idle (no pending user input), updates should appear with no perceptible delay. When the renderer is busy (user typing in a terminal, panel drag in progress), the display update should be deferred until the next idle frame.

React 19's `useDeferredValue` achieves this: the live store value updates immediately, but the value passed to the expensive rendering tree is deferred. This is strictly better than manual `setTimeout` debouncing because it is dynamically scheduled rather than applying a fixed artificial delay.

## Acceptance Criteria

- [ ] Worktree list re-renders are deferred: a git status update arriving during active terminal input does not block input processing
- [ ] Event log list re-renders are deferred: rapid event ingestion during agent activity does not cause input latency
- [ ] Displayed data is always consistent with the store — deferred value, not stale data, is shown
- [ ] React DevTools Profiler confirms lower interaction latency on the terminal grid during active worktree monitoring
- [ ] No visual artifacts (flicker, inconsistent state) during the deferred render window

## Edge Cases & Risks

- `useDeferredValue` must be applied to the value passed *into* the rendering tree, not to the Zustand selector itself. Deferring at the wrong level (e.g., inside the store selector) will not work and could produce stale renders.
- If the event log has auto-scroll-to-bottom behavior, the deferred render may cause the scroll position to jump after a brief delay. The auto-scroll logic should be reviewed to ensure it triggers after the deferred render commits, not immediately on data arrival.
- The worktree list uses `react-virtuoso` for virtualization in some views; confirm `useDeferredValue` interacts correctly with the virtualized list's scroll state.

gregpriday:
## Summary

- Wraps the worktree list and event log data in `useDeferredValue` so that git status poll results and agent event ingestion no longer block higher-priority work (terminal input, panel focus changes)
- Applied at the right level — `useDeferredValue` receives the store value and the deferred copy is passed into the expensive rendering subtree, keeping displayed data always consistent with the store
- `isPending` transitions in `App.tsx` use `useTransition` to give the worktree list a visual opacity hint while a deferred update is in flight

Resolves #3568

## Changes

- `src/App.tsx` — added `useTransition` + `useDeferredValue` for worktree list; `isPending` opacity applied to `SidebarContent`
- `src/components/Diagnostics/EventsContent.tsx` — added `useDeferredValue` wrapping the event log array before mapping to DOM

## Testing

- `npm run check` passes (typecheck, lint, format) after rebasing on latest develop
- No visual artifacts; deferred render is transparent to the user when the renderer is idle

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -543,7 +543,7 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
   }
 
   const rootPath = currentProject?.path ?? "";
-  const hasNonMainWorktrees = worktrees.length > 1;
+  const hasNonMainWorktrees = deferredWorktrees.length > 1;
   const hasFilters = hasActiveFilters();
   const worktreeMatchesQuery = (w: WorktreeState) => {
     if (!query) return true;
@@ -557,7 +557,7 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
   const integrationMatchesQuery = integrationWorktree && worktreeMatchesQuery(integrationWorktree);
   const visibleCount = hasFilters
     ? filteredWorktrees.length + (mainMatchesQuery ? 1 : 0) + (integrationMatchesQuery ? 1 : 0)
-    : worktrees.length;
+    : deferredWorktrees.length;
 
   const hasQuery = query.trim().length > 0;
   const isSortDisabled = isGroupedByType || hasQuery;
@@ -568,7 +568,7 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
       worktree={worktree}
       isActive={worktree.id === activeWorktreeId}
       isFocused={worktree.id === focusedWorktreeId}
-      isSingleWorktree={worktrees.length === 1}
+      isSingleWorktree={deferredWorktrees.length === 1}
       onSelect={() => selectWorktree(worktree.id)}
       onCopyTree={() => worktreeActions.handleCopyTree(worktree)}
       onOpenEditor={() => worktreeActions.handleOpenEditor(worktree)}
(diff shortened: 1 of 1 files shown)


gregpriday:

PR #3583, opened 2026-03-18
fix(terminal): suppress focus-acquiring clicks from propagating to xterm canvas

### LINKED ISSUE #3560: Clicks on unfocused terminal panels propagate to xterm.js canvas
## Summary

When a user clicks on an unfocused terminal panel to activate it, the click event propagates through the panel focus handler and reaches the xterm.js canvas. This causes unintended side effects: the click may place the cursor, trigger a URL open, select text, or perform word selection — all happening as part of what the user intended as a simple "focus this panel" gesture.

## Problem Statement

In a multi-panel grid, users frequently click on inactive panels to bring them into focus. The mental model is "activate this panel." But the click has a second, unintended effect: it also executes as a mouse action inside the terminal emulator.

This is the same problem Ghostty solved with explicit focus-transfer click suppression: when a click causes a focus transition, the pointer event should be consumed by the window manager layer and not forwarded to the terminal application. Ghostty documents this as preventing "accidental terminal inputs when users click to activate a minimized or unfocused panel."

Concrete failure modes in Canopy:

- User clicks on a non-focused agent terminal to check its output — the click lands on a URL in the terminal, opening the browser
- User double-clicks on a non-focused panel header to maximize it — the first click focuses and potentially triggers xterm word selection on the content
- User clicks near a running agent's prompt line — the click registers inside the terminal, potentially sending an accidental newline to the running process

## Current Behavior

`GridPanel.tsx` calls `handleFocus` on panel interaction (line 143), which sets `focusedId` in the store. There is no mechanism to suppress the downstream propagation of that same click event to the `TerminalPane` and its xterm.js canvas.

## Desired Behavior

When a pointer-down event causes a focus transition (i.e., the clicked panel was not already the focused panel), that event should be consumed at the focus layer and not forwarded to the xterm.js canvas. Subsequent events (pointer moves, pointer up) from the same interaction sequence should behave normally.

This should apply to single clicks only. Once a panel is already focused, all mouse events should pass through to xterm.js as normal.

## Context

- [`src/components/Terminal/GridPanel.tsx#L46`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/GridPanel.tsx#L46) — `usePanelHandlers` provides `handleFocus`
- [`src/components/Terminal/GridPanel.tsx#L143`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/GridPanel.tsx#L143) — `onFocus={handleFocus}` wired to the panel
- [`src/components/Terminal/TerminalPane.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/TerminalPane.tsx) — the xterm.js instance that receives the event
- [`src/store/slices/terminalFocusSlice.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/slices/terminalFocusSlice.ts) — `focusedId` state

## Acceptance Criteria

- [ ] Clicking on an unfocused grid terminal panel focuses it without delivering the click to xterm.js
- [ ] Clicking on an already-focused terminal passes all mouse events to xterm.js normally (word selection, URL clicks, etc. all work as expected)
- [ ] The first click after a focus transition does not trigger URL open, word selection, or cursor placement inside the terminal
- [ ] Dock panel popover interaction is unaffected

## Edge Cases & Risks

- xterm.js maintains its own pointer capture for selection and mouse reporting — care is needed not to break in-terminal mouse interactions after the panel is focused
- Maximized panels: if the panel is already focused but maximized, clicks on its content should still pass through normally
- The suppression should only apply to the focus-transition click, not to the general case of clicking while a transition animation is playing

gregpriday:
## Summary

- When a user clicks on an unfocused terminal panel, the click was propagating through the focus handler and landing inside the xterm.js canvas — triggering URL opens, word selection, cursor placement, or accidental newlines as a side effect of what should be a simple panel-focus gesture.
- Added a `captureClick` ref in `TerminalPane` that is set to `true` whenever the panel focus effect fires (indicating the panel was just focused). A `pointerdown` capture-phase handler on the xterm viewport element reads and clears the ref, calling `stopPropagation()` + `preventDefault()` to consume the first pointer event after a focus transition.
- Added `terminalFocus.ts` with the `useIsTerminalFocused` selector hook, and covered the suppression logic with unit tests in `focusClickSuppression.test.ts`.

Resolves #3560

## Changes

- `src/components/Terminal/TerminalPane.tsx` — wire `captureClick` ref to xterm viewport via capture-phase `pointerdown` handler in the focus effect
- `src/components/Terminal/terminalFocus.ts` — `useIsTerminalFocused` hook (selector over `terminalStore` focused ID)
- `src/components/Terminal/__tests__/focusClickSuppression.test.ts` — unit tests verifying suppression fires once after focus transition and passes through on subsequent clicks

## Testing

- Unit tests pass (`npm run check` clean, 0 errors)
- Suppression applies only to the first pointer event after a focus transition; all subsequent events pass through normally
- Already-focused panel clicks are unaffected

### DIFF
--- a/src/components/Terminal/TerminalPane.tsx
+++ b/src/components/Terminal/TerminalPane.tsx
@@ -435,7 +435,7 @@ function TerminalPaneComponent({
 
       const focusTarget = getTerminalFocusTarget({
         isAgentTerminal,
-        isInputDisabled: isBackendDisconnected || isBackendRecovering,
+        isInputDisabled: isBackendDisconnected || isBackendRecovering || isInputLocked,
         hybridInputEnabled,
         hybridInputAutoFocus,
       });
@@ -469,6 +469,7 @@ function TerminalPaneComponent({
       hybridInputAutoFocus,
       isBackendDisconnected,
       isBackendRecovering,
+      isInputLocked,
       isFocused,
       setFocused,
     ]


gregpriday:

- Merge sequence: merge before PR #3586 (both modify TerminalPane.tsx)

PR #3675, opened 2026-03-18
feat(compiler): adopt React Compiler in annotation mode

gregpriday:
## Summary

- Installs `babel-plugin-react-compiler` and `eslint-plugin-react-compiler`, configures Vite and ESLint for annotation-mode compilation
- Opts in `TerminalPane`, `ContentGrid`, and `WorktreeCard` with the `"use memo"` directive as the initial high-impact candidates
- Adds a ratchet baseline for react-compiler lint warnings so new violations are caught while existing ones are tracked

Resolves #3571

## Changes

- **`vite.config.ts`** — Added `babel-plugin-react-compiler` to the Vite React plugin with `compilationMode: "annotation"`
- **`eslint.config.js`** — Added `eslint-plugin-react-compiler` with `react-compiler/react-compiler` rule set to warn
- **`package.json`** — Added `babel-plugin-react-compiler` and `eslint-plugin-react-compiler` as devDependencies
- **`src/components/Terminal/TerminalPane.tsx`** — Added `"use memo"` directive
- **`src/components/Terminal/ContentGrid.tsx`** — Added `"use memo"` directive
- **`src/components/Worktree/WorktreeCard.tsx`** — Added `"use memo"` directive
- **`eslint-warnings-baseline.json`** — Updated ratchet baseline with react-compiler warning counts

## Testing

- Typecheck passes cleanly (`tsc --noEmit`)
- ESLint runs with 0 errors (warnings only, all pre-existing)
- Prettier formatting verified clean

PR #3676, opened 2026-03-18
refactor(store): eliminate active tab state duplication between focus and registry slices

### LINKED ISSUE #3605: Eliminate active tab state duplication between focus and registry slices
## Summary

The active tab for each tab group is stored in two places simultaneously: `activeTabByGroup` in the focus slice and `tabGroups[groupId].activeTabId` in the registry slice. A manual sync override keeps them in sync, but this creates race conditions during async persistence and requires callers to use the correct API entry point.

## Current State

The `setActiveTab` override at [`src/store/terminalStore.ts#L334-L376`](https://github.com/canopyide/canopy/blob/481a6adc/src/store/terminalStore.ts#L334-L376) performs dual updates:

1. **Line 341**: Updates `activeTabByGroup` in the focus slice
2. **Lines 365-366**: Updates `tabGroups[groupId].activeTabId` in the registry
3. **Lines 370-372**: Asynchronously persists via `saveTabGroups(get().tabGroups)`

**Problems with current approach:**
- **Race condition in persistence**: Line 370 uses `get().tabGroups` inside an async `import().then()` — if state changes between the `set()` call and the async persistence, stale data is saved
- **Bypass risk**: Any code calling `setState({ activeTabByGroup: ... })` directly skips the registry sync, creating split-brain state
- **Hydration complexity**: App startup must manually sync registry → focus layer. If this fails, the two copies diverge silently
- **Redundant validation**: The override re-validates group membership (lines 336-348) before updating — this check happens in both the override and the registry's own tab group methods

## Desired State

A single source of truth for the active tab in each group. Components that need to know the active tab read from one location. No manual sync logic. No async persistence race conditions.

## Context

- [`src/store/slices/terminalFocusSlice.ts`](https://github.com/canopyide/canopy/blob/481a6adc/src/store/slices/terminalFocusSlice.ts) — defines `activeTabByGroup` map in the focus slice
- [`src/store/slices/terminalRegistry/tabGroups.ts`](https://github.com/canopyide/canopy/blob/481a6adc/src/store/slices/terminalRegistry/tabGroups.ts) — defines `TabGroup.activeTabId` in the registry

The comment at line 332 explains the rationale: "This ensures the active tab selection is persisted and survives restart." The goal is valid — persistence — but the implementation creates duplication.

## Acceptance Criteria

- [ ] Active tab state exists in exactly one location (no dual storage)
- [ ] Active tab selection persists across app restarts
- [ ] Tab switching works correctly in all scenarios (click, keyboard, drag-drop)
- [ ] No race conditions during rapid tab switching
- [ ] `npm run check` passes

## Edge Cases & Risks

- **Virtual groups**: Line 343-346 handles "virtual groups" (no explicit TabGroup, just a single panel). The single source of truth must handle this case.
- **Components reading stale location**: Some components may read `activeTabByGroup` directly — all must be updated to read from the canonical location.
- **Persistence timing**: If the registry is the source of truth, persistence must happen synchronously (not in an async `import().then()`) to avoid saving stale state.

gregpriday:
## Summary

- Removes dual storage of active tab state (`activeTabByGroup` in focus slice + `tabGroups[groupId].activeTabId` in registry slice) and makes the registry the single source of truth
- Eliminates the manual sync override in `terminalStore.ts` that kept the two locations in sync, removing race conditions during async persistence
- Adds `getActiveTabId` selector to the registry slice that handles both explicit tab groups and virtual (single-panel) groups

Resolves #3605

## Changes

- **`src/store/slices/terminalRegistry/tabGroups.ts`**: Added `getActiveTabId` selector and `setActiveTabInGroup` action to the registry slice
- **`src/store/slices/terminalFocusSlice.ts`**: Removed `activeTabByGroup` map and replaced with delegation to registry
- **`src/store/terminalStore.ts`**: Removed the 40-line `setActiveTab` sync override and simplified store composition
- **`src/components/Layout/DockedTabGroup.tsx`** and **`src/components/Terminal/GridTabGroup.tsx`**: Updated to use `getActiveTabId` from registry
- **`src/hooks/useGridNavigation.ts`**: Updated active tab reads to use the new selector
- **Tests**: Expanded `activeTabTracking.test.ts` with coverage for virtual groups, background panels, rapid switching, and group removal. Updated other test files to remove references to the deleted `activeTabByGroup` state.

## Testing

- `npm run check` passes (typecheck + lint + format)
- All existing and new unit tests pass
- Active tab tracking test suite covers: basic tab switching, virtual groups, background panel exclusion, rapid switching, and group cleanup


gregpriday:

- Merge sequence: merge after PR #3670 (shared test file: actionDefinitions.adversarial.test.ts) and before PR #3675 (shared eslint-warnings-baseline.json)

PR #3712, opened 2026-03-18
feat(terminal): use xterm.js onWriteParsed for adaptive output backpressure

### LINKED ISSUE #3653: Use xterm.js onWriteParsed for adaptive output backpressure
## Summary
Terminal output ingestion uses fixed timer-based batching (8ms standard, 25ms for ink redraws). xterm.js 6.0 provides `onWriteParsed` which fires when the parser has caught up, enabling adaptive backpressure instead of fixed delays.

## Problem Statement
`TerminalOutputIngestService` buffers incoming IPC data and flushes it to xterm.js on fixed timers:
- 8ms for standard output (`IPC_STANDARD_FLUSH_DELAY_MS`)
- 25ms for ink redraw patterns (`IPC_INK_REDRAW_FLUSH_DELAY_MS`)

These fixed delays don't adapt to system load. On a fast machine, 8ms is unnecessarily slow. On a loaded machine with many terminals, 8ms may be too aggressive, causing the xterm.js parser to fall behind and queueing up work.

xterm.js 6.0's `onWriteParsed` event fires at most once per animation frame after the parser has processed all pending data. This enables a demand-driven flow: write data, wait for the parser to catch up, then write the next batch.

**Evidence:**
- [`src/services/terminal/TerminalOutputIngestService.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalOutputIngestService.ts#L9-L11) — fixed flush delays: 8ms and 25ms
- [`src/services/terminal/TerminalOutputIngestService.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalOutputIngestService.ts#L21-L29) — IPC buffer implementation with timeout-based flushing

## Desired Behavior
Instead of flushing on fixed timers, use xterm.js's `onWriteParsed` to know when the parser is ready for more data. This naturally adapts to system load:
- Fast machines: data flows as fast as the parser can handle
- Loaded machines: data is held back until the parser catches up
- Backpressure can propagate upstream: if the parser is behind, signal the PTY host to pause

## Acceptance Criteria
- [ ] Output flushing is driven by parser readiness rather than fixed timers
- [ ] High-throughput output (e.g., `cat` on a large file) renders smoothly without UI freezing
- [ ] Ink redraw detection still works correctly for real-time CLI tools
- [ ] No regression in output latency for interactive terminal usage (typing, command output)
- [ ] Backpressure acknowledgment to the backend still functions

## Edge Cases & Risks
- `onWriteParsed` fires at most once per frame — need to ensure this doesn't slow down low-throughput interactive use
- The ink redraw pattern detection may need to be preserved as a separate concern since those tools need different flush behavior
- Must maintain the existing `acknowledgeData` flow for backend backpressure

gregpriday:
## Summary

- Replaces fixed-timer output batching (8ms/25ms) with watermark-based flow control driven by xterm.js `onWriteParsed`
- Terminal output now adapts to actual parser throughput instead of using hardcoded delays
- Maintains ink redraw detection and backend backpressure acknowledgment

Resolves #3653

## Changes

- **TerminalOutputIngestService**: Rewrote flush logic around a high/low watermark system. Data accumulates in a buffer and flushes when the parser signals readiness via `onWriteParsed`. When the buffer exceeds a high watermark (128KB), upstream writes are paused; they resume when the buffer drains below the low watermark (32KB). Ink redraw detection is preserved as a separate concern.
- **TerminalInstanceService**: Exposes the `onWriteParsed` event from the xterm.js terminal instance so the ingest service can subscribe to parser readiness signals.
- **terminalOutput.worker.ts**: Minor adjustment to worker message types to align with the new flow control interface.
- **Tests**: Expanded test suite with coverage for watermark transitions, pause/resume signaling, parser-driven flush cycles, and the ink redraw path.

## Testing

- All existing and new unit tests pass (23 tests in the ingest service suite)
- TypeScript typecheck passes with no errors
- ESLint and Prettier report no issues

### DIFF
--- a/src/services/terminal/TerminalInstanceService.ts
+++ b/src/services/terminal/TerminalInstanceService.ts
@@ -230,6 +230,9 @@ class TerminalInstanceService {
 
     if (managed.isSerializedRestoreInProgress) {
       managed.deferredOutput.push(data);
+      const deferredBytes =
+        typeof data === "string" ? this.textEncoder.encode(data).length : data.byteLength;
+      this.dataBuffer.notifyWriteComplete(id, deferredBytes);
       return;
     }
 
--- a/src/services/terminal/TerminalOutputIngestService.ts
+++ b/src/services/terminal/TerminalOutputIngestService.ts
@@ -121,10 +121,8 @@ export class TerminalOutputIngestService {
   }
 
   public resetForTerminal(id: string): void {
-    if (!this.pollingActive || !this.worker) {
-      this.clearQueue(id);
-      return;
-    }
+    this.clearQueue(id);
+    if (!this.pollingActive || !this.worker) return;
     const message: WorkerInboundMessage = {
       type: "RESET_TERMINAL",
       id,
@@ -133,10 +131,8 @@ export class TerminalOutputIngestService {
   }
 
   public flushForTerminal(id: string): void {
-    if (!this.pollingActive || !this.worker) {
-      this.forceDrain(id);
-      return;
-    }
+    this.forceDrain(id);
+    if (!this.pollingActive || !this.worker) return;
     const message: WorkerInboundMessage = {
       type: "FLUSH_TERMINAL",
       id,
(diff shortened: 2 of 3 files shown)


gregpriday:

PR #3730, opened 2026-03-19
perf(worktreeDataStore): preserve Map identity in mergeFetchedWorktrees

### LINKED ISSUE #3689: Preserve Map identity in mergeFetchedWorktrees when no worktree changed
## Summary
`mergeFetchedWorktrees` always returns a **new `Map` instance** even when every worktree entry is identical to the existing one. This causes all 12+ consumers of `useWorktrees()` and `useWorktreeDataStore` to re-render on every git status poll cycle, despite no actual data having changed.

## Current Behavior
On every git status poll (every few seconds per worktree):

1. `mergeFetchedWorktrees` constructs a new Map at line 149: `new Map(fetchedStates.map(...))`
2. Individual worktree objects are correctly preserved via `worktreeStatesEqual` — if the fetched data matches the existing entry, the **existing object reference** is kept (line 172)
3. But the **Map container** is always a new reference, regardless of whether any entry changed
4. This new Map propagates to every `useWorktreeDataStore` subscriber:
   - `useWorktrees()` re-runs its `useMemo` sort (line 50), producing a new array
   - `SidebarContent` re-renders, React reconciles all `WorktreeCard` children
   - `ContentGrid`, `AgentButton`, `QuickSwitcher`, `ContentDock`, `useAgentLauncher`, and 6+ other consumers all re-render

With 5-10 worktrees polling every few seconds, this produces dozens of unnecessary re-render cascades per minute across the entire UI.

## Expected Behavior
When a poll returns data that is identical to what's already in the store, no subscriber should re-render. `mergeFetchedWorktrees` should detect when all entries in the result Map are reference-equal to the existing Map and return the **existing Map** unchanged.

## Root Cause Analysis
The function builds a fresh Map unconditionally at line 149, then merges metadata from the existing Map. Even though individual value references are preserved via `worktreeStatesEqual`, the Map constructor always creates a new container. There is no final check comparing the result Map against the input `existingWorktrees` before returning.

**Evidence:**
- [`src/store/worktreeDataStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/worktreeDataStore.ts#L144-L191) — `mergeFetchedWorktrees` function. Line 149 always creates new Map; line 172 preserves individual entries, but the container is new
- [`src/store/worktreeDataStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/worktreeDataStore.ts#L431-L449) — caller in `initialize` always sets `worktrees: map` (line 449), even if `map` is identical to `prev.worktrees`
- [`src/hooks/useWorktrees.ts`](https://github.com/canopyide/canopy/blob/develop/src/hooks/useWorktrees.ts#L20) — `useWorktreeDataStore((state) => state.worktrees)` subscribes to the Map directly without `useShallow`, so any new Map reference triggers re-render
- [`src/hooks/useWorktrees.ts`](https://github.com/canopyide/canopy/blob/develop/src/hooks/useWorktrees.ts#L50-L65) — `useMemo` depends on `effectiveMap`, so new Map → new sorted array → new return value → all consumers re-render

**Consumers affected** (all re-render on every poll):
- `src/App.tsx` (lines 328, 954) — SidebarContent, AppContent
- `src/components/Terminal/ContentGrid.tsx` (line 455)
- `src/components/Layout/ContentDock.tsx` (line 61)
- `src/hooks/useAgentLauncher.ts` (line 35)
- `src/hooks/useQuickSwitcher.ts` (line 60)
- `src/components/Layout/AgentButton.tsx` (line 28)
- `src/components/Project/QuickRun.tsx` (line 79)
- `src/hooks/app/useActiveWorktreeSync.ts` (line 8)
- `src/components/Terminal/TerminalContextMenu.tsx` (line 120)
- `src/components/Layout/TrashContainer.tsx` (line 47)
- `src/components/Project/ProjectSettingsDialog.tsx` (line 92)

## Scope & Constraints
**Not in scope:** Changing the `worktreeStatesEqual` comparison logic or the individual entry preservation (that already works correctly). Also not in scope: refactoring `useWorktrees()` consumers.

- Backwards compatibility: The Map's observable contents must be identical — only the container reference identity should change
- Performance: The identity check itself must be cheap (O(n) where n = number of worktrees, typically 5-10)

## Acceptance Criteria
- [ ] When a git status poll returns data identical to what's already stored, `useWorktreeDataStore`'s `worktrees` reference does not change
- [ ] `useWorktrees()` consumers do not re-render during no-op polls (verifiable via React DevTools Profiler or a `console.count` in `useWorktrees`)
- [ ] When a worktree's status genuinely changes, the Map is updated and consumers re-render as before
- [ ] When a worktree is added or removed, the Map is updated correctly

## Edge Cases & Risks
- A worktree being **removed** between polls means the new Map has fewer entries — this must still produce a new Map reference
- A worktree being **added** between polls means the new Map has more entries — same
- The `issueAssociations` override path (lines 176-188) can modify entries after the main merge — the identity check must account for this
- The callers at lines 435 and 487 both need the same fix

gregpriday:
## Summary

- Preserves Map reference identity in `mergeFetchedWorktrees` when no worktree data has actually changed, preventing unnecessary re-renders across 12+ consumers on every git status poll cycle
- Adds an identity check after merging: if the result Map has the same size and every entry is reference-equal to the existing Map, the original Map is returned unchanged
- Includes comprehensive tests covering no-op polls, genuine changes, additions, removals, and issue association overrides

Resolves #3689

## Changes

- `src/store/worktreeDataStore.ts`: Added identity comparison at the end of `mergeFetchedWorktrees` that returns the existing Map when all entries are reference-equal
- `src/store/__tests__/worktreeDataStore.refresh.test.ts`: New test suite with 6 tests validating Map identity preservation and correct update behavior

## Testing

- Typecheck passes cleanly (`tsc --noEmit`)
- ESLint passes (only pre-existing warning unrelated to changes)
- Prettier confirms files are already formatted
- All new tests pass and cover the key scenarios from the issue's acceptance criteria and edge cases

### DIFF
--- a/src/store/__tests__/worktreeDataStore.refresh.test.ts
+++ b/src/store/__tests__/worktreeDataStore.refresh.test.ts
@@ -223,6 +223,8 @@ describe("worktreeDataStore.refresh", () => {
     expect(mapAfter).not.toBe(mapBefore);
     // Unchanged entry should still preserve its individual reference.
     expect(mapAfter.get("main")).toBe(mainBefore);
+    // Changed entry should have a new reference.
+    expect(mapAfter.get("feature-changing")).not.toBe(mapBefore.get("feature-changing"));
   });
 
   it("returns new Map identity when a worktree is added", async () => {
@@ -247,6 +249,31 @@ describe("worktreeDataStore.refresh", () => {
     const mapAfter = useWorktreeDataStore.getState().worktrees;
     expect(mapAfter).not.toBe(mapBefore);
     expect(mapAfter.size).toBe(2);
+  });
+
+  it("returns new Map identity when a worktree is removed", async () => {
+    const main = createMockWorktree("main", { isMainWorktree: true, branch: "main" });
+    const feature = createMockWorktree("feature-gone");
+
+    getAllMock.mockResolvedValueOnce([main, feature]);
+    refreshMock.mockResolvedValue(undefined);
+
+    useWorktreeDataStore.getState().initialize();
+    await vi.waitFor(() => {
+      expect(useWorktreeDataStore.getState().isInitialized).toBe(true);
+    });
+
+    const mapBefore = useWorktreeDataStore.getState().worktrees;
+    expect(mapBefore.size).toBe(2);
+
+    // feature-gone disappears.
+    getAllMock.mockResolvedValueOnce([{ ...main }]);
+
+    await useWorktreeDataStore.getState().refresh();
+
+    const mapAfter = useWorktreeDataStore.getState().worktrees;
+    expect(mapAfter).not.toBe(mapBefore);
+    expect(mapAfter.size).toBe(1);
   });
 
   it("rejects stale onUpdate events arriving after a project switch (listenerGeneration guard)", async () => {


gregpriday:

PR #3746, opened 2026-03-19
feat(gpu-monitor): expand child-process-gone to non-GPU processes

### LINKED ISSUE #3700: Expand child-process-gone monitoring to PTY and workspace hosts
## Summary

The `app.on('child-process-gone')` listener in `GpuCrashMonitorService` filters to `details.type !== "GPU"`, ignoring PTY and workspace utility process crashes at the app level. This discards Electron's built-in crash classification (`reason: "oom" | "killed" | "crashed" | "abnormal-exit"`) which is more reliable than the exit-code heuristic currently used in PtyClient.

## Problem Statement

When the PTY host or workspace host utility process crashes, Canopy relies on instance-level `exit` event handlers to detect and classify the crash. `PtyClient.ts` has a `classifyCrash()` function that infers crash type from the exit code (e.g., code 137 → OOM, code 134 → assertion failure). This heuristic is imprecise — exit code 137 means SIGKILL, which may or may not be OOM.

Electron's `child-process-gone` event provides the crash `reason` field directly from Chromium's process watcher, which distinguishes `"oom"` from `"killed"` without guessing from exit codes. This data is currently discarded.

## Current Behavior

```typescript
// GpuCrashMonitorService.ts#L35-L36
app.on("child-process-gone", (_event, details) => {
  if (details.type !== "GPU") return;  // Everything non-GPU is ignored
  ...
});
```

The `details` object provides:
- `type`: `"GPU"` | `"Utility"` | etc.
- `reason`: `"oom"` | `"killed"` | `"crashed"` | `"clean-exit"` | `"abnormal-exit"` | `"memory-eviction"`
- `exitCode`: integer
- `serviceName`: string (can be set via `utilityProcess.fork(path, args, { serviceName: '...' })`)

For utility processes spawned with a `serviceName`, the crash can be identified precisely. Currently, Canopy's utility processes are not spawned with `serviceName`, so they would show as `"Node Utility Process"`.

## Desired Behavior

- `child-process-gone` events for non-GPU process types should be logged with full crash details (type, reason, exitCode, serviceName)
- PTY and workspace host utility processes should be spawned with distinct `serviceName` values so their crashes can be individually identified
- The crash `reason` from this event should be available alongside the existing PtyClient crash classification for improved diagnostics
- This is observability/telemetry — it should not interfere with the existing instance-level crash recovery logic in PtyClient and WorkspaceClient

## Context

**Relevant code:**
- [`electron/services/GpuCrashMonitorService.ts#L35-L42`](https://github.com/canopyide/canopy/blob/develop/electron/services/GpuCrashMonitorService.ts#L35-L42) — existing child-process-gone listener, GPU-only
- [`electron/services/PtyClient.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/PtyClient.ts) — PTY host crash classification uses exit code heuristic
- [`electron/services/WorkspaceClient.ts`](https://github.com/canopyide/canopy/blob/develop/electron/services/WorkspaceClient.ts) — workspace host crash handling

## Scope & Constraints

**Not in scope:**
- Changing existing PtyClient/WorkspaceClient crash recovery logic (they work correctly)
- Adding automatic recovery triggered by child-process-gone (the instance-level handlers already do this)
- External crash reporting or uploading

- Backwards compatibility: Purely additive — existing crash handling is preserved
- Performance: The event fires only on process termination, so there's no runtime overhead

## Acceptance Criteria

- [ ] Non-GPU `child-process-gone` events are logged with type, reason, exitCode, and serviceName
- [ ] PTY host and workspace host are spawned with distinct `serviceName` values
- [ ] The GpuCrashMonitorService GPU-specific behavior (crash counting, flag writing, relaunch) is not affected
- [ ] `npm run check` passes

## Edge Cases & Risks

- `child-process-gone` and `utilityProcess.on('exit')` fire nearly simultaneously but are not ordered — the new logging should not interfere with the instance-level exit handler's recovery logic
- If `serviceName` is not set on the fork options, the utility process appears as `"Node Utility Process"` — the fix must set it at the fork site
- The GPU crash monitor's relaunch logic (lines 44-57) must continue to work only for GPU crashes, not utility process crashes

gregpriday:
## Summary

- Expands the `child-process-gone` event listener in `GpuCrashMonitorService` to log non-GPU utility process crashes (PTY host, workspace host, etc.) instead of silently discarding them
- Filters out benign `clean-exit` and `killed` events to avoid noise, only logging actual crashes (oom, crashed, abnormal-exit, etc.)
- GPU-specific crash counting and relaunch logic is completely unaffected

Resolves #3700

## Changes

- `GpuCrashMonitorService.ts`: When `details.type !== "GPU"`, log a warning with type, reason, exitCode, and name for non-clean/non-killed exits, then return (preserving existing GPU-only relaunch behavior)
- `GpuCrashMonitorService.test.ts`: Added helper `emitChildProcessGone()`, replaced the old "ignores non-GPU" test with three targeted tests covering crash logging, clean-exit filtering, and full detail verification

## Testing

- `npm run check` passes (typecheck, lint, format)
- All existing and new unit tests pass

### DIFF
--- a/electron/services/__tests__/GpuCrashMonitorService.test.ts
+++ b/electron/services/__tests__/GpuCrashMonitorService.test.ts
@@ -153,11 +153,10 @@ describe("GpuCrashMonitorService", () => {
     it("logs non-GPU crash with full process details", async () => {
       await loadAndInit();
       emitChildProcessGone("Utility", "oom", 137, "canopy-workspace-host");
-      expect(console.warn).toHaveBeenCalledWith(expect.stringContaining("[ChildProcess]"));
-      expect(console.warn).toHaveBeenCalledWith(expect.stringContaining("reason=oom"));
-      expect(console.warn).toHaveBeenCalledWith(expect.stringContaining("exitCode=137"));
       expect(console.warn).toHaveBeenCalledWith(
-        expect.stringContaining("name=canopy-workspace-host")
+        expect.stringContaining(
+          "type=Utility, reason=oom, exitCode=137, name=canopy-workspace-host"
+        )
       );
     });
 


gregpriday:

PR #3931, opened 2026-03-19
feat(layout): respect grid capacity on undo and sync fit timing

### LINKED ISSUE #3836: Layout undo should respect grid capacity and fit timing
## Summary

Layout undo/redo can restore more panels to the grid than the current window size supports, and terminal fit timing doesn't align with CSS transition duration. Both issues cause layout corruption when the window has been resized between snapshot capture and restoration.

### Grid capacity bypass during undo

`applySnapshot()` validates that terminal IDs still exist but does not check whether the number of grid panels fits within the current grid capacity. The capacity function (`getMaxGridCapacity`) is called during `addTerminal()` and `moveTerminalToGrid()`, but `applySnapshot()` bypasses it entirely.

**Scenario:** User has 10 panels in a wide window → resizes to half width (capacity drops to 6) → hits undo → all 10 panels are restored to the grid. Panels render below the 380px minimum width and 200px minimum height, breaking terminal readability.

### Fit timing misaligned with CSS transition

The grid batch-fit effect fires after a 150ms timeout, but the CSS transition on `grid-template-columns` is 200ms. This means `fit()` runs 50ms before the transition completes, causing xterm to measure fractional pixel dimensions mid-animation. There is no `transitionend` listener anywhere in the codebase — all timing is timeout-based.

## Desired Behavior

1. When restoring a layout snapshot, grid panel count should be clamped to current grid capacity. Excess panels should overflow to the dock rather than violating minimum dimension constraints.

2. Terminal fit calls should be synchronized with CSS transition completion rather than using a fixed timeout that doesn't match the transition duration.

## Context

**Relevant code:**
- [`src/store/layoutUndoStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/layoutUndoStore.ts#L49-L93) — `applySnapshot()` restores layout without capacity validation (line 84 applies state directly)
- [`src/store/layoutUndoStore.ts`](https://github.com/canopyide/canopy/blob/develop/src/store/layoutUndoStore.ts#L56-L61) — only validation is checking that terminal IDs still exist
- [`src/lib/terminalLayout.ts`](https://github.com/canopyide/canopy/blob/develop/src/lib/terminalLayout.ts) — `getMaxGridCapacity()` exists but is not imported or called by layoutUndoStore
- [`src/components/Terminal/ContentGrid.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/ContentGrid.tsx#L791-L820) — batch-fit effect with 150ms timeout (line 813)
- [`src/components/Terminal/ContentGrid.tsx`](https://github.com/canopyide/canopy/blob/develop/src/components/Terminal/ContentGrid.tsx#L1056-L1064) — CSS transition is 200ms ease-out (line 1062)

## Scope & Constraints

**Not in scope:** Changing the undo/redo stack depth, modifying the snapshot data structure, or changing the grid capacity calculation algorithm.

- Backwards compatibility: Layout snapshots already captured must still work (the fix is at restoration time, not capture time)
- Performance: Capacity check should be negligible overhead — it's a simple arithmetic calculation

## Acceptance Criteria

- [ ] Undoing to a snapshot with more grid panels than current capacity auto-docks excess panels (by LRU or reverse order)
- [ ] Terminal fit does not fire before CSS grid transition completes
- [ ] Undo/redo still works correctly when window hasn't been resized (no regression)
- [ ] `npm run check` passes

## Edge Cases & Risks

- Panels auto-docked during capacity-limited undo should be re-dockable — the user should be able to undo the undo (redo) and get the original layout back if they resize the window larger again
- The redo snapshot should capture the adjusted layout (with some panels docked), not the original snapshot
- Two-pane split mode has its own fit timing (50ms timeout) — both paths need alignment

gregpriday:
## Summary

- Layout undo/redo now clamps grid panel count to current grid capacity, auto-docking excess panels instead of violating minimum dimension constraints
- Terminal fit timing aligned with CSS transition duration (200ms) so xterm measures final dimensions, not mid-animation fractional pixels
- Both the batch-fit effect and two-pane split fit path use synchronized timing

Resolves #3836

## Changes

- **`src/store/layoutUndoStore.ts`** — Added `clampSnapshotToCapacity()` that moves excess grid panels to the dock when restoring a snapshot. Handles both panels that existed at snapshot time and panels created after. Capacity check uses existing `getMaxGridCapacity()`.
- **`src/components/Terminal/ContentGrid.tsx`** — Updated batch-fit timeout from 150ms to 200ms and two-pane fit timeout from 50ms to 200ms to match the CSS `grid-template-columns` transition duration. Extracted `GRID_TRANSITION_MS` constant.
- **`src/lib/terminalLayout.ts`** — Exported `GRID_TRANSITION_MS = 200` constant for shared use.
- **`src/store/__tests__/layoutUndoStore.test.ts`** — New test suite covering capacity clamping on undo (excess panels docked, post-snapshot panels preserved, no-op when within capacity).
- **`src/lib/__tests__/terminalLayout.test.ts`** — Added test for `GRID_TRANSITION_MS` constant export.

## Testing

- All new unit tests pass (capacity clamping, post-snapshot terminal handling, no-regression for normal undo)
- TypeScript typecheck passes cleanly
- ESLint and Prettier pass with no new issues

### DIFF
--- a/src/store/__tests__/layoutUndoStore.test.ts
+++ b/src/store/__tests__/layoutUndoStore.test.ts
@@ -609,5 +609,30 @@ describe("layoutUndoStore", () => {
       const gridCount = state.terminals.filter((t) => t.location === "grid").length;
       expect(gridCount).toBe(10); // All fit within capacity 16
     });
+
+    it("post-snapshot grid terminals are clamped during undo", () => {
+      // Snapshot with t1 in grid
+      const t1 = makeTerminal({ id: "t1", location: "grid", worktreeId: "w1" });
+      seedTerminals([t1]);
+      useLayoutUndoStore.getState().pushLayoutSnapshot();
+
+      // After snapshot, add t2 and t3 to grid
+      const t2 = makeTerminal({ id: "t2", location: "grid", worktreeId: "w1" });
+      const t3 = makeTerminal({ id: "t3", location: "grid", worktreeId: "w1" });
+      useTerminalStore.setState({
+        terminals: [{ ...t1, location: "dock" }, t2, t3],
+      });
+
+      // Capacity = 2 → snapshot has t1 in grid, post-snapshot has t2+t3 in grid = 3 slots
+      useLayoutConfigStore.setState({ gridDimensions: { width: 800, height: 400 } });
+
+      useLayoutUndoStore.getState().undo();
+
+      const state = useTerminalStore.getState();
+      // t1 (from snapshot) gets priority, t2 is second, t3 overflows
+      expect(state.terminals.find((t) => t.id === "t1")?.location).toBe("grid");
+      expect(state.terminals.find((t) => t.id === "t2")?.location).toBe("grid");
+      expect(state.terminals.find((t) => t.id === "t3")?.location).toBe("dock");
+    });
   });
 });
--- a/src/store/layoutUndoStore.ts
+++ b/src/store/layoutUndoStore.ts
@@ -138,9 +138,19 @@ function applySnapshot(snapshot: LayoutSnapshot): boolean {
     }
   }
 
+  // Build combined entry list: snapshot entries first, then post-snapshot terminals.
+  // Post-snapshot terminals go at the end so they're overflowed first by capacity clamping.
+  const postSnapshotEntries: TerminalLayoutEntry[] = [];
+  for (const t of currentTerminals) {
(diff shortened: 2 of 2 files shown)


gregpriday:

PR #3980, opened 2026-03-20
fix(pty): remove backend UrlStyler to fix pasted URL highlighting

### LINKED ISSUE #3948: Pasted URLs get highlighted due to dual link handling
## Summary

When pasting a URL into a terminal prompt (e.g., an AI agent chat), the URL appears with blue underlined styling — as if it were a clickable link in program output. This is caused by the dual URL-handling approach: the backend `UrlStyler` wraps URLs in OSC 8 hyperlink sequences on **all** PTY output (including echoed user input), and the frontend `WebLinksAddon` independently detects URLs in the terminal buffer.

## Current Behavior

- Paste a URL like `https://example.com` into any terminal prompt
- The PTY echoes the input back through stdout
- `UrlStyler.styleUrls()` intercepts the echoed text and wraps it in OSC 8 hyperlink sequences with blue underline ANSI styling
- The URL appears highlighted (blue, underlined) in what should be plain user input
- `WebLinksAddon` may also independently detect the same URL, creating overlapping link handlers

This is visually confusing — the user typed/pasted the URL as input, but it looks like styled program output.

## Expected Behavior

- URLs pasted into terminal prompts should appear as plain text (matching the rest of the input)
- URLs in **program output** should remain clickable and styled
- Only one link detection mechanism should handle any given URL to avoid dueling click handlers, hover inconsistencies, and redundant CPU usage

## Root Cause Analysis

The system has two independent URL detection layers that were added/removed/re-added in quick succession, creating an inconsistent architecture:

| Layer | Mechanism | Scope | Problem |
|-------|-----------|-------|---------|
| **Backend** `UrlStyler` | Regex on raw PTY output → OSC 8 + ANSI styling | All stdout including echoed input | Cannot distinguish program output from echoed user input |
| **Frontend** `WebLinksAddon` | Regex on xterm.js buffer lines → link overlay | All visible text in terminal buffer | Detects URLs already handled by OSC 8 |

**History of the inconsistency:**
- **#3659** (`697c7807`) — Removed `WebLinksAddon` as "redundant" since `UrlStyler` handled URLs via OSC 8
- **#3820** (`5982c62c`) — Re-added `WebLinksAddon` because `UrlStyler` alone was fragile (missed URLs split across chunks, embedded in ANSI sequences, etc.)
- Result: both layers are now active simultaneously with no coordination between them

The core issue is that `UrlStyler` operates on **all** PTY stdout data indiscriminately — it has no way to distinguish echoed user input from genuine program output. When a user pastes a URL, the PTY echoes it back through stdout, `styleUrls()` wraps it in OSC 8 + blue underline, and the user sees their plain-text input rendered as a styled hyperlink.

Research confirms that using both OSC 8 and `WebLinksAddon` simultaneously causes: duplicate click handlers, hover/tooltip inconsistencies, and redundant CPU overhead from frontend regex evaluation on every render cycle.

**Evidence:**
- [`electron/services/pty/UrlStyler.ts#L60-L82`](https://github.com/canopyide/canopy/blob/develop/electron/services/pty/UrlStyler.ts#L60-L82) — `styleUrls()` applies to all text indiscriminately
- [`electron/services/pty/TerminalProcess.ts#L1298-L1306`](https://github.com/canopyide/canopy/blob/develop/electron/services/pty/TerminalProcess.ts#L1298-L1306) — `emitDataDirect()` calls `styleUrls()` on every PTY data emission
- [`src/services/terminal/TerminalAddonManager.ts#L40-L44`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalAddonManager.ts#L40-L44) — `WebLinksAddon` loaded on all terminals with `onLinkActivate` callback
- [`src/services/terminal/TerminalLinkHandler.ts`](https://github.com/canopyide/canopy/blob/develop/src/services/terminal/TerminalLinkHandler.ts) — Click handler used by both OSC 8 `linkHandler` option and `WebLinksAddon`

## Scope & Constraints

**Not in scope:**
- `FileLinksAddon` (file path links) — separate system, not affected
- Changing click behavior or the `TerminalLinkHandler` routing logic — that works correctly

**Backwards compatibility:** URLs in program output must remain clickable and styled. The fix should not regress #3820.

**Performance:** The chosen approach should avoid redundant regex evaluation on every render cycle across all terminal instances.

## Acceptance Criteria

- [ ] Pasting a URL into a terminal prompt does **not** cause it to appear highlighted/styled
- [ ] URLs in program output remain visually styled and clickable
- [ ] Only one link detection mechanism handles any given URL (no overlapping link providers)
- [ ] Modifier-click behavior for localhost URLs (open in browser panel) continues to work
- [ ] URLs split across PTY output chunks are still detected and clickable
- [ ] `npm run check` passes

## Edge Cases & Risks

- **Bracketed paste mode:** Most shells/programs use bracketed paste (`\x1b[?2004h`), which wraps pasted content in `\x1b[200~`...`\x1b[201~` markers. These markers appear in the PTY data stream and could be used to identify echoed paste content — but the echo itself may arrive in separate chunks from the markers.
- **Raw mode programs:** Programs like `vim` or `nano` disable echo, so this issue primarily affects canonical-mode prompts and AI agent REPLs.
- **ANSI-colored URLs:** Agent output (Claude, Gemini) often includes URLs surrounded by ANSI color codes. The solution must handle these reliably (#3820 was caused by `UrlStyler` failing on exactly this case).
- **Chunk splitting:** PTY output arrives in arbitrary chunks. A URL could be split across two `onData` callbacks, which is a known weakness of the backend regex approach.

## Dependencies

Related closed issues:
- #3659 — Removed WebLinksAddon (created the gap)
- #3820 — Re-added WebLinksAddon (created the dual approach)

gregpriday:
## Summary

- Removes the backend `UrlStyler` class that wrapped all PTY output URLs in OSC 8 hyperlink sequences, which incorrectly styled pasted URLs as clickable links
- The frontend `WebLinksAddon` already handles URL detection and click behavior in the terminal buffer, making the backend layer redundant and harmful
- Eliminates the dual link handling architecture that caused overlapping click handlers, hover inconsistencies, and redundant CPU usage

Resolves #3948

## Changes

- Deleted `electron/services/pty/UrlStyler.ts` and its test file `electron/services/pty/__tests__/UrlStyler.test.ts`
- Removed `UrlStyler` import and `styleUrls()` call from `TerminalProcess.emitDataDirect()` in `electron/services/pty/TerminalProcess.ts`

## Testing

- Typecheck passes cleanly (`npx tsc --noEmit`)
- ESLint passes with zero errors (only pre-existing warnings)
- Prettier formatting clean
- `WebLinksAddon` continues to handle URL detection and click routing on the frontend, including modifier-click for localhost URLs


gregpriday:

PR #4153, opened 2026-03-24
fix(agent): make error events no-op to eliminate false failed state

### LINKED ISSUE #4043: Agent failed state is unreliable and causes more harm than value
## Summary

The agent "failed" state triggers too easily and is unreliable as an indicator. It frequently shows an error state for agents that are actually functioning normally, causing user confusion and visual noise. Despite recent improvements in PR #4041 (recovery paths from failed state), the fundamental problem remains: the heuristic-based state detection produces false positives that erode trust in the indicator.

## Current Behavior

The agent state system has 7 states, with "failed" being one of the most visually prominent (second-highest priority after "working"). It triggers in two ways:

1. **Non-routine exit codes** — Any non-zero exit code that isn't a routine signal (SIGHUP, SIGINT, SIGPIPE, SIGTERM or their 128+ equivalents) causes `isRoutineExit()` to return false, transitioning to "failed"
2. **Explicit error events** — The `{ type: "error" }` event unconditionally transitions to "failed" from any state

The problems:
- Agents regularly exit with non-zero codes for benign reasons that aren't caught by `isRoutineExit()` (e.g., exit code 1 from normal agent completion)
- The error state gets the highest visual urgency treatment: red XCircle icon, `text-status-error` color, `panel-state-failed` border styling, and the error chip on worktree cards
- Once triggered, even though recovery paths exist (busy→working, prompt→waiting), the brief flash of error state is visually jarring and meaningless
- The state indicator has been observed showing error-like visuals even when the tooltip confirms the state is actually "waiting" — the state transitions faster than the UI can meaningfully communicate

**Evidence:**
- [`electron/services/AgentStateMachine.ts#L30-L32`](https://github.com/canopyide/canopy/blob/develop/electron/services/AgentStateMachine.ts#L30-L32) — Error event unconditionally returns "failed" from any state
- [`electron/services/AgentStateMachine.ts#L86-L88`](https://github.com/canopyide/canopy/blob/develop/electron/services/AgentStateMachine.ts#L86-L88) — Exit handler uses `isRoutineExit()` which only whitelists a small set of signals
- [`electron/services/pty/terminalForensics.ts#L23-L31`](https://github.com/canopyide/canopy/blob/develop/electron/services/pty/terminalForensics.ts#L23-L31) — Only exit code 0 and 4 specific signal codes are considered "routine"
- [`src/components/Worktree/terminalStateConfig.tsx#L41-L49`](https://github.com/canopyide/canopy/blob/develop/src/components/Worktree/terminalStateConfig.tsx#L41-L49) — Failed has second-highest display priority
- [`src/components/Worktree/utils/computeChipState.ts#L15`](https://github.com/canopyide/canopy/blob/develop/src/components/Worktree/utils/computeChipState.ts#L15) — Any failed terminal count triggers error chip on the worktree card

## Expected Behavior

The state system should reliably distinguish between genuine agent failures (actual crashes, unrecoverable errors) and normal lifecycle events (agent finishing work, user-initiated stops, non-zero but expected exit codes). If the system cannot reliably make this distinction — which the pattern of issues #4037, #3606, #3550, #3217 suggests — then the failed state should either be removed entirely or significantly reworked so it only activates with very high confidence.

The user should never see a red error indicator for an agent that is functioning normally.

## Root Cause Analysis

The core problem is that the "failed" state tries to infer error conditions from insufficient signals:

1. **Exit codes are unreliable indicators** — Many agents exit with code 1 for normal completion, timeout, or user cancellation. The whitelist approach in `isRoutineExit()` can never cover all legitimate exit codes across all agents.

2. **State machine is too aggressive** — The `error` event type unconditionally forces "failed" from any state, with no severity filtering or confidence threshold.

3. **Visual weight is disproportionate** — Even transient failed states get full error styling (red icon, error borders, worktree card chip), making brief false positives highly noticeable.

4. **History of fixes shows systemic issue** — The agent state system has required repeated stabilization (issues #3550, #3606, #3217, #4037), suggesting the failed state heuristics are fundamentally unreliable rather than needing incremental tuning.

## Scope & Constraints

**Not in scope:** Changes to the working/waiting/idle/completed state detection — those states are functioning well after recent stabilization work.

- Backwards compatibility: The failed state is internal — no external API depends on it
- Any solution must not lose genuine crash detection capability (actual PTY process crashes with SIGSEGV/SIGABRT should still be surfaced somehow)

## Acceptance Criteria

- [ ] Normal agent lifecycle (start, work, finish with any exit code) never shows an error indicator
- [ ] User-initiated kills and stops never trigger error state
- [ ] Genuine PTY crashes (SIGSEGV, SIGABRT) are still surfaced to the user in some form
- [ ] The worktree card chip does not show "error" for agents that completed their work normally
- [ ] No more transient flashes of error state during normal agent operation

## Edge Cases & Risks

- Some agents may genuinely crash and the user should know — completely removing the failed state risks hiding real problems
- The `TaskOrchestrator` uses the failed state to cancel running tasks — any removal needs to account for this dependency
- The `agent:failed` event is emitted for observability — downstream listeners may depend on it

## Dependencies

Builds on the work from PR #4041 (#4037). The previous fix made recovery from failed state possible but didn't address the root cause of false positives.

gregpriday:
## Summary

- Makes the `error` event a no-op in the agent state machine instead of unconditionally transitioning to "failed", eliminating false positive error indicators during normal agent operation
- Adds exit code 1 and SIGKILL (code 137) to the routine exit whitelist in `AgentStateService`, since agents commonly exit with these codes during normal lifecycle
- Preserves genuine crash detection for SIGSEGV/SIGABRT signals

Resolves #4043

## Changes

- `electron/services/AgentStateMachine.ts` — Remove the unconditional `error -> failed` transition; error events now return the current state unchanged
- `electron/services/pty/AgentStateService.ts` — Expand `isRoutineExit()` to treat exit code 1 and SIGKILL (137) as routine, preventing false failed states from normal agent completion
- `electron/services/__tests__/AgentStateMachine.test.ts` — Updated tests to verify error events are no-ops across all states
- `electron/services/pty/__tests__/AgentStateService.test.ts` — New tests covering exit code 1, SIGKILL, and genuine crash signals
- `electron/services/__tests__/AgentStateDetection.integration.test.ts` — Updated integration test expectations

## Testing

- All existing tests pass with updated expectations
- New unit tests cover the expanded routine exit logic
- TypeScript typecheck, ESLint, and Prettier all pass clean

### DIFF
--- a/electron/services/__tests__/AgentStateMachine.test.ts
+++ b/electron/services/__tests__/AgentStateMachine.test.ts
@@ -238,6 +238,10 @@ describe("AgentStateMachine", () => {
         expect(nextAgentState("waiting", { type: "exit", code: 143 })).toBe("completed");
       });
 
+      it("should transition waiting → failed on crash signal exit (SIGSEGV)", () => {
+        expect(nextAgentState("waiting", { type: "exit", code: 139 })).toBe("failed");
+      });
+
       it("should stay completed on non-crash exit from completed", () => {
         const event: AgentEvent = { type: "exit", code: 1 };
         expect(nextAgentState("completed", event)).toBe("completed");
@@ -245,6 +249,10 @@ describe("AgentStateMachine", () => {
 
       it("should transition working → failed on crash signal exit (SIGABRT)", () => {
         expect(nextAgentState("working", { type: "exit", code: 134 })).toBe("failed");
+      });
+
+      it("should transition completed → failed on crash signal exit (SIGABRT)", () => {
+        expect(nextAgentState("completed", { type: "exit", code: 134 })).toBe("failed");
       });
 
       it("should stay completed on zero exit from completed", () => {
--- a/electron/services/pty/AgentStateService.ts
+++ b/electron/services/pty/AgentStateService.ts
@@ -207,11 +207,9 @@ export class AgentStateService {
     // (crash-signal exits are now the only path to "failed")
     if (newState === "failed") {
       const errorMessage =
-        event.type === "error"
-          ? event.error
-          : event.type === "exit"
-            ? `Process crashed (exit code ${event.code}${event.signal ? `, signal ${event.signal}` : ""})`
-            : "Agent entered failed state";
+        event.type === "exit"
+          ? `Process crashed (exit code ${event.code}${event.signal ? `, signal ${event.signal}` : ""})`
+          : "Agent entered failed state";
       this.emitAgentFailed(terminal, errorMessage);
     }
 


gregpriday:
Fixes regression from #4041. Error events still triggered false failed-state positives despite the first fix attempt.

_Regression audit for training data._

Issue #4563, opened 2026-03-29
Integrated browser blocks cross-origin navigation — OAuth login flows impossible


jobordu:
## Summary

The integrated Canopy browser blocks all same-window cross-origin navigations. This makes OAuth/OIDC login flows (Google, GitHub, SAML, etc.) completely non-functional, since these require redirecting the user to an external identity provider and back.

## Environment

- Canopy: v0.5.3 (Electron 40.8.3, Chrome 144.0.7559.236)
- macOS (Darwin 25.3.0)

## Reproduction

1. Open any page in the integrated browser (e.g., `http://127.0.0.1:3000`)
2. Attempt to navigate to a cross-origin URL

**All of these fail silently (page stays where it is):**
- `<a href="https://example.com" target="_self">` — no navigation
- `window.location.href = 'https://example.com'` — no navigation
- `window.location.assign('https://example.com')` — no navigation
- `window.open('https://example.com', '_self')` — no navigation

**These open in the OS default browser (leave the webview):**
- `<a href="https://example.com" target="_blank">` — opens externally
- `window.open('https://example.com', '_blank')` — opens externally

## Impact

Any web application using OAuth/OIDC (the most common auth pattern) cannot authenticate through the integrated browser. The standard flow is:

1. App redirects to `https://auth-provider.example.com/authorize?...` (cross-origin)
2. User authenticates on the provider's page
3. Provider redirects back to `http://localhost:3000/auth/callback?code=...`

Step 1 is blocked because the webview won't navigate cross-origin.

This affects: Keycloak, Auth0, Okta, Firebase Auth, Supabase Auth, Google OAuth, GitHub OAuth, Azure AD, and virtually every identity provider.

## Expected behavior

Cross-origin navigations via `window.location.href`, `window.location.assign()`, and `<a target="_self">` should work within the integrated browser, same as a regular browser tab. The current behavior breaks the web platform's navigation model.

## Suggested approaches

1. **Allow cross-origin navigation in the webview** — The simplest fix. The webview should behave like a regular browser tab for navigation.

2. **Intercept and delegate OAuth flows** — Detect cross-origin navigations to known OAuth endpoints, open them in the system browser via `shell.openExternal()`, and capture the callback via a registered custom URL scheme or localhost redirect listener. This is the [Electron-recommended pattern for OAuth](https://www.electronjs.org/docs/latest/tutorial/oauth).

[...]

PR #4574, opened 2026-03-29
fix: preserve webview partition + flush storage before tab destruction

### LINKED ISSUE #4573: Webview partition not propagated in will-attach-webview handler + storage flush race on tab destruction
## Summary

Two independent bugs cause Web Storage (localStorage/sessionStorage) data loss in the integrated browser and portal panels, even when the correct `persist:` partition is set on the `<webview>` tag.

These are distinct from the cross-origin navigation blocking tracked in #4563 — as noted in [this comment](https://github.com/canopyide/canopy/issues/4563#issuecomment-4150107710), sessionStorage loss would surface as a separate hurdle even if cross-origin navigation were re-enabled.

## Bug 1: Partition not preserved in `will-attach-webview` handler

**File:** `electron/window/createWindow.ts`

The `will-attach-webview` handler validates `params.partition` against an allowlist, then hardens `webPreferences` (sandbox, contextIsolation, etc.) — but never sets `webPreferences.partition = params.partition`. In Electron 40, modifying the `webPreferences` object without explicitly setting the partition can cause the webview to silently fall back to an ephemeral (non-persistent) session instead of the intended `persist:browser` session.

**Impact:** All webview storage is ephemeral — `localStorage.setItem('key', 'value')` followed by a reload returns `null`.

## Bug 2: Storage flush race on WebContentsView destruction

**File:** `electron/services/PortalManager.ts`

`PortalManager.destroyView()` and `destroyHiddenTabs()` call `view.webContents.close()` synchronously without first flushing pending storage writes to disk. If `localStorage.setItem()` was called recently, the write may not have been persisted before the renderer process is terminated.

**Impact:** Portal tab storage writes can be lost during tab eviction or memory pressure cleanup.

## Related

- #4563 — Cross-origin navigation blocked in integrated browser (separate root cause)
- #4559 — sessionStorage loss during OAuth redirects (symptom)

jobordu:
## Summary

Fixes two independent bugs that cause Web Storage (localStorage/sessionStorage) data loss in the integrated browser and portal panels — even when the correct `persist:` partition is set on the `<webview>` tag.

As [noted in #4563](https://github.com/canopyide/canopy/issues/4563#issuecomment-4150107710), these storage persistence issues are a separate hurdle from the cross-origin navigation blocking.

Closes #4573

## Bug 1: Partition not preserved in `will-attach-webview`

The handler in `createWindow.ts` validates `params.partition` against an allowlist, then hardens `webPreferences` (sandbox, contextIsolation, etc.) — but never sets `webPreferences.partition = params.partition`. In Electron 40, modifying the `webPreferences` object without explicitly re-assigning the partition causes the webview to silently fall back to an ephemeral session.

**Fix:** One-liner — `webPreferences.partition = params.partition` after validation.

## Bug 2: Storage flush race on tab destruction

`PortalManager.destroyView()` and `destroyHiddenTabs()` call `view.webContents.close()` without first flushing pending storage writes. If `localStorage.setItem()` was called recently, the write may not reach disk before the renderer terminates.

**Fix:** Await `session.flushStorageData()` before `close()`. Note: Electron 40's type definitions declare `flushStorageData(): void` but it returns `Promise<void>` at runtime — we use `as unknown as Promise<void>` to work around this.

## Changes

| File | Change |
|------|--------|
| `electron/window/createWindow.ts` | Add `webPreferences.partition = params.partition` after validation |
| `electron/services/PortalManager.ts` | Await `session.flushStorageData()` before `close()` in `destroyView` and `destroyHiddenTabs` |
| `electron/window/windowServices.ts` | Await now-async `destroyHiddenTabs()` |
| `electron/window/__tests__/createWindow.webview-partition.test.ts` | 7 tests for partition preservation logic |

## Test plan

- [x] TypeScript typecheck passes (`npm run typecheck`)
- [x] ESLint passes with no new warnings (`npm run lint:ratchet`)
- [x] Existing PortalManager tests pass (65/65)
- [x] New partition preservation tests pass (7/7)
- [ ] Open a localhost dev server in the browser panel, set `localStorage.setItem('test', 'value')`, reload — value persists

### DIFF
--- a/electron/services/__tests__/PortalManager.test.ts
+++ b/electron/services/__tests__/PortalManager.test.ts
@@ -753,7 +753,10 @@ describe("PortalManager destroyHiddenTabs()", () => {
     let flushResolved = false;
     let closeCalledBeforeFlush = false;
 
-    const { promise, resolve } = Promise.withResolvers<void>();
+    let resolve!: () => void;
+    const promise = new Promise<void>((r) => {
+      resolve = r;
+    });
     wc.session.flushStorageData.mockReturnValue(promise);
     wc.close.mockImplementation(() => {
       if (!flushResolved) closeCalledBeforeFlush = true;


gregpriday (approved):
Reviewed, tested, and improved. Ready to merge.

PR #4656, opened 2026-03-30
perf(project-switch): inline terminal state in PROJECT_SWITCH IPC payload

### LINKED ISSUE #4650: Terminal state save is serialized on the project switch critical path
Project switching awaits two sequential IPC calls before `resetAllStoresForProjectSwitch()` can proceed:

```typescript
await projectClient
  .setTerminals(oldProjectId, terminalsToSave)
  .then(() => projectClient.setTerminalSizes(oldProjectId, terminalSizes))
```

These are disk writes to `project-state.json`. They're on the critical path specifically to prevent a race: `ProjectSwitchService.performSwitch()` calls `saveOutgoingProjectWorktreeState()` which does a read-modify-write of the same JSON file. If the renderer's writes haven't landed first, the main process overwrites them with stale data.

The race is real, but the serialization is the wrong fix. The correct fix is to eliminate the race entirely: pass `terminals` and `terminalSizes` as part of the `PROJECT_SWITCH` IPC payload. The main process handler receives them atomically and can apply them synchronously before its own state save. The two renderer awaits disappear from the critical path, saving 50–200ms.

**Evidence**

The awaited saves in `projectStore.switchProject()`:
https://github.com/canopyide/canopy/blob/075969bdcb76000eae871a5df39040bcb1662c18/src/store/projectStore.ts#L289-L340

`saveOutgoingProjectWorktreeState` in `ProjectSwitchService` — the conflicting write:
https://github.com/canopyide/canopy/blob/075969bdcb76000eae871a5df39040bcb1662c18/electron/services/ProjectSwitchService.ts#L129-L157

The same pattern exists in `projectStore.reopenProject()` and should be fixed there too.

gregpriday:
## Summary

- Eliminates two sequential awaited IPC writes (`setTerminals` + `setTerminalSizes`) from the project switch critical path, saving 50–200ms per switch
- `terminals` and `terminalSizes` are now passed inline as part of the `PROJECT_SWITCH` IPC payload so the main process handler applies them atomically before its own state save — no race condition, no serialization cost
- Same fix applied to `reopenProject()` which had the identical pattern

Resolves #4650

## Changes

- `shared/types/ipc/project.ts` — added `terminals` and `terminalSizes` fields to `ProjectSwitchPayload`
- `shared/types/ipc/api.ts` / `maps.ts` — updated IPC type definitions for new payload shape
- `electron/ipc/handlers/projectCrud.ts` — main process handler now receives and applies terminal state atomically from the payload
- `electron/ipc/handlers/terminalLayout.ts` — updated layout handler to account for the new flow
- `src/clients/projectClient.ts` — updated `switchProject` client to include terminal state in payload
- `src/store/projectStore.ts` — removed the two awaited pre-switch IPC calls; terminal state is bundled into the switch call instead
- `electron/preload.cts` — updated preload bridge for new payload shape
- `src/store/__tests__/projectStore.switching.test.ts` — updated tests to reflect the new call pattern

## Testing

- Unit tests updated and passing for the new bundled payload pattern
- No regression in existing project switch behavior — the main process handler still writes terminal state before calling `saveOutgoingProjectWorktreeState()`, preserving the ordering guarantee that originally motivated the serialized awaits

### DIFF
--- a/src/store/projectStore.ts
+++ b/src/store/projectStore.ts
@@ -160,6 +160,8 @@ function collectOutgoingTerminalState(): Required<ProjectSwitchOutgoingState> {
 
   const terminalSizes: Record<string, { cols: number; rows: number }> = {};
   for (const terminal of currentTerminals) {
+    if (terminal.location === "trash" || terminal.location === "background") continue;
+    if (isSmokeTestTerminalId(terminal.id)) continue;
     const instance = terminalInstanceService.get(terminal.id);
     if (instance) {
       terminalSizes[terminal.id] = {


gregpriday:
**⚠️ Regression Notice** — This PR introduced issues that required a follow-up fix:

- #4681 — fix(terminal): prevent phantom terminal on app restart

**What went wrong:** Inlining terminal state in the PROJECT_SWITCH IPC payload changed the project switch flow, but the unconditional default PTY terminal spawn on startup was not properly gated against the new hydration path. This caused a phantom terminal to appear on every app restart even when saved terminals were about to be restored.

_This comment is part of an automated regression audit for training data._

PR #4782, opened 2026-04-01
fix(demo): align frame naming between capture and encoder

### LINKED ISSUE #4774: Fix frame naming mismatch in demo capture pipeline
The demo capture-to-encode pipeline is broken end-to-end due to inconsistent frame filename formats.

**Capture** writes frames as `frame-000001.png` (hyphen separator, 6-digit zero-padded):
```typescript
const filename = `frame-${String(captureFrameCount + 1).padStart(6, "0")}.png`;
```

**Encoder** expects `frame_0001.png` (underscore separator, 4-digit zero-padded):
```typescript
const framePattern = /^frame_\d{4}\.png$/;
```

**ffmpeg input** uses `frame_%04d.png` (underscore, 4-digit):
```typescript
const inputPattern = path.join(framesDir, "frame_%04d.png");
```

This means calling `startCapture()` → `stopCapture()` → `encode()` always throws `"No PNG frames matching frame_NNNN.png found"` because the captured files never match the encoder's regex.

The unit test masks this by mocking `readdirSync` to return `["frame_0001.png", "frame_0002.png", "frame_0003.png"]` — matching the encoder pattern, not what capture actually produces.

All three locations need to use the same format. The ffmpeg `%06d` pattern should match whatever the capture writes. The test mock should reflect reality.

gregpriday:
## Summary

- The capture pipeline wrote frames as `frame-000001.png` (hyphen, 6-digit) while the encoder regex and ffmpeg pattern both expected `frame_0001.png` (underscore, 4-digit), so `encode()` always threw immediately after a real capture run
- Updated the encoder regex, ffmpeg input pattern, and the "no frames found" error message to all use `frame-NNNNNN.png` consistently
- Fixed the unit test mock (was returning `frame_0001.png` filenames, masking the bug) and added a regression test that confirms the old underscore format is correctly rejected

Resolves #4774

## Changes

- `electron/ipc/handlers/demo.ts`: updated `framePattern` regex to `/^frame-\d{6}\.png$/`, updated ffmpeg `-i` input to `frame-%06d.png`, updated error message text
- `electron/ipc/handlers/__tests__/demo.handlers.test.ts`: corrected mock filenames to `frame-000001.png` format, added assertion on ffmpeg `-i` argument, added regression test verifying old `frame_NNNN.png` names throw

## Testing

Unit tests pass with the corrected mocks. The regression test covers the previously-broken path and confirms the old format is rejected as expected.

### DIFF
--- a/electron/ipc/handlers/__tests__/demo.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/demo.handlers.test.ts
@@ -199,6 +199,26 @@ describe("registerDemoHandlers", () => {
       ).rejects.toThrow("No PNG frames matching frame-NNNNNN.png found");
     });
 
+    it("rejects old underscore-format frame names", async () => {
+      const fsMod = await import("fs");
+      (fsMod.readdirSync as ReturnType<typeof vi.fn>).mockReturnValueOnce([
+        "frame_0001.png",
+        "frame_0002.png",
+        "frame_0003.png",
+      ]);
+
+      const handler = getEncodeHandler();
+      const event = makeEvent();
+
+      await expect(
+        handler(event, {
+          framesDir: "/tmp/old-format",
+          outputPath: "/tmp/out.mp4",
+          preset: "youtube-4k",
+        })
+      ).rejects.toThrow("No PNG frames matching frame-NNNNNN.png found");
+    });
+
     it("rejects on spawn error event", async () => {
       const handler = getEncodeHandler();
       const event = makeEvent();


gregpriday:
**🔗 Regression Fix** — This PR fixes a regression introduced by #4785 (feat(demo): upgrade capture pipeline to raw bitmap piping with text-crisp encoding).

**Root cause:** The capture pipeline rewrite changed frame naming to `frame-000001.png` (hyphen, 6-digit) but the encoder still expected `frame_0001.png` (underscore, 4-digit). The naming contract wasn't kept in sync.

_This comment is part of an automated regression audit for training data._

PR #4980, opened 2026-04-05
feat(dev-preview): OAuth loopback flow for external auth redirects

### LINKED ISSUE #4892: Dev-preview: OAuth popup flow for external auth redirects
## Problem

Dev-preview panels block non-localhost navigations (intentionally, per #4563 and RFC 8252). When a dev-previewed app tries to redirect to an external OAuth provider (Keycloak, Auth0, Google, GitHub), the navigation is blocked. The user can click "Open in External Browser" but the OAuth callback lands in the system browser — the dev-preview webview never gets the session.

The deeper issue: the app's PKCE `code_verifier` is stored in the dev-preview webview's `sessionStorage`. Even if the system browser completes OAuth, it can't exchange the code because it doesn't have the verifier.

## Solution

When a blocked navigation from a dev-preview looks like an OAuth authorization URL, open it in an **isolated popup BrowserWindow** instead of the system browser. The popup:

1. Has its own Electron session (sandboxed, no node integration)
2. Can navigate freely to external IdP pages
3. When it navigates back to a localhost URL (the callback), Canopy captures the full URL
4. Canopy closes the popup and navigates the dev-preview webview to the captured callback URL
5. The webview has the PKCE state → token exchange succeeds → authenticated

This is NOT an embedded webview — it's a separate window. The IdP page runs in isolation from the app.

## Why not `shell.openExternal`?

The system browser doesn't share `sessionStorage` with the dev-preview webview. PKCE requires the `code_verifier` that was stored before the redirect — only the webview has it.

## Files

- `electron/services/OAuthPopupService.ts` — new service managing popup lifecycle
- `electron/setup/protocols.ts` — detect OAuth URLs in blocked navigations
- `electron/ipc/channels.ts` — new channel for OAuth popup
- `src/components/DevPreview/DevPreviewPane.tsx` — trigger popup instead of shell.openExternal

jobordu:
## Summary

Implements RFC 8252 loopback redirect handling so apps running in dev-preview can complete OAuth flows with external identity providers (Keycloak, Auth0, Google, GitHub) without losing PKCE state.

Closes #4892

## Problem

Dev-preview blocks non-localhost navigations. When a previewed app redirects to an external OAuth provider, the user can click "Open in External Browser" — but the callback lands in the system browser, which doesn't have the webview's `sessionStorage` containing the PKCE `code_verifier`. The token exchange fails.

## How it works

1. Dev-preview blocks external OAuth redirect → banner offers **"Sign in via Browser"**
2. Captures webview `sessionStorage` (PKCE verifier/state) before handoff
3. Starts ephemeral HTTP server on `127.0.0.1:0`, rewrites `redirect_uri`, opens system browser via `shell.openExternal`
4. User authenticates at IdP in system browser (RFC 8252 compliant — no embedded user-agent)
5. IdP redirects to loopback → server captures `code` + `state`, shuts down
6. Restores `sessionStorage` into webview via CDP `Page.addScriptToEvaluateOnNewDocument`
7. Navigates webview to original callback URL with captured params
8. CDP `Fetch` interceptor rewrites `redirect_uri` in the token exchange POST body to match the loopback URI (RFC 6749 §4.1.3 compliance)
9. App completes token exchange → authenticated inside dev-preview

## What changed

| File | Change |
|---|---|
| `shared/utils/urlUtils.ts` | `looksLikeOAuthUrl` heuristic (shared between main + renderer) |
| `electron/services/OAuthLoopbackService.ts` | Ephemeral loopback server + redirect_uri rewrite |
| `electron/ipc/handlers/webview.ts` | `handleOAuthLoopback` — CDP Fetch orchestration, sessionStorage restore |
| `electron/ipc/channels.ts` | `WEBVIEW_OAUTH_LOOPBACK` channel |
| `electron/preload.cts` | `startOAuthLoopback` bridge API |
| `shared/types/ipc/api.ts` | IPC type |
| `electron/setup/protocols.ts` | Routes `window.open` OAuth URLs through blocked-nav banner |
| `src/components/DevPreview/DevPreviewPane.tsx` | "Sign in via Browser" banner for OAuth redirects |

## Test evidence

- **14 unit tests** — URL detection heuristics, server lifecycle, callbacks, PKCE, state, timeouts, concurrency
- **5 integration tests** — real HTTP against loopback server, Keycloak-specific flow
- **26 protocol tests** — URL/protocol handling in `protocols.ts`
- **21 webview handler tests** — IPC handler coverage
- **1 full E2E test** — fake Keycloak + fake Next.js app, 3-layer assertions: protocol events from fake IdP, app-level `__oauthDebug` diagnostic state, UI token storage in webview `localStorage`

## Design decisions

- **Loopback server, not popup BrowserWindow** — per @gregpriday's [review](https://github.com/canopyide/canopy/issues/4892#issuecomment-gregpriday-2026-04-03): popup violates RFC 8252 §8.12, Google blocks `Electron/` user-agent
- **CDP Fetch interception** — rewrites `redirect_uri` in token exchange without modifying the user's app code
- **`window.open` coverage** — OAuth URLs from `window.open`/`target="_blank"` are routed through the banner, not straight to `shell.openExternal`
- **5-minute timeout** — server tears down if user abandons auth
- **One session per panel** — concurrent attempts cancel the previous one

## Limitation

The IdP must accept `http://127.0.0.1:*` as a valid redirect_uri (RFC 8252 §7.3). Supported by Keycloak, Auth0, Google, GitHub, Okta, and Azure AD for native/public clients.

### DIFF
--- a/electron/services/OAuthLoopbackService.ts
+++ b/electron/services/OAuthLoopbackService.ts
@@ -13,6 +13,15 @@ export { looksLikeOAuthUrl } from "../../shared/utils/urlUtils.js";
 const CALLBACK_PATH = "/oauth/callback";
 const TIMEOUT_MS = 300_000; // 5 minutes
 
+function escapeHtml(str: string): string {
+  return str
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#39;");
+}
+
 interface LoopbackSession {
   server: Server;
   timeout: NodeJS.Timeout;
@@ -85,8 +94,10 @@ export function startOAuthLoopback(
 
       // Respond to the system browser
       const title = hasError ? "Authentication Failed" : "Authentication Complete";
+      const rawDetail =
+        callbackParams.get("error_description") || callbackParams.get("error") || "unknown error";
       const message = hasError
-        ? `Authentication was not completed: ${callbackParams.get("error_description") || callbackParams.get("error")}`
+        ? `Authentication was not completed: ${escapeHtml(rawDetail)}`
         : "You can close this tab and return to Canopy.";
 
       res.writeHead(200, { "Content-Type": "text/html; charset=utf-8" });
--- a/electron/services/__tests__/OAuthLoopbackService.test.ts
+++ b/electron/services/__tests__/OAuthLoopbackService.test.ts
@@ -179,6 +179,33 @@ describe("OAuthLoopbackService", () => {
       expect(resultUrl.searchParams.get("error_description")).toBe("User cancelled");
     });
 
+    it("escapes HTML in error_description to prevent XSS", async () => {
+      const authUrl =
+        "https://auth.example.com/authorize?client_id=abc&response_type=code&redirect_uri=http://localhost:3000/auth/callback";
+
+      const loopbackPromise = startOAuthLoopback(authUrl, "test-panel");
+
+      await vi.waitFor(() => {
+        expect(shell.openExternal).toHaveBeenCalledTimes(1);
+      });
+
+      const rewrittenUrl = (shell.openExternal as ReturnType<typeof vi.fn>).mock
+        .calls[0][0] as string;
+      const newRedirectUri = new URL(rewrittenUrl).searchParams.get("redirect_uri")!;
+
+      const xssPayload = `<script>alert(1)</script><img onerror=alert(1) src=x>`;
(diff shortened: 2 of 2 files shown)


gregpriday (dismissed):
Thanks for the PR. The OAuth loopback flow is a well-designed solution to a genuinely tricky problem — using an ephemeral RFC 8252 loopback server and CDP Fetch interception to preserve PKCE state across system browser handoffs is the right approach.

On top of the base implementation, I've rebased onto current develop and run the full lint/format pass. The test coverage across unit, integration, protocol, webview handler, and E2E layers is thorough.

Spot on. Let's run with this.

gregpriday (approved):
Approving after rebase. Changes look good.

PR #5042, opened 2026-04-08
feat(setup): add skip-permissions toggle to agent setup wizard

### LINKED ISSUE #5026: Add skip-permissions toggle to agent setup wizard
The agent setup wizard (`AgentSetupWizard`) walks users through health check → agent selection → per-agent setup → complete, but never surfaces the "Skip Permissions" toggle. Users who want autonomous agent operation (a common workflow) have to finish onboarding, then navigate to Settings > Agents > [Agent] to enable it for each agent individually.

This toggle should be a prominent part of the per-agent setup step, since it fundamentally changes how the agent operates and many users will want it enabled from the start.

### What this should look like

During the per-agent setup step (the "agent" phase in `AgentSetupWizard`), each agent's screen should include a clearly visible toggle for skip-permissions mode. The toggle should:

- Show what CLI flag it enables (e.g. `--dangerously-skip-permissions` for Claude, `--yolo` for Gemini, `--dangerously-bypass-approvals-and-sandbox` for Codex, `--force` for Cursor)
- Include the same warning language used in the existing Settings UI — this is a dangerous mode and users should understand that
- Persist the choice via the existing `dangerousEnabled` field in agent settings
- Only appear for agents that have a default dangerous arg defined (skip for agents like OpenCode that have no equivalent)

### Why it matters

For power users who routinely run agents in autonomous mode, having to hunt through settings after onboarding is friction on first launch. The setup wizard is the natural place to make this decision — it's when users are actively configuring how their agents behave.

### Existing infrastructure

The `dangerousEnabled` / `dangerousArgs` mechanism already works end-to-end: the settings store, IPC handlers, flag generation (`generateAgentFlags`), and command construction all support it. The per-agent defaults are defined in `DEFAULT_DANGEROUS_ARGS` in `shared/types/agentSettings.ts`. This is purely a UI addition to the wizard — no backend changes needed.

gregpriday:
## Summary

- Adds a skip-permissions toggle to the per-agent setup step in `AgentSetupWizard`, so users can enable autonomous mode during onboarding rather than hunting through Settings afterwards
- Shows the relevant CLI flag chip (e.g. `--dangerously-skip-permissions`, `--yolo`) only when `dangerousEnabled` is on, matching the behaviour in the Settings UI
- Only surfaces the toggle for agents that have a `DEFAULT_DANGEROUS_ARGS` entry — agents without an equivalent flag are unaffected

Resolves #5026

## Changes

- `AgentSetupWizard.tsx`: added dangerous mode toggle with warning text in the agent setup phase; CLI flag chip now gated on `dangerousEnabled`
- `AgentSetupWizardDangerous.test.ts`: new test suite covering toggle render, label text, persistence via `setAgentSettings`, chip visibility, and agents with no dangerous args

## Testing

Unit tests added covering all the key scenarios. `npm run check` passes cleanly.

### DIFF
--- a/src/components/Setup/AgentCliStep.tsx
+++ b/src/components/Setup/AgentCliStep.tsx
@@ -13,6 +13,8 @@ import { AGENT_REGISTRY, getAgentConfig } from "@/config/agents";
 import { BUILT_IN_AGENT_IDS } from "@shared/config/agentIds";
 import { getInstallBlocksForCurrentOS, isBlockExecutable } from "@/lib/agentInstall";
 import { systemClient } from "@/clients";
+import { useAgentSettingsStore } from "@/store";
+import { DEFAULT_DANGEROUS_ARGS } from "@shared/types/agentSettings";
 import { CopyableCommand } from "./InstallBlock";
 import { AGENT_DESCRIPTIONS } from "./AgentSetupWizard";
 import type { CliAvailability } from "@shared/types";
@@ -172,6 +174,13 @@ export function AgentCliStep({ availability, selections, onInstallComplete }: Ag
     const status = cardStatuses[id];
     return status === "idle" || status === "error";
   });
+
+  const updateAgent = useAgentSettingsStore((s) => s.updateAgent);
+  const agentSettings = useAgentSettingsStore((s) => s.settings?.agents);
+
+  const agentsWithDangerousToggle = selectedAgentIds.filter(
+    (id) => (DEFAULT_DANGEROUS_ARGS[id] ?? "") !== ""
+  );
 
   return (
     <div className="flex flex-col gap-4">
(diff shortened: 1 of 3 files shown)


gregpriday:

This PR was reviewed as part of an automated merge pipeline and has issues that need to be addressed before merging:

- **Blocking dependency conflict with PRs #5037 and #5038:** Both upstream PRs eliminate the `{ type: "agent"; agentId: string }` step type from the wizard's `WizardStep` union. PR #5037 replaces it with `"system-tools" | "agent-cli" | "complete"`. After those PRs merge, `state.step.type === "agent"` in this PR will be a dead branch and `currentAgentId` will always be `undefined`, making `dangerousEnabled` always `false` and the toggle UI invisible.

- **Required action:** This PR must be rebased onto `develop` after both #5037 and #5038 land, and the skip-permissions toggle logic must be ported to `AgentCliStep.tsx` (the replacement component for per-agent configuration).

### Minor issues (can be fixed alongside)
- **AgentSetupWizard.tsx** — `updateAgent(currentAgentId!, ...)` uses a non-null assertion that could fail if the wizard step advances between render and async invocation. Add a guard: `if (!currentAgentId) return;`
- **AgentSetupWizard.tsx** — `void useAgentSettingsStore.getState().initialize()` in the reset effect is redundant (store is already idempotent).

These issues need manual attention — the PR has been excluded from this merge batch.

PR #5081, opened 2026-04-09
feat(theme): add command palette theme switching

### LINKED ISSUE #5069: Add command palette theme switching
Theme switching currently requires navigating to Settings > Appearance > App tab > clicking a theme. There's no keyboard-driven way to browse or switch themes without opening settings.

## Desired behavior

**Quick toggle** — a single keybinding (e.g. `Cmd+Shift+T`) that toggles between the user's preferred dark and preferred light theme. The store already tracks `preferredDarkSchemeId` and `preferredLightSchemeId`. The toggle should check the current theme's `type` field and swap to the opposite preference. Show a brief toast with the new theme name.

**Theme palette** — a dedicated keybinding (e.g. `Cmd+K Cmd+T`, matching VS Code convention) that opens a lightweight palette-style dropdown listing all available themes. Should support:

- Type-to-filter (same as existing `ThemeSelector` filter)
- Arrow key navigation with live preview on each focused item (temporarily apply the theme to the full app, revert on Escape)
- Enter to commit, Escape to revert to the original theme
- Color swatch indicators inline with each theme name (using the existing `PaletteStrip` / `APP_THEME_PREVIEW_KEYS` data)

This needs new action IDs registered in the action system (e.g. `app.theme.toggle`, `app.theme.pick`) with corresponding keybindings. The action palette (`Cmd+Shift+P`) should also surface these actions via search.

gregpriday:
## Summary

- Adds `app.theme.toggle` (toggles between preferred dark/light schemes based on current theme type, with valid-fallback logic and error toast on persistence failure) and `app.theme.pick` (opens a live-preview theme palette with arrow navigation, Enter to commit, Escape to revert)
- Mounts `ThemePalette` in `App.tsx` via `paletteStore` and the `canopy:open-theme-palette` event, following the same pattern as the notes palette
- Extracts a shared `PaletteStrip` component from `AppThemePicker` so both the settings picker and the command palette reuse the same swatch rendering

Resolves #5069

## Changes

- `src/services/actions/definitions/appActions.ts` — two new actions with keybindings and descriptions
- `src/components/ThemePalette/ThemePalette.tsx` — new palette component with live preview, filter, and keyboard navigation
- `src/components/ui/PaletteStrip.tsx` — extracted from `AppThemePicker`, now shared
- `src/components/Settings/AppThemePicker.tsx` — updated to use the shared `PaletteStrip`
- `shared/types/keymap.ts` — `Cmd+K T` chord keybinding added
- `src/hooks/app/useAppEventListeners.ts` — listens for `canopy:open-theme-palette` event
- `src/App.tsx` — mounts `ThemePalette` alongside other palette components

## Testing

7 unit tests covering toggle direction (dark-to-light, light-to-dark), no-op when already on correct type, wrong-type preferred-ID fallback to first built-in of correct type, persistence failure toast, pick action event dispatch, and action metadata shape. All passing. Typecheck, lint, and format clean.

### DIFF
--- a/shared/types/keymap.ts
+++ b/shared/types/keymap.ts
@@ -165,6 +165,8 @@ export type BuiltInKeyAction =
   | "help.launchAgent"
   | "help.togglePanel"
   | "app.settings"
+  | "app.theme.toggle"
+  | "app.theme.pick"
 
   // Voice input
   | "voiceInput.toggle"
@@ -314,6 +316,8 @@ export const KEY_ACTION_VALUES: ReadonlySet<string> = new Set<string>([
   "help.launchAgent",
   "help.togglePanel",
   "app.settings",
+  "app.theme.toggle",
+  "app.theme.pick",
   "voiceInput.toggle",
   "layout.undo",
   "layout.redo",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -92,6 +92,7 @@ import { BulkCommandPalette } from "./components/BulkCommandCenter";
 import { ConfirmDialog } from "./components/ui/ConfirmDialog";
 import { PanelLimitConfirmDialog } from "./components/Terminal/PanelLimitConfirmDialog";
 import { NotesPalette } from "./components/Notes";
+import { ThemePalette } from "./components/ThemePalette";
 
 function preloadSettingsDialog() {
   return import("./components/Settings/SettingsDialog");
@@ -254,6 +255,7 @@ function App() {
   }, [isSettingsOpen]);
   const [isShortcutsOpen, setIsShortcutsOpen] = useState(false);
   const isNotesPaletteOpen = usePaletteStore((state) => state.activePaletteId === "notes");
+  const isThemePaletteOpen = usePaletteStore((state) => state.activePaletteId === "theme");
   const {
     isWorktreeOverviewOpen,
     toggleWorktreeOverview,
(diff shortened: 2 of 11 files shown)


gregpriday:

Flagged during an automated merge review. A couple of UX regressions need attention before this PR can merge:

### Palette opens on the wrong theme

**`src/components/ThemePalette/ThemePalette.tsx:91-98`** — `selectedIndex` is never seeded to the currently active theme on open. `useSearchablePalette` defaults it to `0`, so the live-preview effect at lines 103-108 immediately injects `results[0]` (the first built-in scheme) the moment the palette opens. Two visible consequences:

1. A brief "flash" to the wrong theme every time the palette opens.
2. Pressing Enter without first arrowing through the list commits `results[0]` — so a user who opens the palette and hits Enter expecting a no-op will silently switch away from their current theme.

A fix is to seed `selectedIndex` from `results.findIndex(s => s.id === selectedSchemeId)` in an effect that runs when `isOpen` flips to true. This would also fix the related "stale selectedIndex on reopen" issue (the palette remembers the previous arrow position across open/close cycles because `useSearchablePalette.close()` is never called in the close path).

### Silent persistence failures

**`src/components/ThemePalette/ThemePalette.tsx:117-127`** — `commit()` swallows `appThemeClient.setColorScheme` failures with just `console.error`. The existing toggle action shows a toast on the same failure; the pick path should match, otherwise the user's selection appears to work in-session but reverts on next restart with no explanation.

### Minor — test gap

**`src/services/actions/definitions/__tests__/themeActions.test.ts:116-149`** — the "no-op when resolved target equals current scheme" test admits in its own comments that it cannot construct a true no-op case and ends up retesting the fallback path. The underlying `if (target.id === selectedSchemeId) return;` branch in `appActions.ts:142` is still uncovered. Not blocking, but worth cleaning up.

Excluding this PR from the current merge batch — the two significant items above need a fix pass before it can ship. Happy to pick it back up in the next cycle once the palette seeds its initial index and surfaces persistence errors.

Issue #5097, opened 2026-04-09
PTY pool stays at homedir and relies on fragile cd fixup


gregpriday:
The PTY pool is warmed once at `os.homedir()` during pty-host boot and never moves. Every pooled acquire then depends on a single fire-and-forget `cd` write to relocate the shell to the project path before handing it off. This is the likely root cause of #5091, and it has two separate fragility issues that both deserve to go away.

### 1. `setDefaultCwd` is dead code

[`PtyPool.setDefaultCwd()`](https://github.com/canopyide/canopy/blob/ece0d160188773f4e664969471f84c744bf8b460/electron/services/PtyPool.ts#L190-L197) exists but is called from nowhere in the codebase (only from its own tests). The pool is constructed in [`pty-host.ts`](https://github.com/canopyide/canopy/blob/ece0d160188773f4e664969471f84c744bf8b460/electron/pty-host.ts#L1549-L1560) with `warmPool(homedir)` and never reconfigured when a project activates. So refills forever use `homedir` as the spawn cwd, and every pooled acquire has to do the `cd` dance.

### 2. The `cd` fixup can be hijacked by the user's shell

[`acquirePtyProcess`](https://github.com/canopyide/canopy/blob/ece0d160188773f4e664969471f84c744bf8b460/electron/services/pty/terminalSpawn.ts#L138-L149) writes a bare `cd` command to the pooled shell:

```ts
pooledPty.write(
  `cd "${options.cwd.replace(/"/g, '\\"')}" && printf '\033[H\033[2J\033[3J'\r`
);
```

Because it's just `cd`, any user-defined `cd` function, alias, or plugin in the user's `.zshrc` / `.bashrc` runs instead of the builtin. Common culprits: `zoxide` / `z` / `autojump`, `direnv` hooks, `oh-my-zsh` `chpwd_functions`, or hand-rolled directory-toggle functions. If the user's `cd` override does anything stateful, the fixup's behavior becomes unpredictable — and #5091's strict parity alternation (project, home, project, home...) is exactly the shape you'd expect from a toggling `cd` function.

Even setting that theory aside, the fixup is racy on its own: there's no handshake confirming the shell is in REPL mode, no confirmation the `cd` actually executed, no retry, and a timing gap between disposing the pool's drain listener and `TerminalProcess` attaching its real `onData`.

### Desired behavior

[...]

PR #5100, opened 2026-04-09
fix(terminal): propagate project cwd to pooled pty refills

### LINKED ISSUE #5091: Terminals alternate between project path and home directory on spawn
Opening terminals in sequence alternates the working directory: the first terminal opens in the project path (correct), the second opens in the user's home directory, the third in the project path, the fourth in home, and so on. Every other spawn lands in `~` instead of the project root.

This is reproducible across repeated opens — it's not a one-off race, it's a consistent parity pattern, which points at something stateful in the PTY spawn path rather than an intermittent timing bug.

### Likely suspect: the PTY pool

[`PtyPool`](https://github.com/canopyide/canopy/blob/ece0d160188773f4e664969471f84c744bf8b460/electron/services/PtyPool.ts#L1-L264) pre-warms shells at startup with `os.homedir()` as the cwd (see `pty-host.ts` calling `warmPool(homedir)`). On `acquire()`, the caller in [`terminalSpawn.ts`](https://github.com/canopyide/canopy/blob/ece0d160188773f4e664969471f84c744bf8b460/electron/services/pty/terminalSpawn.ts#L126-L156) writes a `cd "<projectCwd>" && clear` command to the pooled shell to relocate it before handing it off:

```ts
// acquirePtyProcess — pooled branch
pooledPty.write(
  `cd "${options.cwd.replace(/"/g, '\\"')}" && printf '\033[H\033[2J\033[3J'\r`
);
```

The default pool size is 2, and the pool refills eagerly after each acquire. The alternation pattern lines up suspiciously well with a pool-based path where the `cd` fixup succeeds for some pooled shells but not others — e.g., a shell that hasn't fully initialized by the time the `cd` is written, or a refill entry that races with the acquire. The fresh-spawn fallback (`pty.spawn` with `cwd: options.cwd`) would always land in the correct directory, which matches the "every other one works" symptom.

### Expected

Every terminal opened while a project is active should start in that project's path, regardless of whether the PTY came from the pool or was spawned fresh.

### Reproduction

1. Open the app on any project
2. Open a new terminal panel — lands in project path
3. Open another — lands in `~`
4. Repeat — alternates project, home, project, home

gregpriday:
## Summary

- `PtyPool` pre-warms shells using `os.homedir()` as the cwd. After each `acquire()`, it eagerly refills the pool — but without knowing the project path, so the replacement shell lands in `~`. The next acquire pulls that home-dir shell, writes the `cd` fixup, and hands it off. This creates the consistent parity pattern: project, home, project, home.
- The fix calls `ptyPool.setDefaultCwd(options.cwd)` before `ptyPool.acquire()` in `acquirePtyProcess`. This means `refillPool()` (triggered synchronously inside `acquire()`) spawns the replacement shell at the correct project path rather than the stale home-dir default.

Resolves #5091

## Changes

- `electron/services/pty/terminalSpawn.ts` — call `setDefaultCwd` before `acquire` in the pooled branch of `acquirePtyProcess`
- `electron/services/__tests__/PtyPool.test.ts` — regression test verifying refill after `setDefaultCwd` uses the updated cwd
- `electron/services/pty/__tests__/terminalSpawn.test.ts` — 5 new tests covering the call-site ordering, pool-bypass branches, null pool handling, and the preserved `cd` relocation write

## Testing

All 17 tests across both files pass. Regression guard confirmed empirically: reverting the fix causes 2 tests to fail as expected. Typecheck, lint, and format all clean.

### DIFF
--- a/electron/services/__tests__/PtyPool.test.ts
+++ b/electron/services/__tests__/PtyPool.test.ts
@@ -81,6 +81,39 @@ describe("PtyPool", () => {
     pool.dispose();
   });
 
+  it("refills using updated defaultCwd after setDefaultCwd + acquire", async () => {
+    const warmed = createFakeProcess(601);
+    const refilled = createFakeProcess(602);
+    spawnMock.mockReturnValueOnce(warmed).mockReturnValueOnce(refilled);
+
+    const pool = new PtyPool({ poolSize: 1, defaultCwd: "/initial" });
+    await pool.warmPool();
+
+    expect(spawnMock).toHaveBeenCalledTimes(1);
+    expect(spawnMock).toHaveBeenNthCalledWith(
+      1,
+      expect.any(String),
+      expect.any(Array),
+      expect.objectContaining({ cwd: "/initial" })
+    );
+
+    pool.setDefaultCwd("/project/path");
+    const acquired = pool.acquire();
+    expect(acquired).toBe(warmed);
+
+    // acquire() triggers refillPool() synchronously; the refill spawn must
+    // use the updated cwd, not the stale "/initial" default. This is the
+    // regression guard for issue #5091 (alternating terminal cwd bug).
+    expect(spawnMock).toHaveBeenCalledTimes(2);
+    expect(spawnMock).toHaveBeenNthCalledWith(
+      2,
+      expect.any(String),
+      expect.any(Array),
+      expect.objectContaining({ cwd: "/project/path" })
+    );
+    pool.dispose();
+  });
+
   it("drops dead pooled terminals on acquire and refills the pool", async () => {
     spawnMock
       .mockReturnValueOnce(createFakeProcess("missing"))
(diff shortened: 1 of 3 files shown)


gregpriday:
Closing as superseded by #5102.

Both PRs target the same root bug — pooled PTYs landing in the wrong cwd — but take different approaches. #5102 goes deeper:

- Drains and refills the pool on project switch via an IPC message, so pooled entries are always at the correct cwd before acquire
- Adds a cwd-mismatch guard in `acquirePtyProcess` that falls back to a fresh spawn, giving multi-window / different-project setups a safe path
- Removes the `cd` shell write entirely, which is the underlying fragility issue #5097 documents (user `cd` overrides from zoxide, direnv, oh-my-zsh chpwd hooks could hijack that write)
- Uses an epoch counter to handle races between in-flight spawns and new drain cycles

This PR's minimal approach — call \`setDefaultCwd\` at the spawn site and keep the \`cd\` write — is exactly what #5102 intentionally moves away from. The new tests here (\`writes cd relocation to the acquired pooled shell\`) directly contradict #5102's regression guard that asserts no \`cd\` is written to pooled shells, and the \`setDefaultCwd\` test exercises a method #5102 removes in favor of \`drainAndRefill\`.

Issue #5091 (the user-visible symptom this PR targets) is the same bug as #5097 — fixing #5097 resolves it. I'll close #5091 pointing at #5102 once that lands.

Thanks for the fix — the analysis here helped confirm #5102 is on the right track.

PR #5101, opened 2026-04-09
hotfix(agents): add Kiro AWS SSO auth token to auth check

ychamare:
Kiro's auth detection missed users authenticated via AWS SSO because the cached token path was not checked. Add the SSO token file to configPathsAll so toolbar buttons render correctly. this was prevent Canopy from picking up Kiro-cli autoamtically

### DIFF
--- a/electron/services/__tests__/CliAvailabilityService.test.ts
+++ b/electron/services/__tests__/CliAvailabilityService.test.ts
@@ -443,6 +443,13 @@ describe("CliAvailabilityService", () => {
 
       const result = await service.checkAvailability();
       expect(result.kiro).toBe("ready");
+
+      // Upper-bound guard: the only Kiro auth file probed must be the SSO
+      // token cache. Catches any future reintroduction of extra Kiro paths.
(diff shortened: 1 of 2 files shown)


gregpriday (approved):
Cheers for spotting the SSO token path and getting the initial fix in. Good contribution.

I've refined the scope a bit on top of your work. The `.kiro/credentials` and `.kiro/config.json` paths were removed because Kiro CLI doesn't actually write to either of those for non-SSO auth (it goes straight to the OS keychain, same as Cursor). Keeping them would've caused false positives. The final `configPathsAll` is just `.aws/sso/cache/kiro-auth-token.json`, with `fallback: "installed"` for non-SSO users so they still get toolbar buttons when the binary is present.

Also updated the `CliAvailabilityService` tests to cover the SSO-ready path properly, added a regression guard against the bogus `.kiro/*` paths, and updated the diagnostic log assertion to expect the SSO path in the "checked: ..." output.

That sorts it out.

PR #5118, opened 2026-04-10
fix(theme): restore fontFamily in CodeMirror editor theme

### LINKED ISSUE #4053: File Preview Font Family
Image

Might just be my machine; I'm not sure the correct font family is landing in file preview.

Misplon:
## Summary

Fixes a regression of #4053 — the file preview (`CodeViewer`) was rendering in the browser's default monospace font instead of JetBrains Mono. The original fix (`9f5ec7b`) was undone in `43718d5` when `fontFamily` was removed from the shared `canopyTheme` to support sans-serif in the notes editor.

- Re-add `fontFamily: DEFAULT_TERMINAL_FONT_FAMILY` to `canopyTheme` settings
- Remove the ineffective `font-mono` CSS class from the `CodeViewer` wrapper (CodeMirror ignores inherited font-family)
- Set `text-[13px]` on the `CodeViewer` wrapper for consistent sizing with the notes editor

The notes editor is unaffected — it already overrides the font at the scroller level via `codeBlockExtension.ts`.

Closes #4053

## Test plan

- [ ] Open a file in the file preview — confirm it renders in JetBrains Mono at 13px
- [ ] Open the notes editor — confirm body text is sans-serif at 13px
- [ ] Add a fenced code block in notes — confirm it renders in JetBrains Mono
- [ ] Open the notes palette — confirm same behavior as notes editor

### DIFF
--- a/src/components/FileViewer/CodeViewer.tsx
+++ b/src/components/FileViewer/CodeViewer.tsx
@@ -197,7 +197,7 @@ export const CodeViewer = forwardRef<CodeViewerHandle, CodeViewerProps>(function
   return (
     <div
       className={cn(
-        "overflow-auto font-mono [&_.cm-editor]:min-h-full [&_.cm-scroller]:!overflow-visible",
+        "overflow-auto [&_.cm-editor]:min-h-full [&_.cm-scroller]:!overflow-visible",
         className
       )}
     >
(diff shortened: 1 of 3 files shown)


gregpriday (approved):
Nice work tracking down the root cause here. The regression was a bit subtle — the original `fontFamily` removal was scoped to fix the notes editor sans-serif issue, but it quietly broke CodeMirror's font in the file preview since CodeMirror ignores inherited CSS font-family and needs it set on the theme directly.

On top of your three commits I added a regression test that pins `fontFamily` on `canopyTheme`, so this specific path is covered going forward. Also fixed a minor typecheck issue in the test with an `Extension` type cast.

Happy to call this done.

PR #5152, opened 2026-04-16
fix(tests): resolve Windows CI failures from path separator and shell quoting

### LINKED ISSUE #4671: [Nightly] Tests failed on 2026-03-31
The nightly test suite failed.

**Run:** https://github.com/canopyide/canopy/actions/runs/23780421607
**Date:** 2026-03-31T04:31:49.218Z

dev-en-m:
## Summary

- **`WorkspaceService.resource.test.ts`**: Wrap the four variable-substitution tests in a nested `describe` block that pins `process.platform` to `'linux'`. These tests verify substitution logic, not OS-specific quoting — on Windows, `shellEscapeValue` produces double quotes which broke the single-quote assertions at lines 724, 739, and 770.
- **`environment.test.ts`**: Extend the `existsSync` mock to match both `/Canopy` and `\Canopy` path endings. On Windows, `path.join` uses backslash separators so the forward-slash-only check returned `false`, the migration gate was never entered, and `cpSync` was never called (line 639).

Closes #4671 (Windows wave — April 12–16 nightly runs).

## Test plan

- [x] `npx vitest run electron/workspace-host/__tests__/WorkspaceService.resource.test.ts` — 51 tests pass locally
- [x] `npx vitest run electron/setup/__tests__/environment.test.ts` — 34 tests pass locally
- [ ] Confirm April 17 nightly goes green on Windows for these two files

### DIFF
--- a/electron/setup/__tests__/environment.test.ts
+++ b/electron/setup/__tests__/environment.test.ts
@@ -630,7 +630,7 @@ describe("Canopy -> Daintree userData migration gating", () => {
     // Legacy Canopy userData exists; new Daintree userData does not.
     fsMock.existsSync.mockImplementation((p: string) => {
       if (p.endsWith(".rebrand-migrated")) return false;
-      if (p.endsWith("/Canopy")) return true;
+      if (p.endsWith("/Canopy") || p.endsWith("\\Canopy")) return true;
       return false;
     });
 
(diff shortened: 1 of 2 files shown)


gregpriday (approved):
Nice work tracking down these Windows failures. The platform-pinning approach for the variable-substitution tests is the right call — those tests are about substitution logic, not OS quoting, so isolating them from `shellEscapeValue`'s platform behaviour is cleaner than trying to make the assertions cross-platform.

Two small additions on top: switched the hardcoded `"/tmp/user-data/Canopy"` assertion inside the `cpSync` block to `path.join` (the `existsSync` widening let Windows enter that block, but the forward-slash string still caused the assertion to fail), and made the `process.platform` descriptor configurable in the `afterEach` restore to match the defensive teardown pattern used in `AutoUpdaterService.test.ts`.

Happy to call this done.

Issue #5137, opened 2026-04-13
Agent panel reverts to original worktree after project switch


gregpriday:
Drag an agent panel from its original worktree into another worktree in the same project — it moves correctly and works there. Switch to a different project, switch back, and the panel has snapped back to the worktree where it was originally created. The dragged-into worktree no longer shows it.

## Reproduction

1. Create an agent panel in worktree A (root worktree).
2. Drag the panel onto worktree B.
3. Confirm it's working in B.
4. Switch to a different project.
5. Switch back to the original project.

**Expected:** panel remains in worktree B.
**Actual:** panel reappears in worktree A.

## Where the mismatch lives

Renderer-side drag updates the panel's `worktreeId` in the store ([`moveTerminalToWorktree` in `src/store/slices/panelRegistry/restart.ts`](https://github.com/canopyide/canopy/blob/a7377f69cd157dc753d104dbede2c0ede5040d04/src/store/slices/panelRegistry/restart.ts)). Persistence captures the updated id correctly ([`panelToSnapshot` in `src/store/persistence/panelPersistence.ts`](https://github.com/canopyide/canopy/blob/a7377f69cd157dc753d104dbede2c0ede5040d04/src/store/persistence/panelPersistence.ts#L22)).

The backend PTY, however, is never moved — drag is a renderer-only store mutation. The main-process terminal record still carries its original `worktreeId`.

On project re-activation, hydration prefers the backend terminal's `worktreeId` over the saved snapshot:

- [`buildArgsForBackendTerminal` in `src/utils/stateHydration/statePatcher.ts`](https://github.com/canopyide/canopy/blob/a7377f69cd157dc753d104dbede2c0ede5040d04/src/utils/stateHydration/statePatcher.ts#L162) sets `worktreeId: backendTerminal.worktreeId` with no fallback.
- The reconnected-fallback path a few lines down already does the right thing: `reconnectedTerminal.worktreeId ?? saved.worktreeId`.

So the backend's stale id overwrites the snapshot's correct id, and the panel reappears in its creation worktree.

## Secondary observation

After switching back, the agent looked paused — a single "continue" prompt resumed it. Possibly unrelated (agent-side idle), but worth noting in case stdin detach/reattach during `detachTerminalsForProjectSwitch` → rehydration is stalling input briefly. Primary bug is the worktree reversion; this is just a data point.

## Fix direction

[...]

PR #5176, opened 2026-04-17
fix(pty): treat panel worktreeId as renderer-owned layout state

### LINKED ISSUE #5139: Treat panel worktreeId as renderer-owned layout state
Follow-up to #5137. That issue is a targeted fix for one symptom; this is the underlying structural problem.

## Problem

`worktreeId` is currently duplicated across two owners:

- The renderer panel store treats it as **layout state** — which worktree tab the panel appears under. Drag-and-drop mutates it here.
- The backend PTY/terminal record carries its own `worktreeId`, treated as process metadata.

Nothing keeps them in sync. Drag updates only the renderer. Hydration on project re-activation ([`buildArgsForBackendTerminal` in `src/utils/stateHydration/statePatcher.ts`](https://github.com/canopyide/canopy/blob/a7377f69cd157dc753d104dbede2c0ede5040d04/src/utils/stateHydration/statePatcher.ts#L162)) prefers the backend's copy, so the renderer's layout change gets clobbered. #5137 is the most visible symptom; similar drift is likely latent elsewhere (serialization, cross-window moves, reconnect paths).

The PTY doesn't actually need a `worktreeId` — it needs `cwd` and `env`. Worktree association is purely a UI placement concern.

## Desired end state

- `worktreeId` lives in renderer panel state only. Snapshots remain the source of truth across project switches.
- Backend PTY/terminal records drop `worktreeId` from their schema. Anything the backend genuinely needs from the worktree (cwd, env, git context at spawn) is resolved at creation time and stored as its own field.
- Hydration reconciles panels against backend PTYs by pty/terminal id, not worktree. The backend has no opinion about which worktree a panel belongs to.
- Drag-to-worktree becomes a pure renderer-side layout mutation — no IPC round-trip needed to stay consistent.

## Relationship to #5137

No hard dependency. #5137 can land first as a quick fix (fallback to saved snapshot in `buildArgsForBackendTerminal`). This refactor supersedes it — once `worktreeId` is renderer-only, the hydration branch in #5137 becomes moot and can be simplified or removed as part of this work.

If this lands first, close #5137 as resolved by it.

## Scope notes

Touches: PTY host types, terminal record schema in main, IPC payloads that currently carry `worktreeId`, hydration (`statePatcher.ts`), persistence (`panelPersistence.ts`), and any callers that read `worktreeId` off a backend terminal. Worth a careful grep — this field has spread.

gregpriday:
## Summary

- `worktreeId` was being stored and forwarded by the backend PTY/terminal layer, meaning hydration on reconnect would clobber any drag-and-drop placement the user had made during the session
- Strips `worktreeId` from all backend types and emission points (spawn options, snapshots, terminal info, agent event payloads, host events), making it strictly renderer-owned layout state
- Adds `inferWorktreeIdFromCwd` in the state patcher for segment-aware longest-prefix matching, so the orphan path can still resolve which worktree a reconnected terminal belongs to

Resolves #5139. Supersedes #5137.

## Changes

- Removed `worktreeId` from `PtyHostSpawnOptions`, `PtyHostTerminalSnapshot`, `PtyHostTerminalInfo`, `TerminalPublicState`, `TerminalSnapshot`, `BackendTerminalInfo`, `TerminalReconnectResult`, `TerminalInfoPayload`, `TerminalSpawnOptions`, `TerminalSpawnOptionsSchema`, and all agent event payloads
- `TerminalProcess`, `AgentStateService`, `PtyManager.getTerminalInfo`, pty-host snapshot/forward paths, and `terminalSpawn` env injection all stop emitting `worktreeId`
- `cwd` threaded end-to-end through `agent:state-changed` so `PreAgentSnapshotService` still has a git target (keys on `cwd`, which was always the same absolute path as `worktreeId`)
- `statePatcher.ts`: removed the `?? backendTerminal.worktreeId` fallbacks that were the direct cause of hydration clobbering; added `inferWorktreeIdFromCwd` with POSIX + Windows separator support
- `buildArgsForReconnectedFallback` now gets the `activeWorktreeId` fallback so legacy saved panels without `worktreeId` still land in the active worktree
- Renderer spawn callers (`TerminalRegistryController`, `panelRegistry/core.ts`, `restart.ts`) drop `worktreeId` from options

## Testing

Full suite: 10419 passed / 1 skipped / 2 todo / 0 failed.

Updated and added tests in `statePatcher.test.ts` (7 new `inferWorktreeIdFromCwd` cases, removed the two tests that locked in the old bad behaviour), `PreAgentSnapshotService.test.ts`, `lifecycle.spawn.test.ts`, `PtyManager.integration.test.ts`, `PtyEventsBridge.test.ts`, and `devPreview.session.test.ts`.


gregpriday:

This came up in an automated merge pass and needs a closer look before we land it.

The change strips `worktreeId` from `agent:state-changed` on the backend producer side (`AgentStateService` and `pty-host`), and the corresponding field is no longer present in the `events.ts` `DaintreeEventMap` for that event. But `AgentNotificationService.handleStateChanged` still reads `payload.worktreeId` and feeds it into `isFocusedOnWorktree`. With `worktreeId` permanently undefined for backend-emitted state changes, `isFocusedOnWorktree(undefined)` returns `true` whenever the window is focused (line 557: `if (!worktreeId) return true;`), which suppresses both completion and waiting-escalation notifications for agents in worktrees the user isn't currently viewing.

In practice: I'm working in worktree A, an agent finishes in worktree B, my window is focused, and I get no notification. That's a meaningful UX regression. The `getLabel` and dedup-key derivations also degrade (`key = agentId ?? worktreeId ?? "agent"`), and `EventBuffer.getFiltered({ worktreeId })` returns empty for any of the agent:* event types now.

A few options for resolving this:

1. Have `AgentNotificationService` derive worktreeId in main from `cwd` plus the persisted `appState.terminals[].worktreeId` lookup by terminalId. Cheapest fix, but the renderer-owned worktreeId may briefly lag if the user just dragged the panel.
2. Keep emitting `worktreeId` on `agent:state-changed` specifically, since the consumer is a backend service that isn't part of the renderer-owned-layout intent. The PR's framing is about hydration clobbering panel placement, which this event doesn't drive.
3. Add a small main-side `inferWorktreeIdFromCwd` (mirroring the renderer helper this PR adds) and feed it the worktrees list via an existing service. More work but most consistent with the design.

Excluding this PR from the current merge batch while we sort out which way to go. Same concern applies to `agent:spawned`/`output`/`completed`/`killed` if anything else in main filters on `worktreeId`.

Once the notification path is sorted I think the rest of the PR is in good shape, the segment-aware `inferWorktreeIdFromCwd` is the right tool for the orphan path.

gregpriday:
Good catch, that's a real regression.

Went with option 1: resolve worktreeId in main from `appState.terminals` by terminalId. Cheapest fix, and AgentNotificationService already reads from `store.appState.terminals` elsewhere (location, watched set, active worktree), so it stays consistent with existing patterns in that file. The renderer is the authority for this value, `appState` is the sync point, so we're just pulling from it one more place.

Trade-off is the lag you mentioned: if someone drag-and-drops a panel, the resolved worktreeId may be a tick behind. The fallback is `undefined` though, which just means routing and dedup degrade for that tick, not that notifications fire incorrectly. Happy with that.

Added four tests: completion and waiting both resolve worktreeId from appState, explicit `payload.worktreeId` still wins, and missing-terminal degrades to `undefined`. Pushed as 2bc840140.

Agree on the broader EventBuffer filter concern for a separate issue if we want to revisit. For this PR the notification path is the load-bearing one.

Issue #5158, opened 2026-04-16
Default-pin model pins agents the user has never installed


gregpriday:
On 0.7.0 the agent pin model flipped to opt-out: any registered built-in agent with no explicit `pinned` value is treated as pinned. That was intended to cover "agents you've already adopted" — but the defaulting fires for every agent in the registry, including ones the user has never installed and never interacted with. OpenCode, Cursor, Kira, Copilot, etc. all show up as pinned in Settings → Agents and in the new tray's pin pill, despite never being on the machine. For users upgrading from 0.6.x, the effect compounds with store migration 012, which grandfathers *every* prior `agentSettings.agents[id]` entry where `selected !== false` into `pinned: true` — regardless of whether that entry represented deliberate user intent or just a default the normalizer wrote.

## Desired behavior

- A new, never-touched built-in agent should default to `pinned: false` (or have no synthesized entry at all, so `isAgentPinned(undefined)` resolves to `false`) when its CLI is not installed on the machine.
- An installed or ready agent with no explicit user choice may default to `pinned: true` — the user has demonstrated intent by installing the CLI.
- A one-shot cleanup should run on 0.7.1 upgrade for users who already hit the bug: for each stored entry whose `pinned` was synthesized by the default-pin flip (no other user-set fields on the entry) and whose agent is not currently installed, clear `pinned` so the tray + Settings reflect reality.
- Explicit user choices (`pinned: true` set via the tray or Settings, `pinned: false` set by unpinning) must be preserved untouched — the cleanup is specifically about purging phantom defaults, not rewriting real preferences.

## Evidence

In-memory normalizer defaults every registered agent to pinned regardless of install state:

```ts
export function normalizeAgentSelection(settings: AgentSettings): AgentSettings {
  const registeredIds = getEffectiveAgentIds();
  let changed = false;
  const agents = { ...settings.agents };

  for (const id of registeredIds) {
    const entry = agents[id];

    if (!entry) {
      agents[id] = { pinned: true };
      changed = true;
      continue;
    }

    if (entry.pinned === undefined) {
      agents[id] = { ...entry, pinned: true };
      changed = true;
    }
  }

  return changed ? { ...settings, agents } : settings;
}
```

https://github.com/daintreehq/daintree/blob/f7978e5c6d1ddab8f164c68aceff42887ea14a73/src/store/agentSettingsStore.ts#L14-L35

Migration 012 persists the grandfathering to disk, including for entries the user never explicitly selected:

```ts
[...]

PR #5295, opened 2026-04-17
fix(toolbar): sync agent pin state between tray and Settings toolbar (#5112)

### LINKED ISSUE #5112: Sync agent tray pin toggle with Settings Toolbar and refine tray UX
Two related cleanups in the same file surface: fix the split-brain between the Agent Tray and Settings > Toolbar, and refine the tray dropdown structure.

Depends on #5109.

## Split-brain visibility gate

Unpinning an agent from the Agent Tray dropdown writes to `agentSettingsStore`. Toggling the same agent's visibility in Settings > Toolbar writes to `toolbarPreferencesStore.hiddenButtons[]`. Neither store knows about the other, so users can unpin in one surface and see the agent still marked "visible" in the other. Both surfaces must write to the same field and agree on the result.

Under the Option A architecture (agent settings store owns pin state), the Settings > Toolbar visibility checkbox handler is conditional: for agent IDs it calls `setAgentPinned` directly; for non-agent button IDs it continues to use the existing `toggleButtonVisibility` path. This is leaky — the conditional leaks store-awareness into a UI component — but it's the right trade-off because the alternative (moving all pin state to toolbar prefs) downgrades persistence from main-process to renderer-only localStorage.

Validation: a test that unpins in the tray and asserts the settings tab reflects it, and vice versa. The bug has to actually be fixed, not just refactored.

## Agent tray dropdown refinements

- **Don't duplicate pinned agents in the Launch list.** A pinned agent is already visible in the main toolbar — listing it again in the tray's Launch section wastes dropdown space and violates information density.
- **Fixed two-category structure:** Launch (ready, unpinned) on top, Needs Setup (present-but-not-ready or missing) below. No dynamic grouping, no collapsible sections.
- **Sort Launch by recency of last launch** if the tracking is cheap to add (timestamp on `agent.launch` action dispatch, transient in-memory is fine). Alphabetical if not.
- **No search field.** 20 items fit without one. Add search only if the supported-agent list grows past ~30.
- **No hide-from-tray affordance.** Recency sort naturally sinks unused agents; an explicit hide toggle is cognitive load without payoff.

## Out of scope

- Renaming internal APIs (Phase 1's job, #5109)
- "Enable" → "Pin" UI copy pass (separate filter/copy audit issue)
- First-run welcome card (#5111)

gregpriday:
## Summary

- The Agent Tray and Settings > Toolbar were using separate stores to track agent pin state, causing split-brain where unpinning in one surface had no effect on the other. Both surfaces now write to `agentSettingsStore` as the single source of truth.
- `ToolbarSettingsTab` now routes agent button visibility toggles through `setAgentPinned` instead of `toggleButtonVisibility`, keeping IPC-backed persistence for agent pin state while leaving the non-agent button path unchanged.
- Pinned agents are filtered out of the Agent Tray's Launch list since they're already visible in the toolbar proper.

Resolves #5112

## Changes

- `toolbarPreferencesStore.ts`: added `isAgentId` guard so agent IDs fall back to `agentSettingsStore.setAgentPinned` for the visibility toggle
- `ToolbarSettingsTab.tsx`: conditional dispatch based on whether the button ID maps to an agent
- `AgentTrayButton.tsx`: filters out already-pinned agents from the Launch section of the dropdown
- New integration test (`agentPinSync.integration.test.tsx`) covering the round-trip sync in both directions
- Extended unit tests for `ToolbarSettingsTab` and `toolbarPreferencesStore`

## Testing

Unit and integration tests added covering the cross-surface sync. All existing tests pass. The integration test specifically exercises the scenario described in the issue: unpin via tray, assert Settings reflects it, and vice versa.

### DIFF
--- a/src/components/Layout/AgentTrayButton.tsx
+++ b/src/components/Layout/AgentTrayButton.tsx
@@ -22,6 +22,7 @@ import {
 import { getBrandColorHex } from "@/lib/colorUtils";
 import { getAgentConfig, type AgentIconProps } from "@/config/agents";
 import { actionService } from "@/services/ActionService";
+import { useActionMruStore } from "@/store/actionMruStore";
 import { useAgentSettingsStore } from "@/store/agentSettingsStore";
 import { useCliAvailabilityStore } from "@/store/cliAvailabilityStore";
 import { usePanelStore } from "@/store/panelStore";
@@ -87,6 +88,8 @@ export function AgentTrayButton({
 }: AgentTrayButtonProps) {
   const agentSettings = useAgentSettingsStore((s) => s.settings);
   const setAgentPinned = useAgentSettingsStore((s) => s.setAgentPinned);
+
+  const actionMruList = useActionMruStore((s) => s.actionMruList);
 
   const refreshAvailability = useCliAvailabilityStore((s) => s.refresh);
   const hasRealData = useCliAvailabilityStore((s) => s.hasRealData);
@@ -190,10 +193,13 @@ export function AgentTrayButton({
 
       const state = agentAvailability?.[id];
       if (isAgentReady(state)) {
-        launchable.push(row);
+        // Pinned agents already live in the main toolbar — listing them in
+        // the tray's Launch section wastes dropdown space. Users unpin via
+        // the main toolbar button or Settings > Toolbar.
+        if (!pinned) launchable.push(row);
       } else if (isAgentInstalled(state)) {
         // "installed" means the CLI is on PATH but not fully authenticated
-        // or configured yet. These belong in "Also Available" with a setup
+        // or configured yet. These belong in "Needs Setup" with a setup
         // badge. Missing agents do NOT get promoted here.
         needsSetup.push(row);
       }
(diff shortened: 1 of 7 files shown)


gregpriday:
This has caused a regression in the agent tray. Opening the tray no longer shows a list of agents for anyone who already has agent CLIs installed.

The `if (!pinned) launchable.push(row)` change at `AgentTrayButton.tsx:376` collides with the auto-pin logic in `normalizeAgentSelection` (`agentSettingsStore.ts:49-52`, from #5158). That normaliser synthesises `pinned: true` for any entry with an installed CLI on upgrade paths, so for existing users every ready agent ends up pinned. With the new exclusion, Launch renders empty and what remains is the "Available Agents" fallback, which labels every row "Setup" and routes to Settings instead of launching.

Plan is to drop the `!pinned` exclusion and let pinned agents show in Launch again. The pin indicator on each row already signals toolbar state, so the duplication argument doesn't really hold up once you factor in the normaliser. The unit test at `AgentTrayButton.test.tsx:323` ("excludes all pinned agents from the Launch section") will need to flip as well.

PR #5330, opened 2026-04-17
perf(store): migrate orchestrator to subscribeWithSelector (#5247)

### LINKED ISSUE #5247: Use subscribeWithSelector in renderer store orchestrator
The renderer-side store orchestrator registers five cross-store rules, four of which listen to the panel store. All five use vanilla `store.subscribe(fn)` which fires on every `setState`, regardless of whether the field a rule cares about actually changed. Each rule then re-implements its own prev/next reference guard inline.

So for one panel-store mutation, four callbacks wake up, each does its own equality check against a different field (`focusedId`, `focusedId`, `panelIds`, `panelIds`), and three return early. The work is correct but wasteful, and the guards obscure intent — the reader has to scan each callback to learn which field the rule actually cares about.

Zustand ships `subscribeWithSelector` precisely for this. The rule declares its selector explicitly, Zustand handles the diff, and the callback body becomes the effect only.

```ts
usePanelStore.subscribe(
  (s) => s.focusedId,
  (focusedId, prev) => { /* effect */ }
);
```

Rule 1 additionally conflates three concerns behind a single `focusedId` change: tracking focus on the worktree selection store, switching the active worktree when a focused terminal belongs to a different one, and recording terminal MRU. Splitting into three named rules would make each intent self-documenting and testable in isolation.

### Evidence

[`src/store/rendererStoreOrchestrator.ts`](https://github.com/daintreehq/daintree/blob/185ccb2691cfbdfbf2c49fcf6ffe8b7a902cf16a/src/store/rendererStoreOrchestrator.ts) — all five subscriptions are vanilla `subscribe`, each with a manual `state.x === prevState.x` guard. Rule 1 runs three unrelated side effects off one focus-change guard.

### Desired end state

Orchestrator rules express their dependency via `subscribeWithSelector` selectors, not inline guards. Rule 1 splits into separate rules per concern. Equivalent runtime behaviour, fewer spurious wakeups during panel mutation storms, callback bodies contain only the effect.

The existing orchestrator test coverage should continue to pass unchanged. No persisted-state or IPC changes.

gregpriday:
## Summary

- Wrapped `usePanelStore` and `useCliAvailabilityStore` with `subscribeWithSelector` middleware so subscription callbacks only fire when the selected slice actually changes, cutting redundant re-evaluations in the orchestrator.
- Rewrote `rendererStoreOrchestrator.ts` to use selector-based subscriptions throughout. Rule 1 is split into three named rules (focus tracking, active worktree switch, MRU recording) and the module-level `prevTerminalIds`, `prevTerminalsById`, and `prevIdSet` closures are gone entirely.
- Compound selectors in Rules 3 and 5 use `shallow` equality from `zustand/shallow` to avoid unnecessary callbacks on unrelated state updates.

Resolves #5247

## Changes

- `src/store/panelStore.ts` — added `subscribeWithSelector` middleware wrapper
- `src/store/cliAvailabilityStore.ts` — same middleware addition
- `src/store/rendererStoreOrchestrator.ts` — fully rewritten subscription logic; 5 rules now use selector-based subscriptions with no module-level mutable closures

## Testing

908-line orchestrator test suite passes unchanged (29/29 tests). All 941 store tests green. Typecheck, lint (401/401 ratchet), and format checks clean.

### DIFF
--- a/src/store/cliAvailabilityStore.ts
+++ b/src/store/cliAvailabilityStore.ts
@@ -1,4 +1,5 @@
 import { create } from "zustand";
+import { subscribeWithSelector } from "zustand/middleware";
 import type { CliAvailability, AgentAvailabilityState } from "@shared/types";
 import { cliAvailabilityClient } from "@/clients";
 import { getAgentIds } from "@/config/agents";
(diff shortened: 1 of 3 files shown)


gregpriday:
Attempted rebase onto develop failed — conflicts with recent panelStore changes (trashPanel/trashPanelGroup reshuffled the maximized-state logic region this PR rewrites). Needs manual resolution. Likely easier once #5327 lands, since both PRs touch `rendererStoreOrchestrator.ts`.

gregpriday:
Re-attempted rebase onto develop — still conflicting. The intersection with #5327 (DisposableStore) which landed earlier is structural: both PRs refactor \`rendererStoreOrchestrator.ts\` via different strategies.

The rebase needs a manual pass that combines both:
- Keep 5330's selector-based subscribes (the perf win)
- Wrap each unsubscribe in \`toDisposable\` and add to the \`DisposableStore\` from 5327

Conflicts to resolve:
- \`src/store/rendererStoreOrchestrator.ts\` — 3 regions
- \`src/store/panelStore.ts\` — 1 region (lines 180-320) where trashPanel/trashPanelGroup logic evolved

Once rebased, verify \`subscribeWithSelector\` middleware is still applied to \`usePanelStore\` and \`useCliAvailabilityStore\`.

Issue #5179, opened 2026-04-17
Clone layout creates terminals with title but no command


gregpriday:
When creating a new worktree with "clone existing layout", the cloned terminals spawn with the correct title and kind but the agent command is never sent. The PTY starts a bare shell — the agent (Claude, Codex, etc.) never launches.

Reproduces with a single clone, not just bulk.

The clone path in `BulkCreateWorktreeDialog` calls `usePanelStore.addPanel()` passing only `kind`, `agentId`, `title`, `cwd`, `worktreeId`, and `exitBehavior`. No `command`, no `initialPrompt`, no agent launch flags. `terminalClient.spawn()` then runs with `command: undefined`, so the shell starts but the agent invocation is skipped.

```ts
for (const t of cloneTerminals) {
  await usePanelStore.getState().addPanel({
    kind: t.type === "dev-preview" ? "dev-preview"
        : t.type === "terminal" ? "terminal"
        : "agent",
    agentId: t.type !== "terminal" && t.type !== "dev-preview" ? t.type : undefined,
    title: t.title,
    cwd: worktreePath,
    worktreeId,
    exitBehavior: t.exitBehavior,
  });
}
```

See https://github.com/daintreehq/daintree/blob/82a3274686a3d8d2eafabb89772cccee902280e1/src/components/GitHub/BulkCreateWorktreeDialog.tsx#L777-L792

Compare to the recipe path in the same file, which routes through `useRecipeStore.runRecipeWithResults()` and rebuilds the agent command via `generateAgentCommand()` using current settings and flags.

## Desired behavior

Cloning a layout should clone the **structure** (which panels exist, their kinds, titles, agent assignments, dev-preview presence, exit behavior) and then start each terminal **fresh** — rebuilding the spawn command from the current agent config the same way a new terminal does, not replaying a captured command string.

Specifically:
- For `agent` panels: rebuild the spawn command from the current agent config (model, flags, settings) — don't carry over the old command verbatim
- For `terminal` panels: start a plain shell
- For `dev-preview` panels: no PTY, just restore the panel

The cloned layout should give the user the same workspace shape as the source worktree, but with every agent freshly started in the new worktree's cwd.

PR #5339, opened 2026-04-17
refactor(worktree): extract shared cloneLayout helper in NewWorktreeDialog

### LINKED ISSUE #5202: Dedupe clone-layout loops in NewWorktreeDialog
`NewWorktreeDialog.tsx` has two near-identical clone-layout blocks: one in the "create from scratch" flow and one in the "create from issue" flow. They both call `generateRecipeFromActiveTerminals`, iterate, spawn panels through `addPanel`, and call `notify` on failure. The bodies are copy-pasted.

Any change to one has to be made to the other, and the bulk clone flow in `BulkCreateWorktreeDialog.tsx` has a third variant of roughly the same shape. This is exactly the surface that #5179 and #5197 are fixing: three forks of the same logic, each missing a slightly different piece.

Extracting the loop into one helper (either a local function in the dialog or a utility alongside `panelDuplicationService`) would make the next fix land once instead of three times.

See the two offending blocks:

- [`NewWorktreeDialog.tsx:519-547`](https://github.com/daintreehq/daintree/blob/05ce228f21e14ca94fcdbdb2bcb4e7124774a832/src/components/Worktree/NewWorktreeDialog.tsx#L519-L547)
- [`NewWorktreeDialog.tsx:701-728`](https://github.com/daintreehq/daintree/blob/05ce228f21e14ca94fcdbdb2bcb4e7124774a832/src/components/Worktree/NewWorktreeDialog.tsx#L701-L728)

And the adjacent third variant in [`BulkCreateWorktreeDialog.tsx:771-803`](https://github.com/daintreehq/daintree/blob/05ce228f21e14ca94fcdbdb2bcb4e7124774a832/src/components/GitHub/BulkCreateWorktreeDialog.tsx#L771-L803).

No behaviour change. Just collapse the duplication so the next clone-layout fix doesn't have to be applied three times.

Better to land this after #5179 and #5197 have settled, so the helper captures the right shape.

gregpriday:
## Summary

- Extracted a `cloneLayout` helper at module scope in `NewWorktreeDialog.tsx`, collapsing two byte-identical clone-layout blocks (lines ~519 and ~701) into a single shared function. Both the create-from-scratch and create-from-issue flows now delegate with a single call each.
- The helper owns the `generateRecipeFromActiveTerminals → addPanel` loop and its `try/catch + notify` fallback. Any future fix to this logic lands once instead of twice.
- `BulkCreateWorktreeDialog.tsx` is intentionally untouched. Its PQueue concurrency, per-panel error tracking, and progress events make it a meaningfully different contract.

Resolves #5202

## Changes

- `src/components/Worktree/NewWorktreeDialog.tsx`: extracted `cloneLayout` module-scope helper, replaced two duplicate blocks with single-line delegates
- `src/components/Worktree/__tests__/NewWorktreeDialog.test.tsx`: added 4 tests covering happy-path (existing-branch), happy-path (new-branch), failure-path (notify fires, worktree creation still succeeds), and no-source-worktree skip

## Testing

All 13 tests in `NewWorktreeDialog.test.tsx` pass. Typecheck, lint ratchet, and format all clean.

### DIFF
--- a/src/components/Worktree/NewWorktreeDialog.tsx
+++ b/src/components/Worktree/NewWorktreeDialog.tsx
@@ -82,6 +82,38 @@ function HighlightBranchText({
   return <>{nodes}</>;
 }
 
+async function cloneLayout(
+  sourceWorktreeId: string,
+  worktreePath: string,
+  worktreeId: string
+): Promise<void> {
+  try {
+    const terminals = useRecipeStore.getState().generateRecipeFromActiveTerminals(sourceWorktreeId);
+    for (const t of terminals) {
+      const isDevPreview = t.type === "dev-preview";
+      const isAgent = !isDevPreview && t.type !== "terminal";
+      await usePanelStore.getState().addPanel({
+        kind: isDevPreview ? "dev-preview" : isAgent ? "agent" : "terminal",
+        agentId: isAgent ? t.type : undefined,
+        title: t.title,
+        cwd: worktreePath,
+        worktreeId,
+        exitBehavior: t.exitBehavior,
+        devCommand: isDevPreview ? t.devCommand : undefined,
+        agentModelId: isAgent ? t.agentModelId : undefined,
+        agentLaunchFlags: isAgent ? t.agentLaunchFlags : undefined,
+      });
+    }
+  } catch (cloneErr) {
+    const message = cloneErr instanceof Error ? cloneErr.message : "Failed to clone layout";
+    notify({
+      type: "warning",
+      title: "Could not clone layout",
+      message: `${message} — worktree was created successfully`,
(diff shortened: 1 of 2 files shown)


gregpriday:
Closing — this refactor is now superseded by #5345, which landed a richer `cloneLayoutPanels` helper that handles agent command regeneration (see #5179/#4781 for the session-ID path-scoping issue that the simpler `cloneLayout` extraction here would miss).

The two call sites in `NewWorktreeDialog.tsx` already use the new helper, so there's no remaining deduplication to do. Closing rather than rebasing since the commit would be empty after conflict resolution.

PR #5473, opened 2026-04-18
feat(telemetry): scrub free-text secrets from telemetry and diagnostic bundles

### LINKED ISSUE #5418: Scrub free-text secrets from telemetry and diagnostic bundles
Current redaction is key-name based (fields called `token`, `password`, `apiKey`, and friends) plus a home-directory replacement on paths. That catches structured fields but misses secrets embedded in free-text log messages and stack traces.

Real examples that reach Sentry and diagnostic bundles today:

- `git clone failed: https://x-access-token:ghp_abc123@github.com/org/repo.git`, full PAT in the URL
- Error messages containing `Authorization: Bearer eyJ...` when HTTP requests are logged
- Shell output containing `AKIA...` AWS keys from a misconfigured `.envrc`
- `sk-ant-...` / `sk-...` keys from agent CLI configs echoed into a terminal and captured in diagnostic events

None of these fire the existing redactor, because the secret sits in a message string, not a `{ token: ... }` field. Stack traces and URLs leak the same way: query strings like `?access_token=...` on OAuth callback URLs, basic-auth URLs beyond the GitHub case the current regex handles.

The scrubbing pass wants to be pattern-based (sigil-anchored: `ghp_`, `github_pat_`, `sk-ant-`, `sk-`, `xox[abprs]-`, `AKIA`, `Bearer `, JWT triple-dot shape, PEM blocks) and applied at two boundaries:

1. Sentry `beforeSend`, before upload, after existing sanitisation
2. `DiagnosticsCollector` output, before the bundle is handed to the user

Not on the hot logger write path. Runtime regex on every log line invites ReDoS and CPU spikes. Local-only log files can stay as-is; the gate is anything that leaves the user's machine.

`mazen160/secrets-patterns-db` keeps a ReDoS-tested regex library that Gitleaks and TruffleHog both consume. Any pattern that ships should pass `safe-regex` (or equivalent) in CI, because a catastrophic backtracking pattern can lock the main thread during a log-heavy session.

## Evidence

- Key-based redaction today: [logger.ts](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/electron/utils/logger.ts#L116-L130)
- Path-based sanitiser: [TelemetryService.ts](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/electron/services/TelemetryService.ts#L19-L29)
- Diagnostic bundle `redactDeep`: [DiagnosticsCollector.ts](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/electron/services/DiagnosticsCollector.ts#L24-L53)

gregpriday:
## Summary

- Adds `electron/utils/secretScrubber.ts` with 15 sigil-anchored bounded regex patterns covering GitHub PATs (classic, fine-grained, app), Anthropic, OpenAI, Stripe, Slack, Google, AWS, npm, Azure connection strings, PEM blocks, JWTs, Bearer tokens, and OAuth params. All patterns are validated against ReDoS using `safe-regex2` in tests.
- Wires scrubbing into `TelemetryService.sanitizeEvent()` (exception values, stack frames, breadcrumbs, extra fields, request headers/cookies/data/query strings/URLs) and into `DiagnosticsCollector.redactDeep()` and `collectLogs()`. Explicitly not on the logger hot path per the issue spec.
- 134 tests total: 42 scrubber unit tests, 57 telemetry integration tests (including 12 adversarial cases for partial matches, boundary conditions, and fail-closed error handling), and 23 existing diagnostics tests.

Resolves #5418

## Changes

- `electron/utils/secretScrubber.ts` — new module, 15 patterns, `scrubSecrets()` and `scrubSecretsDeep()` exports
- `electron/services/TelemetryService.ts` — `sanitizeEvent()` wired through scrubber across all string fields
- `electron/services/DiagnosticsCollector.ts` — `redactDeep()` and `collectLogs()` wired through scrubber
- `electron/utils/__tests__/secretScrubber.test.ts` — unit + ReDoS safety tests
- `electron/services/__tests__/TelemetryService.test.ts` — telemetry sanitisation tests
- `electron/services/__tests__/DiagnosticsCollector.adversarial.test.ts` — adversarial edge case tests
- `package.json` / `package-lock.json` — adds `safe-regex2@^5.1.0` devDependency

## Testing

All 134 tests pass. Adversarial cases cover partial token matches, deeply nested structures, null/undefined elements, and throw paths (fail-closed behaviour). ReDoS safety is asserted per-pattern using `safe-regex2` so any future pattern additions are automatically validated.

### DIFF
--- a/electron/services/DiagnosticsCollector.ts
+++ b/electron/services/DiagnosticsCollector.ts
@@ -5,6 +5,7 @@ import { app, screen } from "electron";
 import { sanitizePath } from "./TelemetryService.js";
 import { logBuffer } from "./LogBuffer.js";
 import { getPtyManager } from "./PtyManager.js";
+import { scrubSecrets } from "../utils/secretScrubber.js";
 import { store } from "../store.js";
 import type { HandlerDependencies } from "../ipc/types.js";
 
@@ -24,11 +25,15 @@ function withTimeout<T>(promise: Promise<T>, ms: number, fallback: T): Promise<T
 const SENSITIVE_KEY_PATTERN =
   /token|password|secret|apiKey|api_key|credential|authorization|private_key|passphrase/i;
 
+function sanitizeString(value: string): string {
+  return scrubSecrets(sanitizePath(value));
+}
+
 function redactDeep(value: unknown): unknown {
   if (value === null || value === undefined) return value;
 
   if (typeof value === "string") {
-    let result = sanitizePath(value);
+    let result = sanitizeString(value);
     result = result.replace(/https?:\/\/[^@\s]+@/g, "https://<redacted>@");
     return result;
   }
@@ -365,7 +370,7 @@ async function collectLogs() {
       totalEntries: entries.length,
       recentEntries: recent.map((e) => ({
         ...e,
-        message: truncateDiagnosticString(sanitizePath(e.message)),
+        message: truncateDiagnosticString(sanitizeString(e.message)),
         context: e.context ? truncateDeep(redactDeep(e.context)) : undefined,
       })),
     };
--- a/electron/services/TelemetryService.ts
+++ b/electron/services/TelemetryService.ts
@@ -187,7 +187,11 @@ export async function initializeTelemetry(): Promise<void> {
         environment: app.isPackaged ? "production" : "development",
         // Do not set `sampleRate` — it defaults to 1.0 (100% error capture). If
         // performance tracing is ever added, use `tracesSampleRate` instead.
-        beforeSend: (event) => sanitizeEvent(event as SentryEvent) as typeof event,
+        // The local `SentryEvent` interface is a narrower projection of the
+        // SDK's `Event` type — cast through `unknown` at the hook boundary so
+        // our scrubbing logic can use the shape we control.
+        beforeSend: (event) =>
+          sanitizeEvent(event as unknown as SentryEvent) as unknown as typeof event,
         initialScope: {
           tags: {
             platform: process.platform,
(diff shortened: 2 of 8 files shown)


gregpriday:

Reviewed as part of an automated merge pipeline. Scrubbing coverage has a few meaningful gaps for a PR whose contract is "scrub free-text secrets":

- **electron/utils/secretScrubber.ts** — AWS secret access key (40-char `[A-Za-z0-9/+=]{40}`) is not caught. The access-key ID (`AKIA...`) is scrubbed, but the far more dangerous half of the pair slips through. Recognize that a general 40-char pattern is false-positive prone (matches base64 chunks, hashes, random IDs), so this may need context-anchoring (e.g. `aws_secret_access_key\s*[:=]\s*...`) rather than a raw 40-char regex. Worth deciding whether AWS credentials are in scope and, if so, adding a safe pattern.
- **electron/utils/secretScrubber.ts** — GitHub token prefixes `ghu_` (user-to-server) and `gho_` (OAuth) are not in the alternation, while `ghs_` is. These do appear in `gh` CLI output and git-credential-manager logs. Adding them is safe (namespaced like `ghs_`).
- **electron/utils/secretScrubber.ts:107** — the OAuth-query-param pattern uses the `m` flag, so a log line starting with `code=42 not found` is redacted as an OAuth code regardless of URL context. Consider dropping `code` from this alternation, or anchoring it after `?`/`&`.

The rest of the PR is strong: `safe-regex2` assertions, bounded quantifiers, fail-closed `beforeSend` behavior, deep recursion with scalar leaves scrubbed past the depth cap, 134 tests.

These gaps need author judgment on scope, so the PR is excluded from this merge batch.

PR #5503, opened 2026-04-19
perf(github): virtualize PR/issue list and bound resource cache

### LINKED ISSUE #5412: Virtualize the PR and issue list popover
The PR/issue list popover mounts every loaded row. Cursor pagination loads 20 items per page, so "Load More" repeated 25 times on a 500-issue repo holds 500 rows in the DOM, each with its avatar, label badges, tooltip trigger, and worktree-match lookup. Scrolling is noticeably laggy on larger repos and the popover gets sluggish to interact with.

`react-virtuoso` is already a dependency and used in the logs viewer and event timeline, so the pattern is established.

Two adjacent cleanups worth folding into the same change:

The renderer cache backing the popover is a plain `Map` with no TTL, no size cap, and no eviction. It accumulates one entry per `projectPath:type:filterState:sortOrder` combination across the session and never lets go. The existing `TtlCache` utility was introduced for the same reason on tooltip caches in #3792 and would slot in cleanly.

The `IssueSelector` in the new-worktree dialog shows plain "Loading issues..." text while the list dropdown shows a proper skeleton. The skeleton components in `GitHubDropdownSkeletons` already cover this pattern and would give visual parity in the two places people pick issues.

Evidence:

- [`GitHubResourceList` row mounting](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/src/components/GitHub/GitHubResourceList.tsx#L803-L823) — `data.map` over the full loaded list, no windowing.
- [`githubResourceCache`](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/src/lib/githubResourceCache.ts#L1-L43) — unbounded `Map`.
- [`IssueSelector` loading state](https://github.com/daintreehq/daintree/blob/9f95aacf38ddfa55dad2c10eb5c8b51ddf0ca005/src/components/GitHub/IssueSelector.tsx#L139-L140) — plain text.

gregpriday:
## Summary

- Replaces the flat-rendered `GitHubResourceList` with a react-virtuoso virtual list (fixed item height 68px, stable item keys), so large PR/issue lists no longer mount hundreds of DOM nodes at once.
- Migrates `githubResourceCache` from an unbounded `Map` to a `TtlCache` (maxSize=20, ttl=5min) and caps the internal `generationMap` with the same FIFO limit, preventing unbounded memory growth across repo switches.
- Swaps the "Loading issues..." text in `IssueSelector` for the existing `GitHubResourceRowsSkeleton` component for visual consistency.

Resolves #5412

## Changes

- `GitHubResourceList.tsx` — virtual list via react-virtuoso; `Footer` slot drives Load More; `endReached` triggers auto-fetch; `VirtuosoHandle.scrollIntoView` preserves keyboard navigation.
- `githubResourceCache.ts` — bounded TtlCache replaces unbounded Map; generationMap capped at same size.
- `IssueSelector.tsx` — skeleton loader instead of plain text.
- `GitHubResourceList.swr.test.tsx` / `githubResourceCache.test.ts` — updated and extended tests (134 passing).

## Testing

Typecheck, lint, and format all clean. 134 unit tests pass across the affected files. Verified virtual rendering in the PR/issue popover with a repo holding 100+ open issues.

### DIFF
--- a/src/components/GitHub/GitHubResourceList.tsx
+++ b/src/components/GitHub/GitHubResourceList.tsx
@@ -40,6 +40,70 @@ function sanitizeIpcError(message: string): string {
   return cleaned.length > 120 ? cleaned.slice(0, 117) + "…" : cleaned;
 }
 
+interface LoadMoreFooterContext {
+  hasMore: boolean;
+  loadingMore: boolean;
+  isLoadMoreActive: boolean;
+  loadMoreError: string | null;
+  type: "issue" | "pr";
+  onLoadMore: () => void;
+  onOpenSettings: () => void;
+}
+
+function LoadMoreFooter({ context }: { context?: LoadMoreFooterContext }) {
+  if (!context || !context.hasMore) return null;
+  const { loadingMore, isLoadMoreActive, loadMoreError, type, onLoadMore, onOpenSettings } =
+    context;
+  return (
+    <div className="p-3 space-y-2">
+      {loadMoreError && (
+        <div className="p-2 rounded-[var(--radius-md)] bg-overlay-soft border border-[var(--border-divider)]">
+          <p className="text-xs text-muted-foreground">{sanitizeIpcError(loadMoreError)}</p>
+          {isTokenRelatedError(loadMoreError) ? (
+            <Button
+              variant="ghost"
+              size="sm"
+              onClick={onOpenSettings}
+              className="mt-1 text-muted-foreground hover:text-daintree-text h-6 text-xs"
+            >
+              <Settings className="h-3 w-3" />
+              Open GitHub Settings
+            </Button>
+          ) : (
+            <Button
+              variant="ghost"
+              size="sm"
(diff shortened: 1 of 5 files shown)


gregpriday:
Heads up: this introduced a regression — the issues/PR dropdowns render empty when there's data to show.

`<Virtuoso style={{ height: "100%" }} />` is mounted inside a `flex-1 min-h-0` wrapper whose only height constraint is `max-h-[500px]` on the outer container. Without an explicit height, `flex-1` has no space to distribute and the Virtuoso viewport resolves to 0px, so items fetch but never paint. The pre-virtualized `data.map` worked because items had intrinsic height.

Fix on `develop`: switch the dropdown (and matching skeleton) to `h-[500px]` so Virtuoso has a definite viewport.

PR #5512, opened 2026-04-19
feat(actions): add action.repeatLast bound to Cmd+Shift+.

### LINKED ISSUE #5409: Add run-last-action for keyboard-first workflows
For keyboard-first users, repeatedly invoking the same action is a common loop — fire an action, check the result, fire it again (often against a different target or with a small change). VS Code ships `workbench.action.reruncommand` bound to `⌃⇧;`. Daintree has no equivalent. The current workflow is `⌘⇧P`, retype, Enter, every time.

Evidence: `grep -r "repeatLast\|lastAction\|rerunCommand" src/` returns zero matches. The action registry has no "last dispatched" tracking.

Desired behavior:

- Capture the last dispatched `{actionId, args}` from user/menu/keybinding sources. Not from agent dispatch — agents have their own replay loops and shouldn't perturb user state.
- Re-evaluate args against **current** context on repeat. If the original `terminalId` is closed, fall back to the focused terminal or show a short "Original target gone — retargeted to focused" message. Frozen args cause ghost double-dispatches.
- Opt out of repeatability for actions that shouldn't repeat: palette opens (`action.palette.*`, `worktree.openPalette`, …), navigation (`nav.*`, `terminal.focus*`, `worktree.up/down/next/previous`), modal control (`modal.close`, `ui.escape`), and settings-open actions. Destructive actions (`danger: "confirm"`) should still hit their confirmation dialog on repeat — the confirmation is part of the action's contract.
- New action `action.repeatLast` (or equivalent ID). Natural binding is `⌘⇧.` — mirrors VS Code's `⌃⇧;` convention on a non-conflicting key.

The footgun to avoid: replaying a destructive action silently against the wrong target because args were captured at the original dispatch. Re-resolve panel IDs, worktree IDs, etc. at repeat time.

gregpriday:
## Summary

- Adds `action.repeatLast` (bound to `Cmd+Shift+.`) that re-dispatches the last eligible user/keybinding/menu/context-menu action via ActionService, mirroring VS Code's rerun-command pattern on a non-conflicting key.
- Introduces `nonRepeatable: boolean` on `ActionDefinition` to opt palette-openers, pure navigation, settings-open, and modal-control actions out of being tracked. `action.repeatLast` itself is marked nonRepeatable to prevent self-looping.
- Agent-source dispatches never become the last action. Args are `structuredClone`-snapshotted at capture time to prevent mutation leakage between dispatches.

Resolves #5409

## Changes

- `shared/types/actions.ts` — adds `action.repeatLast` to `ActionId` union; adds `nonRepeatable` field to `ActionDefinition`
- `shared/types/keymap.ts` — registers the `Cmd+Shift+.` keybinding entry
- `src/services/ActionService.ts` — adds `_lastAction` tracking with source and nonRepeatable filtering; implements `repeatLast` dispatch logic
- `src/services/actions/definitions/actionActions.ts` — new definition file for the `action.repeatLast` action
- `src/services/actions/definitions/{appActions,navigationActions,notesActions,panelActions,preferencesActions,projectActions,terminalNavigationActions,worktreeActions}.ts` — marks palette-opens, navigation, modal, and settings-open actions as `nonRepeatable`
- `src/services/defaultKeybindings.ts` — binds `action.repeatLast` to `Cmd+Shift+.`
- 50 unit tests covering capture ordering, nested-dispatch ordering, validation-fail skip, nonRepeatable opt-out, and end-to-end exercise of the registered action

## Testing

Unit tests run via `npm test`. 50 tests added across `ActionService.test.ts`, `actionDefinitions.test.ts`, and `actionActions.test.ts` covering all stated behaviours including edge cases (repeat-of-repeat guard, agent-source skip, structuredClone isolation, and target-gone scenarios).

### DIFF
--- a/shared/types/actions.ts
+++ b/shared/types/actions.ts
@@ -180,6 +180,7 @@ export type BuiltInActionId =
   | "worktree.overview.open"
   | "worktree.overview.close"
   | "action.palette.open"
+  | "action.repeatLast"
   | "actions.list"
   | "actions.getContext"
   | "actions.persistedStores"
@@ -348,6 +349,13 @@ export interface ActionDefinition<
    * copied verbatim (no further sanitization), so the allowlist is the policy.
    */
   safeBreadcrumbArgs?: readonly string[];
+  /**
+   * When true, `action.repeatLast` will not record this action as the last
+   * dispatched action. Use for palette-openers, pure navigation, modal control,
+   * and settings-open actions whose intent is transient/UI rather than a
+   * repeatable operation.
+   */
+  nonRepeatable?: boolean;
 }
 
 export interface ActionManifestEntry {
--- a/shared/types/keymap.ts
+++ b/shared/types/keymap.ts
@@ -167,6 +167,7 @@ export type BuiltInKeyAction =
   // Action palette
   | "action.palette"
   | "action.palette.open"
+  | "action.repeatLast"
 
   // Project actions
   | "project.switcherPalette"
@@ -334,6 +335,7 @@ export const KEY_ACTION_VALUES: ReadonlySet<string> = new Set<string>([
   "notes.openPalette",
   "action.palette",
   "action.palette.open",
+  "action.repeatLast",
   "project.switcherPalette",
   "project.mruCycleOlder",
   "project.mruCycleNewer",
--- a/src/services/ActionService.ts
+++ b/src/services/ActionService.ts
@@ -168,11 +168,11 @@ export class ActionService {
       if (
         REPEATABLE_SOURCES.has(source) &&
         !definition.nonRepeatable &&
-        definition.danger !== "confirm"
+        definition.danger === "safe"
       ) {
-        // danger:"confirm" actions rely on originating UI dialogs (e.g. WorktreeDeleteDialog)
-        // to gather consent — replaying them from a keybinding would bypass that UI and
-        // silently repeat a destructive operation. Exclude from capture entirely.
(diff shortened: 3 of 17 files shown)


gregpriday:

Looking at this more carefully before merging, there's a safety gap worth sorting out:

`ActionService.dispatch` only enforces the confirm-gate for `source === "agent"`, but `action.repeatLast` dispatches with `source: "keybinding"`. That means any `danger: "confirm"` action whose confirmation lives in the caller UI (not inside `run()`) gets silently replayed by Cmd+Shift+. without re-prompting.

Concrete hazards if a user deletes a worktree, confirms, then hits Cmd+Shift+. by accident:

- `worktree.delete`, `git.push`, `git.commit`, `project.remove`, `project.close`
- `terminal.closeAll`, `worktreeSessions.end/trash/delete`
- Portal `tab.closeAll/closeOthers/closeToRight`
- `recipes.promote`, `github.setToken/clearToken`
- `preferences.resetAgent/resetAllKeybindings`, `envVars.*`, `app.forceQuit`

Actions like `notes.delete` are safe because their confirm lives inside `run`, so the prompt re-appears on replay.

Might be simpler to just skip capture when `definition.danger !== "safe"` — that way repeatLast only ever repeats truly safe operations, and destructive ones have to be re-invoked through their normal UI. Alternative is to pass a flag that forces re-confirmation on repeat for confirm-danger actions, or audit every destructive action and add `nonRepeatable: true`, but the one-line danger-gate feels like the cleanest default.

Excluding from this merge batch — keen to hear which direction you'd prefer.

PR #5536, opened 2026-04-19
feat(settings): modernise Checkbox and Switch primitives on Radix

### LINKED ISSUE #5455: Modernise settings Checkbox and Switch primitives on Radix
`SettingsCheckbox` wraps a native `<input type="checkbox">`, so it inherits macOS blue / Windows Fluent / Linux GTK checkbox styling depending on platform. No brand identity, no themability, and focus styles don't match the rest of the app.

`SettingsSwitchCard`'s inner toggle is a hand-rolled button with `translate-x-6` motion and whatever ARIA got wired manually. It works, but re-implements what upstream primitives handle better.

Two related things we want:

- A checkbox that's branded across platforms, with proper indeterminate state, and consistent focus behaviour.
- A standalone `SettingsSwitch` primitive extracted from the card variant, so places that want a bare toggle (without the card chrome, icon, title, subtitle) aren't forced into `SettingsSwitchCard`.

Vercel Geist's switch spec (roughly 36×20 track, knob with a subtle drop-shadow, 150ms ease on the knob translate) is a good reference ([geist reference](https://vercel.com/geist)) but needs to stay token-driven so both dark and light themes get the right contrast. The `knob-base` token from the foundation issue covers the structural knob fill.

Depends on #5450.

gregpriday:
## Summary
- Replaced native `<input type="checkbox">` with Radix UI's Checkbox primitive for consistent cross-platform branding
- Extracted a standalone `SettingsSwitch` component from `SettingsSwitchCard` for use cases that need a bare toggle without card chrome
- Both components now use theme tokens for proper dark/light mode contrast and follow the Geist design reference

## Changes
- `SettingsCheckbox.tsx` — migrated to Radix Checkbox with proper indeterminate state and token-driven styling
- `SettingsSwitch.tsx` — new standalone switch component extracted from the card variant
- `SettingsSwitchCard.tsx` — refactored to use the new `SettingsSwitch` primitive
- Added comprehensive unit tests for both components covering interaction, keyboard navigation, and accessibility

## Testing
- Verified checkbox toggle, indeterminate state, and keyboard navigation
- Tested switch toggle with keyboard shortcuts and focus management
- Checked dark/light theme contrast for both components
- All unit tests passing

Resolves #5455


gregpriday:
Small regression worth flagging from this one. `Switch.Root` uses `inline-flex` without `items-center`, so the 16px thumb aligns to the top of the 24px track rather than sitting vertically centred. Subtle in isolation but reads as off across a list of toggles. Adding `items-center` to the class list in `SettingsSwitch.tsx:62` sorts it.

PR #5567, opened 2026-04-19
feat(settings): preset fields inherit from agent defaults with reset UI

### LINKED ISSUE #5556: Preset fields should inherit from agent defaults with reset UI
When editing a custom preset, the per-preset override fields (`dangerousEnabled`, `customFlags`, `inlineMode`) behave as independent settings rather than overrides on a base. The UI has no way to express "inherit from agent default". Once a user toggles Skip Permissions on a preset, it carries an explicit true/false value forever — there's no path back to "follow whatever the agent default is."

The preset env editor has the same problem: agent-level `globalEnv` keys that the preset isn't overriding aren't shown at all. A user who wants to override just one global key has to retype it from scratch because the editor treats preset env as a standalone map, not a delta on top of global.

## Why it matters

The data model and launcher already implement sparse-override semantics — a preset field left `undefined` falls through to the agent default. The sanitizer in `getMergedPresets` coerces invalid/missing fields to `undefined`. So "inherit" is already the serialization convention; the UI is what's missing.

Symptoms this will fix:
- User creates a preset for "longer timeouts", the preset forever claims to have an explicit Skip Permissions value they never intended to set.
- User updates the agent-level Skip Permissions default → existing presets silently don't change, because they have explicit values from the first time the toggle was touched.
- User wants to override one global env key for a preset → has to recreate it.

## Desired behavior

In custom preset detail:

- **Boolean overrides** (Skip Permissions, Inline Mode) become tri-state: `Inherit (Off)` / `On` / `Off`. The "Inherit" label displays the current agent-default value, so users don't have to hunt for it. Picking "Inherit" stores `undefined`.
- **String overrides** (Custom Arguments) get a small reset-to-inherit button visible only when the preset has an explicit value. Clicking clears the override back to `undefined`. The empty state shows the inherited value as placeholder text.
- **Env editor** gets an "inherited from global env" strip above the editable rows. Each agent-level `globalEnv` key that the preset isn't overriding shows as a grayed row with a "+ override" opt-in. Clicking seeds the preset env with the inherited key/value so the user can edit it.

No override indicator is shown for fields that inherit — absence is the signal (Windows Terminal pattern). Only explicitly-set fields show the reset affordance.

## Reference patterns

- Windows Terminal: circular-arrow "Reset to inherited value" icon, visible only when the field is overridden; no icon means inheriting.
- Research on tri-state boolean idioms: dropdown with `Inherit (True) / True / False` where the Inherit label shows the parent value. Single-checkbox-indeterminate is an anti-pattern (collides with "mixed selection" meaning).

## Dependency

Depends on #5555 (the Defaults section needs to exist before "Inherit" labels can reference a visible source).

## Evidence

- Launcher's merge semantics (`undefined` = inherit already works): https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/src/hooks/useAgentLauncher.ts#L252-L261
- Preset sanitizer coercing invalid/missing fields to `undefined`: https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/src/config/agents.ts#L150-L160
- Preset shape (fields already optional): https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/shared/config/agentRegistry.ts#L144-L166

## Edge case

Existing custom presets may carry explicit `dangerousEnabled: false` (or similar) written by the current UI when the user just meant "default". Under the new UI that reads as an explicit Off override, not an inherit. Rendered behavior is the same today — both show as off. Auto-migrating stale values to `undefined` is lossy (can't tell intentional from incidental). Leave existing data alone; users can clear overrides manually via the new "Inherit" option.

## Out of scope

Preset-to-preset inheritance chains, parent-preset selection. The single-layer model (agent default → preset) is sufficient for this issue.

gregpriday:
## Summary

- Boolean overrides (Skip Permissions, Inline Mode) in custom preset detail are now tri-state selects: Inherit/On/Off. The Inherit option shows the current agent-default value in its label so users can see what they're inheriting without hunting elsewhere.
- Custom Arguments gets a RotateCcw reset button that only appears when the preset holds an explicit value, clearing it back to `undefined` (inherit). The placeholder text shows the inherited value.
- The env editor gains a read-only "Inherited from global env" strip above the editable rows. Each `globalEnv` key not already overridden by the preset appears as a grayed row with a "+ Override" button that seeds the preset env with the inherited value for editing.

Resolves #5556

## Changes

- `AgentSettings.tsx`: tri-state `SelectValue` helper, `InheritBoolSelect` component for `dangerousEnabled`/`inlineMode`, `CustomArgsField` with reset affordance, `InheritedEnvStrip` component above preset env editor
- `selectValueToBool` hardened against option-value drift with exhaustive switch + fallback
- `AgentSettings.presetInherit.test.ts`: 161-line unit test suite covering the new components and helpers

## Testing

Unit tests added covering `selectValueToBool` round-trips, `InheritBoolSelect` label generation (Inherit shows parent value, On/Off explicit), `CustomArgsField` reset button visibility, and `InheritedEnvStrip` filtering (only shows keys not already in preset env).

### DIFF
--- a/src/components/Settings/AgentSettings.tsx
+++ b/src/components/Settings/AgentSettings.tsx
@@ -617,46 +617,87 @@ export function AgentSettings({
               // Only rendered inside the custom-preset detail panel. Reads
               // from the preset, falling back to agent-level defaults when
               // the preset omits an override. Writes via handleUpdatePreset.
+              //
+              // Booleans render as tri-state selects (Inherit/On/Off): a
+              // `boolean | undefined` override where `undefined` inherits the
+              // agent-level default at launch. Custom Arguments uses a text
+              // input with a reset button that clears the override.
 
-              const presetSkipPerms =
-                selectedPreset?.dangerousEnabled ?? activeEntry.dangerousEnabled ?? false;
+              const agentDefaultDangerous = activeEntry.dangerousEnabled ?? false;
+              const agentDefaultInline = activeEntry.inlineMode ?? true;
+              const agentDefaultCustomFlags = activeEntry.customFlags ?? "";
 
-              const presetInlineMode = selectedPreset?.inlineMode ?? activeEntry.inlineMode ?? true;
+              const dangerousOverride = selectedPreset?.dangerousEnabled;
+              const inlineOverride = selectedPreset?.inlineMode;
+              const customFlagsOverride = selectedPreset?.customFlags;
 
-              const presetCustomFlags =
-                selectedPreset?.customFlags ?? activeEntry.customFlags ?? "";
+              // Effective (merged) values — the dangerous-arg strip must
+              // reflect what actually gets passed to the CLI at launch.
+              const effectiveSkipPerms = dangerousOverride ?? agentDefaultDangerous;
 
-              const onPresetSkipPermsToggle = () => {
+              const onDangerousOverrideChange = (value: boolean | undefined) => {
                 if (!selectedPreset) return;
-                handleUpdatePreset(selectedPreset.id, { dangerousEnabled: !presetSkipPerms });
+                handleUpdatePreset(selectedPreset.id, { dangerousEnabled: value });
               };
 
-              const onPresetInlineModeToggle = () => {
+              const onInlineOverrideChange = (value: boolean | undefined) => {
                 if (!selectedPreset) return;
-                handleUpdatePreset(selectedPreset.id, { inlineMode: !presetInlineMode });
+                handleUpdatePreset(selectedPreset.id, { inlineMode: value });
               };
 
-              const onPresetCustomFlagsChange = (value: string) => {
+              const onCustomFlagsOverrideChange = (value: string) => {
                 if (!selectedPreset) return;
-                const updated = (activeEntry.customPresets ?? []).map((f) =>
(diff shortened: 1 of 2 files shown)


gregpriday:
Deferred from merge pipeline — rebase onto `develop` produced a structural conflict in `src/components/Settings/AgentSettings.tsx` after PR #5572 landed.

#5572 renamed the preset-level handlers (`skipPerms`/`inlineMode`/`customFlags` → `presetSkipPerms`/`presetInlineMode`/`presetCustomFlags`), removed the `selectedIsCustom ? … : SettingsSwitchCard` branching (the behavioral block is now custom-preset-only), and deleted the `isDefault` branch entirely. The tri-state inherit UI here needs to be re-applied on top of that new structure rather than the pre-hoist one.

Needs a manual rebase: re-target the tri-state selects at `presetSkipPerms`/`presetInlineMode`/`presetCustomFlags`, drop the custom/non-custom branching, and drop the `onAgentCustomFlagsChange` handler (agent-level editing lives in the hoisted defaults section now).

PR #5626, opened 2026-04-20
refactor(preset): unify env editor with inherited global env

### LINKED ISSUE #5604: Unify preset env editor with inherited global env
When editing a custom preset's env vars, env keys inherited from agent-level globals render in a separate bordered strip *above* the preset's env editor, each with a "+ Override" button to copy the key into the preset. Below that strip sits the actual editable env table with its own "+ Add" button.

The result is two similar-looking env tables with an unclear relationship. It's not obvious whether the top strip is inherited, overriding, or duplicated, and having "+ Override" on one table and "+ Add" on the other splits the mental model of what ought to be a single operation.

The end state should be one env table per preset scope. Inherited rows are visible but visually muted so it's obvious they come from the global default. Overriding an inherited key, adding a new preset-only key, and clearing an override back to inherit all happen inside the same table with consistent affordances.

One specific thing worth getting right: clearing an override should be an explicit action that drops the key and falls back to inherit, not just emptying the value field (which currently creates an explicit empty string override and silently overrides the default with `""`).

Evidence:

- The separate inherited strip and its "+ Override" buttons: [AgentSettings.tsx L1024-L1079](https://github.com/daintreehq/daintree/blob/245ac3ebe343c3462be7878f9a2c6e96c075c621/src/components/Settings/AgentSettings.tsx#L1024-L1079)
- The preset env editor that sits below it: [AgentSettings.tsx L1070-L1078](https://github.com/daintreehq/daintree/blob/245ac3ebe343c3462be7878f9a2c6e96c075c621/src/components/Settings/AgentSettings.tsx#L1070-L1078)
- Current `EnvVarEditor` has no awareness of inherited values: [EnvVarEditor.tsx](https://github.com/daintreehq/daintree/blob/245ac3ebe343c3462be7878f9a2c6e96c075c621/src/components/Settings/EnvVarEditor.tsx)

Runtime merge order is already `{ ...globalEnv, ...presetEnv }` (preset wins on key collision), so the data side already supports the model this UI should reflect.

Best landed after #5602 reshapes the surrounding structure, but the editor enhancement stands alone if it ships first. The `preset-env-inherited-strip` and `preset-env-inherited-override` testids disappear as part of this.

gregpriday:
## Summary
- Unified the preset env editor into a single table, removing the separate inherited strip
- Inherited global env keys now render as muted rows in the same table with clear visual distinction
- Added explicit override semantics: "Override" button copies inherited key with editable value, "Revert" button clears override back to inherited
- Fixed the empty string override issue by requiring explicit action to drop keys

Resolves #5604

## Changes
- `EnvVarEditor.tsx`: Added `inheritedEnv` prop, visual treatments for inherited rows, revert/override button logic
- `AgentSettings.tsx`: Removed the inherited env strip, passed `inheritedEnv` to the unified editor

## Testing
- Verified inherited keys render with muted styling and "Override" button
- Confirmed overriding creates an editable row with "Revert" button
- Tested that reverting drops the key and falls back to inherited value
- Confirmed emptying a value field no longer creates a silent empty string override

### DIFF
--- a/src/components/Settings/AgentSettings.tsx
+++ b/src/components/Settings/AgentSettings.tsx
@@ -1033,64 +1033,13 @@ export function AgentSettings({
                         <span className="text-[11px] text-daintree-text/50 font-medium uppercase tracking-wide block">
                           Env overrides
                         </span>
-                        {(() => {
-                          const globalEnv =
-                            (activeEntry.globalEnv as Record<string, string> | undefined) ?? {};
-                          const presetEnv = selectedPreset.env ?? {};
-                          const inheritedEntries = Object.entries(globalEnv).filter(
-                            ([k]) => !(k in presetEnv)
-                          );
-                          if (inheritedEntries.length === 0) return null;
-                          return (
-                            <div
-                              className="rounded-[var(--radius-md)] border border-daintree-border overflow-hidden bg-daintree-bg/20"
-                              data-testid="preset-env-inherited-strip"
-                            >
-                              <div className="grid grid-cols-[2fr_3fr_auto] text-[10px] uppercase tracking-wide text-daintree-text/50 bg-daintree-bg/40 border-b border-daintree-border">
-                                <div className="px-2.5 py-1.5">Inherited from global env</div>
-                                <div
-                                  className="px-2.5 py-1.5 border-l border-daintree-border/60"
-                                  aria-hidden="true"
-                                />
-                                <div className="px-2.5 py-1.5 w-[88px]" aria-hidden="true" />
-                              </div>
-                              <div className="divide-y divide-daintree-border">
-                                {inheritedEntries.map(([key, value]) => (
-                                  <div
-                                    key={key}
-                                    className="grid grid-cols-[2fr_3fr_auto] items-center font-mono text-[12px] text-daintree-text/50"
-                                  >
-                                    <span className="px-2.5 py-2 truncate">{key}</span>
-                                    <span className="px-2.5 py-2 border-l border-daintree-border/60 truncate text-daintree-text/40">
-                                      {value}
-                                    </span>
-                                    <div className="flex items-center justify-center w-[88px] border-l border-daintree-border/60">
-                                      <button
-                                        type="button"
-                                        className="text-[10px] px-2 py-1 rounded text-daintree-accent hover:text-daintree-accent/80 hover:bg-daintree-bg/60 transition-colors"
-                                        onClick={() =>
-                                          handleUpdatePreset(selectedPreset.id, {
-                                            env: { ...presetEnv, [key]: value },
-                                          })
-                                        }
-                                        aria-label={`Override ${key} in this preset`}
-                                        data-testid="preset-env-inherited-override"
-                                      >
-                                        + Override
-                                      </button>
-                                    </div>
-                                  </div>
-                                ))}
-                              </div>
-                            </div>
-                          );
-                        })()}
                         <EnvVarEditor
                           env={selectedPreset.env ?? {}}
                           onChange={(env) => handleUpdatePreset(selectedPreset.id, { env })}
                           suggestions={getAgentConfig(activeAgent.id)?.envSuggestions ?? []}
                           datalistId="env-key-suggestions"
                           contextKey={selectedPreset.id}
+                          inheritedEnv={activeEntry.globalEnv as Record<string, string> | undefined}
                           data-testid="preset-env-editor"
                         />
                         {envVarReference}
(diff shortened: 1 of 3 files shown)


gregpriday:

This PR was reviewed as part of an automated merge pipeline and has issues that need to be addressed before merging.

### Test coverage gap
- **`src/components/Settings/__tests__/EnvVarEditor.test.tsx`** — no tests were added for the new inherited-rendering, override, revert, or empty-string override paths. The empty-string fix specifically has no regression test. The existing suite only exercises the pre-refactor API.

### Effect perf / stability
- **`src/components/Settings/EnvVarEditor.tsx`** — the sync effect compares `inheritedEnv` via `JSON.stringify` on every render. `AgentSettings.tsx` passes `globalEnv ?? {}`, which allocates a fresh object identity per render when `globalEnv` is undefined, so this path runs every render. Use a shallow key+value compare (same pattern as the existing env comparison) to avoid wasted work and a potential loop risk given `rows` is in the dep array.

### Rebase coordination with #5622
- PR #5622 rewrites `EnvVarEditor.tsx` into a bordered grid table (`grid grid-cols-[2fr_3fr_auto]`, `before:` pseudo-stripe system, `isEmpty` CTA, reveal toggle, focus recovery). This PR's `border-l-[3px]` inline stripe, flex-row layout, and disabled-input patterns don't compose with that structure. A clean rebase will need to:
  1. Layer `inheritedEnv` / `isInherited` onto #5622's draft model.
  2. Replace the inline stripe with a fourth branch in #5622's `before:` stripe system (so override composes with empty/duplicate/secret stripes).
  3. Move the `+ Override` / `RotateCcw` revert controls into #5622's grid column layout.
  4. Reconcile the `isEmpty` placeholder (never reached once inherited rows synthesize) with the "no env vars defined" message this PR adds.

### Minor issues (can be folded in alongside)
- **`EnvVarEditor.tsx`** — inherited rows are produced via `Array.from(allKeys).sort()`, which reorders existing override rows alphabetically on each reseed. Previous behavior preserved `Object.entries(env)` insertion order.
- **`EnvVarEditor.tsx`** — `handleValueChange` unconditionally clears `isInherited`; safer to pair with `handleValueBlur`'s revert logic so a cleared value reverts to inherited without silently converting to an empty override.
- **`EnvVarEditor.tsx`** — `+ Override` button sits inside the value input's right padding area with no padding-right on the input; long inherited values render under the button text.

These issues need manual attention — the PR has been excluded from the current merge batch. Recommend landing #5622 first, then re-approaching this on top of the new table structure.

gregpriday:
Pushed a force-update that rebases onto develop (now includes #5622's bordered-table editor) and addresses each of the review findings. Collapsed into a single commit since the old two were written against the pre-#5622 flex-row layout.

**Rebased on top of #5622's table.** The inheritance layer is now built into the new draft-row model rather than bolted on. `envToDraft` takes an optional `inheritedEnv` and synthesizes inherited rows after the override rows. `draftToEnv`, `findDuplicateKeys`, and `isValid` all skip inherited rows so they don't leak into commits or validation.

**Stripe system.** Override rows get a fourth branch in the existing `before:` stripe chain (accent), sitting alongside error/duplicate/secret. No more `border-l-[3px]` inline stripe.

**Actions cell.** `+` Override and `RotateCcw` revert live in the grid's action column now, so the earlier problem of long inherited values rendering under the button text is gone. Remove on an override that shadows an inherited key falls back to inherited rather than wiping the row (otherwise the inherited entry would immediately reappear as a separate row).

**Shallow compare instead of `JSON.stringify`.** Both `env` and `inheritedEnv` now go through a key+value shallow compare, so the fresh `{}` identity that `AgentSettings` allocates per render no longer thrashes the reseed effect. `AgentSettings` also passes `inheritedEnv={activeEntry.globalEnv}` directly now (no `?? {}`), since `undefined` and `{}` are equivalent under the new compare.

**Insertion order preserved.** Overrides come first in `env`'s insertion order, then inherited-only keys in `inheritedEnv`'s insertion order. No more alphabetical re-sort.

**`handleValueChange` no longer touches `isInherited`.** Inherited inputs are disabled, so value-change can't fire from them, and the inherit→override transition goes through the explicit Override button. The empty-string guard in `handleValueBlur` stays: clearing an override's value reverts to inherited so we don't silently ship `KEY=""` masking the inherited value.

**Root-caused a stomp bug while writing tests.** The old `commitIfValid` updated `lastEnvRef` to the synthesized env before the parent had echoed the commit back. On the next effect pass, `env` (prop, still the stale parent value) mismatched `lastEnvRef` (our committed value), which triggered a reseed that reset the row back to inherited. `lastEnvRef` now tracks the last seen prop only, and the prop echo updates it on the next effect.

**Tests.** 12 new cases cover inherited rendering, override-promote and commit, revert, the empty-string guard, preset-only vs shadowed clear, insertion order, inheritance-only empty state, shallow-compare stability across fresh `{}` identities, and reseed on inheritance change. 37/37 editor tests pass, 403/403 across `src/components/Settings`, plus typecheck, lint (down 7 warnings), and format all clean.

Issue #5553, opened 2026-04-19
Notes panel fails to save or restore content


gregpriday:
Notes panels can't save or restore content. Typing in the editor doesn't persist to disk, and restored panels don't reload their content — the feature is effectively non-functional for end users.

The storage layer itself looks correct: `NotesService` writes YAML+markdown atomically and is covered by unit tests (`electron/services/__tests__/NotesService.test.ts`), and the snapshot round-trip for panel metadata (`notePath`, `noteId`, `scope`, `createdAt`) is covered by `src/panels/__tests__/registry.roundtrip.test.ts` and the "notes panels" case in `panelPersistence.test.ts`. So the break is somewhere between those correct pieces and the running app — either in how the main-process service is wired per project, in the save/load lifecycle inside `NotesPane`/`useNoteEditor`, or in how panel hydration threads `notePath` through `createNotesDefaults`.

Two code paths look suspicious during investigation and are worth verifying first:

- The notes IPC handler holds a module-level `NotesService` singleton keyed to `projectStore.getCurrentProject()` — a global DB lookup, not per-window. After the Apr 19 multi-window changes (#5492, commits `f6fc21f6d` and `7b10c9e24`) decoupled new windows from global project state, this may return null or the wrong project, so every notes IPC call either throws "No active project" or writes to the wrong directory. See [`electron/ipc/handlers/notes.ts`](https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/electron/ipc/handlers/notes.ts#L53-L66).

- Both `NotesPane` and `useNoteEditor` silently no-op when `notePath` is empty or when `metadata` hasn't loaded yet — no error surfaced, no console warning. If the initial `notesClient.read` fails (e.g. for the reason above), `metadata` stays `null` and every subsequent keystroke silently skips the save. `createNotesDefaults` defaults `notePath` to `""` if missing, which trips the same early-return. See [`src/components/Notes/NotesPane.tsx`](https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/src/components/Notes/NotesPane.tsx#L203-L283) and [`src/panels/notes/defaults.ts`](https://github.com/daintreehq/daintree/blob/a1121f87e99f22f2a5ecf8b884249b1631df99a5/src/panels/notes/defaults.ts#L4-L11).

## Expected behavior

[...]

Issue #5119, opened 2026-04-10
Version 0.6.0 - Cannot save the content of 'Note'


TrinhNhatHung:
The 'Note' feature is currently broken and cannot be used. I edited the content, but it cannot be saved.

https://github.com/user-attachments/assets/0afca4bc-8beb-40bb-8d35-f0e0d739c95d

TrinhNhatHung:
Sometimes previously saved notes also get cleared unexpectedly.

<img width="1205" height="792" alt="Image" src="https://github.com/user-attachments/assets/a1aab629-2577-4853-b515-aab67174aa6a" />

gregpriday:
Good catch, and sorry about the lost notes.

The video is pretty clear: content gets typed, palette closes, reopen, gone. That's a real save bug, not just a display thing, and the second comment about previously saved notes clearing is the same behaviour from the other angle. The palette editor has a debounced write plus a flush-on-close that should handle this, so something in `useNoteEditor` is dropping writes on the way out. Not fully root-caused yet, but it's reproducible from the frames and I'll dig into it properly.

Two bits of context worth flagging while you're testing v0.6.0:

Notes storage moved from `{project}/.canopy/notes/` to `{userData}/notes/{projectId}/` in this build (on macOS that's `~/Library/Application Support/Canopy/notes/{projectId}/`). v0.6.0 is a pre-release going out to internal testers, and I skipped the auto-migration step on purpose for this round. If you had older notes from v0.5.x, they're still on disk at the old path. Nothing has been deleted, and you can copy the `.md` files across manually if you want them back.

The "Untitled / Empty note" shown in the comment screenshot is the palette's auto-create behaviour: when the notes directory is empty on open, it spawns a fresh note for you, which on top of the save bug makes it feel like the existing content was wiped.

Marking this as the fix target for the save bug. Thanks for the thorough repro.

PR #5629, opened 2026-04-20
feat(notes)!: remove Notes panel feature

### LINKED ISSUE #5616: Remove Notes panel feature
The Notes panel should be removed from core. It's ~3,000 lines of bug-prone code that doesn't serve Daintree's orchestration mission, and its main agent-integration hook is architecturally broken.

## Why

**The save bug is architectural, not a tractable defect.** CodeMirror's internal document state + React state + debounced autosave + mtime-based external-conflict detection is a distributed-systems consensus problem squeezed into a text input. Every few releases it breaks differently. #5119 ("Cannot save the content of Note") was followed by #5553 ("Notes panel fails to save or restore content") within the same release cycle. We've been papering over this category of race condition since the feature shipped.

**The `@.daintree/notes/<path>` copy button is a dead link.** Notes live in `userData/notes/{projectId}/`, which on macOS is something like `~/Library/Application Support/Daintree/notes/.../`. Agents in Daintree run with CWD set to the worktree, so an agent asked to resolve `@.daintree/notes/foo.md` can't find it. The one feature that justified an in-app notes panel for an agent orchestrator was never actually wired up end-to-end.

See [`electron/services/NotesService.ts` line 98-100](https://github.com/daintreehq/daintree/blob/06c7dbc2a37c0520be940b301d39d4dda78ad56f/electron/services/NotesService.ts#L98-L100) for the storage path, and [`src/components/Notes/NotesPane.tsx` line 362-364](https://github.com/daintreehq/daintree/blob/06c7dbc2a37c0520be940b301d39d4dda78ad56f/src/components/Notes/NotesPane.tsx#L362-L364) for the addressable-path generation with no resolver anywhere else in the codebase.

**Daintree is a terminal manager, not a document store.** The feature has steadily drifted toward being its own app-within-an-app. A successor to the scratchpad use case, if one is ever needed, will come from a clearer product direction, not as a backfill of what's being removed.

## Scope

Remove the Notes panel kind entirely. That includes the panel component, `NotesService`, all notes IPC channels and handlers, action definitions, the toolbar button, the terminal context-menu "Create note" entry, the notes palette, the voice-to-note hook, panel registry and serializer entries, and associated tests.

Drop the deps that become orphaned: `@uiw/react-codemirror`, `@codemirror/lang-markdown`, `@codemirror/language-data`, `gray-matter`, `remark-gfm`. Verify `nanoid` before removing (likely used elsewhere).

## Constraints

**Auto-export user data before deletion.** On first launch post-removal, copy `userData/notes/{projectId}/*.md` into `app.getPath('documents')/Daintree Notes Export/<projectName>/`, rename `userData/notes` to `userData/notes_archived` so the export doesn't re-run, and fire a single notification telling the user where the files went. No blocking dialog on startup.

**Migrate persisted panel state.** Bump the persist version on the panel and layout stores and strip any `kind: "notes"` entries in the migration step. Silently drop rather than showing tombstone placeholders. Users who had a Notes panel in their saved layout should get a clean workspace on upgrade.

**One release, hard remove.** Pre-1.0, no soft-deprecation cycle needed.

## Related

- Recent save/restore bugs: #5553, #5119, #5372, #5370, #4398, #4396, #4397
- Earlier storage migration that moved notes out of the project dir: #4399
- The `.daintree/notes/` line still present in `.gitignore` is a leftover from that migration

gregpriday:
## Summary

- Removes the Notes panel entirely: `NotesService`, all IPC channels/handlers, renderer components (`src/components/Notes/`, `src/panels/notes/`), hooks, store, action definitions, toolbar button, and context menu entries.
- The save-race bug was architectural (CodeMirror state + React state + debounced autosave + mtime conflict detection), not a tractable defect. It broke twice in the same release cycle (#5553 after #5119). Removing is cleaner than patching it again.
- The `@.daintree/notes/<path>` agent integration was never wired up end-to-end: notes live in `userData/notes/{projectId}/` but agents run with CWD set to the worktree, so the path was never resolvable.

Resolves #5616

## Changes

- Deletes `NotesService`, IPC handler `notes.ts`, all `notes.*` IPC channels, and the `notes` namespace from `preload.cts`
- Removes `src/components/Notes/`, `src/panels/notes/`, `src/hooks/useNote*.ts`, `src/store/notesStore.ts`, `src/clients/notesClient.ts`, and `src/services/actions/definitions/notesActions.ts`
- Drops `gray-matter`, `remark-gfm`, and `@codemirror/lang-markdown` from `package.json`
- Moves `editorTheme.ts` and `editorSearchTheme.ts` to `src/components/FileViewer/` (still used by `CodeViewer`)
- Bumps `toolbarPreferencesStore` persist version 5 to 6, with a migration that strips `"notes"` from saved button lists
- Adds migration `018-archive-notes.ts`: on first launch post-removal, renames `userData/notes` to `userData/notes_archived` and fires a notification pointing users to the archived path
- Guards state hydration to skip persisted panels with unregistered kinds, preventing ghost panels for users who had a Notes panel in their saved layout
- Removes `e2e/core/core-notes-panel.spec.ts` and all unit tests for notes-related code

## Testing

13,047 unit tests pass. Typecheck, lint, and format are clean. Hydration guard is covered by new unit tests in `src/utils/__tests__/stateHydration.test.ts`. `toolbarPreferencesStore` migration is covered in `src/store/__tests__/toolbarPreferencesStore.test.ts`.


gregpriday:
Quick note on the rationale here: saving had been broken for a while and I'd taken a few runs at fixing it. It's fixable, but it started feeling like a bloat feature. There are hundreds of good note-taking apps out there, and I don't think Daintree needs one baked in.

This is a prime candidate for a plugin. Once the plugin system is finalised, the Notes plugin will take everything we've removed here and build it out properly with the saving issues sorted.

Right now the focus is on core: the terminal manager for agent terminals, the dev preview, the worktree workflow, and panels. Plugin system comes after that's solid.

PR #5760, opened 2026-04-21
feat(fleet): multi-terminal input micro-animation polish

### LINKED ISSUE #5700: Multi-terminal input micro-animation polish
Motion polish to apply once the broadcast redesign lands (#5743, #5745, #5746, #5747, #5748, #5749, #5750). Do this as a dedicated follow-up rather than folding into the core work — tuning motion is easier once the interactions themselves are settled.

Philosophy: every animation here clarifies what just happened without adding perceived latency. Sub-300ms across the board. Compositor-only properties (transform, opacity, shadow). Reduced-motion removes transforms and stagger, keeps opacity fades.

## Bar entrance/exit

The broadcast bar appears when `armedCount` crosses from 1 to 2 and disappears when it drops below 2. Slide-down + fade on entry (~200ms), faster slide-up + fade on exit (~120ms). Gives the bar a felt "selection engaged" moment instead of just popping in. Use a snappy spring (roughly `stiffness: 300, damping: 30`) — confident, not theatrical.

## Count chip bump on change

When the selection count changes (handle click, range extend, scoped ⋯-menu command), the count text should swap with a 150ms fade + translate. Makes multi-select gestures feel acknowledged.

## Title-bar selection indicator transition

When a pane becomes selected or the primary shifts (new primary gains the full-opacity border + elevated title-bar; old primary transitions to the member styling), crossfade the border and title-bar treatment over ~150ms. This is the single highest-value motion in the feature — it shows the user the thread of where focus will land on exit.

## Primary-pane focus-restoration pulse

When the selection is cleared and focus returns to the primary pane (`lastArmedId`), fire a ~200ms ring pulse on that pane so the user sees where focus landed. Event, not state — it only plays once.

## Menu-item preview highlight

Hovering a scoped selection item (e.g., "Select Waiting Agents" in a worktree ⋯ menu, or the global ⋯ menu's selection commands) should highlight the panes that **would** be selected with the title-bar tint at roughly 30% opacity. 100ms fade-in on hover, instant on hover-off. This is what makes bulk-select menus usable rather than a coin-flip — it's a preview-as-you-browse affordance.

## Focus-loss pulse

When the app window regains focus and the selection is non-empty, pulse the broadcast bar's border once (~800ms) to defuse forgotten-mode mistakes. This is the single most important animation for the live-echo safety story.

## Commit flash on send

When a keystroke or paste is committed to targets, a ~100ms opacity flash on the bar's input edge confirms "it went." Subtle — this is confirmation feedback, not celebration.

## Out of scope for this issue

- Anything that removes the bar itself or changes gesture behavior — those live in the referenced prerequisite issues.
- The diagonal-stripe overlay and related visual treatments — those are deleted by #5745, not polished here.
- Stagger/ripple effects across the selection set — explicitly excluded; the selection operates as a uniform set and a stagger would misrepresent that.

Depends on: #5743, #5745, #5746, #5747, #5748, #5749, #5750.

gregpriday:
## Summary

- Adds commit flash, exit pulse, and selection-menu preview affordances to the fleet broadcast bar, rounding out the motion layer for multi-terminal input
- Commit flash fires a brief opacity pulse on the bar's input edge whenever a keystroke or paste is dispatched to armed targets, confirming the send without fanfare
- Exit pulse fires a single ring animation on the primary pane when the selection clears, showing exactly where focus landed
- Menu-item preview highlights the panes that *would* be selected when hovering scoped selection commands, making bulk-select menus legible rather than a coin-flip

Resolves #5700

## Changes

- `fleetArmingStore.ts` — added `commitFlashSeq` (monotonic counter) and `previewArmedIds` (set for hover previews) to the arming store
- `useFleetLiveKeyCapture.ts` — wired `onSend` callback ref so call sites can trigger imperative animations without forcing effect re-subscription on every render
- `FleetComposer.tsx` — connects `onSend` to `triggerCommitFlash`, clears preview on menu close
- `FleetArmingRibbon.tsx` — renders commit flash and exit pulse animations; subscribes to `previewArmedIds` and `commitFlashSeq`
- `ContentPanel.tsx` — applies preview tint to panes in `previewArmedIds`
- `index.css` — keyframe definitions for `fleet-commit-flash` and `fleet-exit-pulse`
- Three test files covering store shape, ribbon animation triggers, and composer preview lifecycle (including negative-path cases)

## Testing

Unit tests cover the commit flash counter increment, preview set population/clearing, ribbon rendering under flash/pulse/preview states, and the composer menu-close clear path. All pass. Typecheck and lint clean.

### DIFF
--- a/src/components/Fleet/FleetArmingRibbon.tsx
+++ b/src/components/Fleet/FleetArmingRibbon.tsx
@@ -5,7 +5,14 @@ import { useShallow } from "zustand/react/shallow";
 import { cn } from "@/lib/utils";
 import { isMac } from "@/lib/platform";
 import { useEscapeStack } from "@/hooks";
-import { useFleetArmingStore, type FleetArmStatePreset } from "@/store/fleetArmingStore";
+import {
+  useFleetArmingStore,
+  collectEligibleIds,
+  isFleetArmEligible,
+  type FleetArmStatePreset,
+  type FleetArmScope,
+} from "@/store/fleetArmingStore";
+import { useWorktreeSelectionStore } from "@/store/worktreeStore";
 import { useWorktreeFilterStore } from "@/store/worktreeFilterStore";
 import {
   useFleetPendingActionStore,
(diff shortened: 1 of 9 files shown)


gregpriday:
Closing because of significant changes to functionality.

PR #5800, opened 2026-04-22
feat(pty): policy for spawn-sealed agent behaviours on runtime-promoted panels

### LINKED ISSUE #5776: Policy for spawn-sealed agent behaviours on runtime-promoted panels
Several agent-vs-terminal behaviours are decided at PTY spawn time and cannot be reconfigured live. When a plain terminal runs an agent at runtime, it inherits none of them:

- **Environment variables.** Agent spawns set `FORCE_COLOR=3`, `COLORTERM=truecolor`, `CI=1`, `NONINTERACTIVE=1`, plus per-agent exclusions (gemini and opencode need `CI` unset because their ink/termenv renderers degrade to ASCII, claude needs `CLAUDECODE` unset, etc.). A promoted shell misses all of this — TUI agents can render broken output.
- **PTY pool bypass.** Agents cold-spawn; terminals acquire from a warm pool with a stripped env. A promoted pooled terminal has the wrong env and cannot be fixed without respawn.
- **Scrollback size.** Agents get 10k lines, terminals 1k. xterm allows shrinking but not growing a live scrollback buffer — a promoted terminal keeps its small buffer and agent output overflows.
- **Shell exec path, graceful shutdown config, session persistence snapshots.** All set at spawn.

This issue is the policy decision before any implementation:

1. **Restart-required parity.** Promoted panels run in degraded mode until restarted as a proper agent. UI surfaces "Restart for full agent support." Cheapest to build, worst UX.
2. **In-PTY repair on promotion.** Write `export` lines for env into the PTY (only works for env, not for pool/scrollback), resize scrollback by dropping to the smaller value only, accept pool mismatch. Medium cost, medium quality.
3. **Transparent respawn on promotion.** When detection fires, transparently respawn the PTY with agent settings, replaying the command. Full parity, highest cost and risk (breaks in-flight command, loses scrollback).

Pick a policy, then implement. The policy should also cover the reverse case (agent panel that demotes to plain shell) — most of these behaviours don't need to be reverted, but the scrollback decision at least deserves thought.

Depends on #5770, #5771, #5772, #5773, #5774, #5775.

gregpriday:
## Summary

- Adopts Policy 1 (restart-required UX) with one in-process repair: on first runtime promotion of a `kind="terminal"` panel, detection scrollback is grown to 10k lines via a new `applyAgentPromotion` method on `TerminalInstanceService`.
- Policy 2 was rejected because env vars written into a live PTY cannot propagate to an already-running agent child process. Policy 3 was rejected because a transparent respawn would SIGHUP the in-flight agent.
- A non-blocking "degraded mode" warning banner gives users a one-click "Restart as Agent" path via the existing `convertTerminalType` flow. The banner is gated on `spawnKind="terminal"` + `everDetectedAgent` + `detectedAgentId`, hides on exit/restart/dismiss, and resets its dismissal state when the detected agent identity changes so that switching from Claude to Gemini in the same shell still prompts.

Resolves #5776

## Changes

- `electron/services/pty/TerminalProcess.ts` — wires `onAgentDetected` listener to call `applyAgentPromotion` on promotion, growing scrollback in-process.
- `src/services/terminal/TerminalInstanceService.ts` — new `applyAgentPromotion(id, type)` that grows the live xterm scrollback buffer to the agent limit.
- `src/components/Terminal/TerminalDegradedModeBanner.tsx` — new warning banner component with dismiss and restart-as-agent actions.
- `src/components/Terminal/TerminalPane.tsx` — mounts the banner above the xterm viewport when the panel is in degraded mode.
- `src/components/Terminal/restartStatus.ts` — extends banner variant logic to cover all degraded-mode states.
- `src/store/panelStoreListeners.ts` — wires up the listener that triggers scrollback repair and banner visibility on agent detection events.
- `src/utils/panelProps.ts` — exposes `spawnKind` on derived panel props for downstream consumption.

## Testing

16 new unit tests covering scrollback growth (5 cases), banner variant logic (9 cases), and listener wiring (3 cases). All targeted tests pass. Full verify clean: typecheck, lint, format, channel drift, and build all green.


gregpriday:
Deferred from the merge pipeline — needs a discriminator redesign after #5801 landed.

#5801 removed `"agent"` from `BuiltInPanelKind` (now `"terminal" | "browser" | "dev-preview"`) and replaced `kind === "agent"` checks with `Boolean(agentId)` across the codebase. Git rebase onto `develop` completes without textual conflicts, but the resulting code is semantically broken in three places where this PR's "spawn-sealed vs runtime-promoted" logic uses `kind === "terminal"` as its discriminator:

- `src/components/Terminal/restartStatus.ts:59` — `getDegradedBannerVariant` gates on `input.kind !== "terminal"` to separate cold-spawned agents (no banner) from promoted terminals (show banner). Post-#5801 no PTY panel has `kind: "agent"`, so the gate never trips the way it used to.
- `src/store/panelStoreListeners.ts:281,313` — `priorKind === "terminal"` gate for `applyAgentPromotion` — always true now.
- `src/components/Terminal/TerminalPane.tsx:740-742,881` — hardcodes `const kind = "terminal" as const` as the `spawnKind` into the degraded banner, so the spawn-sealed branch can never be taken.
- `src/components/Terminal/__tests__/restartStatus.test.ts:114` — sets `kind: "agent"` in a fixture that's no longer a valid `BuiltInPanelKind` (typecheck will fail on rebase).

The fix is a new persisted spawn-time signal — e.g. `spawnAgentId` alongside the runtime `agentId`, or a `wasColdSpawnedAsAgent: boolean` flag — so "spawn-sealed" can still be distinguished from "runtime-promoted" now that `kind` doesn't encode it. That's a design call, not a mechanical merge, so leaving it for you.

PR #5882, opened 2026-04-27
fix(tooltip): hoist single TooltipProvider to App root, fix shared delay group

### LINKED ISSUE #5850: Tooltip delays vary across surfaces, breaking shared delay group
The renderer mounts roughly 77 separate `TooltipProvider` instances scattered across components, each with its own delay configuration. The sidebar uses a 400ms initial delay, but most other surfaces fall back to Radix's default 700ms. Hovering from the sidebar across to a panel header feels inconsistent.

More importantly, Radix's shared-delay-group behaviour (where the second tooltip in a group opens instantly after the first) only works within a single provider. Today every surface acts like its own island, so the second tooltip you hover always waits the full delay rather than feeling snappy.

Hover should feel coordinated across the whole window the way VS Code and Linear do, with the second-tooltip-instant behaviour working between any two icon buttons regardless of which surface they live on.

Evidence: searching for `<TooltipProvider` in the renderer turns up around 77 instances. Most should be removed in favour of one mounted near the app root with sensible delays (something like `delayDuration={500} skipDelayDuration={150}`).

gregpriday:
## Summary

- Radix's shared-delay-group behaviour (where the second tooltip opens instantly after the first) only works within a single provider. With ~77 separate `TooltipProvider` instances scattered across the renderer, every surface was acting as its own island and the second-tooltip fast-open never fired.
- Hoisted a single `TooltipProvider` with `delayDuration={500} skipDelayDuration={150}` to the App root and removed all 58 per-component providers, so the shared delay group now spans the whole window.
- Added `TooltipProvider` to the Radix mock in 7 test files that were rendering tooltip components without a provider context.

Resolves #5850

## Changes

- `src/App.tsx` — single root `TooltipProvider` wrapping the full app tree
- 58 components across Layout, Panel, Sidebar, Terminal, Browser, GitHub, Settings, Worktree, and more — removed local `TooltipProvider` wrappers and their imports
- 7 test files — added `TooltipProvider` to the `@radix-ui/react-tooltip` mock so components render cleanly in tests

## Testing

Hoisting verified manually: hovering between icon buttons on different surfaces (sidebar, toolbar, panel header) now uses the fast-path correctly after the first tooltip opens. Unit tests pass with the updated mocks.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -114,6 +114,7 @@ import { OnboardingFlow } from "./components/Onboarding/OnboardingFlow";
 import { GettingStartedChecklist } from "./components/Onboarding/GettingStartedChecklist";
 import { CelebrationConfetti } from "./components/Onboarding/CelebrationConfetti";
 import { ErrorBoundary } from "./components/ErrorBoundary";
+import { TooltipProvider } from "./components/ui/tooltip";
 import { DndProvider } from "./components/DragDrop";
 import {
   usePanelStore,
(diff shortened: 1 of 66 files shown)


gregpriday:
Closing — rebase against current develop produced too many tangled conflicts (recent UI PRs added new local TooltipProvider wrappers in the same files this PR removes). Cleaner to redo against current develop: sweep src/ for local TooltipProvider usages, remove them, verify the App-root provider covers all callers.

Issue #5902, opened 2026-04-27
Improve agent button right-click context menu


gregpriday:
Right-click on an agent toolbar button currently exposes launch shortcuts (Launch / Launch in Dock / Launch with Preset / Launch in Worktree) plus Unpin and Settings. Two friction points worth fixing together since both live in the same context menu.

**No way to manage presets from the right-click menu.** The only management entry is `${name} Settings...` which dumps the user into the agent's generic settings tab. Right-click is the natural place to jump straight to preset editing — most users right-clicking an agent button want to either launch it or tweak its presets, not visit the broader settings UI. There's room for a dedicated entry that opens the preset editor for this specific agent.

**"Launch in Worktree" nests three submenus deep.** The flow is right-click → Launch in Worktree → \<worktree name\> → Grid / Dock. Three cascading hover-flyouts means the user has to steer the mouse across two parent rows without overshooting, with the steering penalty compounding at each level. The Grid/Dock choice doesn't really earn its own submenu — it could collapse into a default plus a modifier (Alt-click for dock), or render inline on each worktree row similar to the inline pin button already in use elsewhere in the agent tray ([`AgentTrayButton.tsx#L693-L714`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentTrayButton.tsx#L693-L714)).

Both the with-presets and no-presets branches of the menu render the same nested worktree picker and both lack the preset-management entry:

- With-presets branch: [`AgentButton.tsx#L473-L575`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L473-L575)
- No-presets branch: [`AgentButton.tsx#L244-L324`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L244-L324)

Desired end state: right-click is the obvious path to managing this agent's presets (not a detour through Settings), and the worktree-launch picker is one hover deep, not three.

Issue #5901, opened 2026-04-27
Surface the active preset on agent toolbar buttons


gregpriday:
The toolbar agent button uses MRU/worktree-scoped semantics — clicking "Claude Code" launches whatever preset was last picked for the current worktree. That state is invisible. A user can't tell from looking at the button whether the next click will spawn the default Claude or a custom preset (which might be wired to a different model, dangerous mode, or different env vars). The currently-saved preset id is already computed at [`AgentButton.tsx#L111`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L111) — the data is there, it's just never surfaced.

Three related papercuts on the same surface:

1. **The tooltip omits the active preset.** Hovering the button shows `Start Claude Code (⌥⌘C)` regardless of which preset will fire. The user has to open the chevron menu and look for the marked item to find out what's about to launch.

2. **The split button uses one tooltip across two hit zones.** The icon side (launch) and the chevron side (open preset menu) are wrapped by a single `Tooltip` ([`AgentButton.tsx#L332-L471`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L332-L471)). Hovering the chevron tells you to "Start Claude Code" rather than describing what the chevron actually does.

3. **The saved-preset indicator is rendered three different ways depending on which menu you opened.** The chevron dropdown bolds the saved item with `font-medium` ([`AgentButton.tsx#L371`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L371)). The right-click context submenu appends a literal `" ✓"` text suffix ([`AgentButton.tsx#L515`](https://github.com/daintreehq/daintree/blob/ed35ad55632e11d7ceb97909d2a7b62ed1501b81/src/components/Layout/AgentButton.tsx#L515)). The agent tray's `SplitLaunchItem` doesn't mark the saved preset at all. Same data, three different treatments.

Desired end state: the active preset is visible without opening a menu (the tooltip is the obvious place — `Start Claude Code · Sonnet 4.5 (⌥⌘C)`), the chevron has its own tooltip describing its job, and the saved preset is marked the same way everywhere it appears — a single consistent indicator rather than bolding-or-suffix-or-nothing.

PR #5907, opened 2026-04-27
feat(toolbar): redesign agent button preset selection and launch menus

### LINKED ISSUE #5903: Redesign agent button preset selection and launch menus
The toolbar agent button has overlapping problems that are best solved together rather than as the two separate issues currently filed (#5901, #5902). They share the same surfaces, the same saved-preset resolution, and the same menu markup. Splitting them would cause churn and drift.

## Problem

**1. The active preset is invisible.** Clicking "Claude Code" launches whatever preset was last picked for the current worktree. The button doesn't say which one. Tooltip reads `Start Claude Code (⌥⌘C)` regardless of whether the next click will spawn the default Claude or a custom preset wired to a different model, dangerous mode, or env vars. The saved preset id is computed in `AgentButton.tsx` but never surfaced.

**2. The chevron that selects a preset reads as decoration, not an affordance.** It's `w-4` (16px) with a `h-3 w-3 opacity-70` icon. Most users miss it entirely and assume preset switching only lives in the right-click menu. The whole split group is wrapped in a single tooltip that says "Start Claude Code" no matter which half you hover, so even users who notice the chevron get told it's a launch button.

**3. The "saved preset" indicator is rendered three different ways.** The chevron dropdown bolds the saved item with `font-medium`. The right-click context submenu appends a literal `" ✓"` text suffix. The agent tray's `SplitLaunchItem` doesn't mark the saved preset at all. Same data, three treatments.

**4. The right-click menu has no way to manage presets.** The only management entry is `${name} Settings...` which dumps users into the agent's generic settings tab. People right-clicking an agent button mostly want to either launch it or tweak its presets. There should be a direct path to preset editing.

**5. "Launch in Worktree" nests three submenus deep.** Right-click → Launch in Worktree → `<worktree name>` → Grid / Dock. Three cascading flyouts means steering the mouse across two parent rows without overshooting, with the steering penalty compounding at each level.

## Desired end state

**The button labels its mode.** Tooltip names the active preset, e.g. `Start Claude Code · Sonnet 4.5 (⌥⌘C)`. Wording signals worktree scope, since the saved preset is per-worktree, not global. If the saved preset has been deleted or sanitised away, the tooltip falls back to "Default" rather than showing the raw preset id.

**The chevron is actually a button.** Width that reads as an affordance (around 24px), persistent when presets exist, with its own tooltip describing what it does. The chevron menu also gets a "Manage Presets..." footer item, since users opening the preset picker are exactly the users likely to want preset management.

**Saved preset indicator is unified.** A single trailing check-mark icon with muted text colour (no accent, since accent is reserved per CLAUDE.md), plus an `sr-only` "Saved preset" label. Same treatment in the chevron dropdown, the right-click submenu, and `SplitLaunchItem` in the agent tray.

**Right-click is the management surface.** Adds a "Manage Presets..." entry in both the with-presets and no-presets branches. Should deep-link to the preset editor for this specific agent. The existing `app.settings.openTab` action already supports `{ tab: "agents", subtab: type, sectionId: "agents-presets" }`, so no new Action needs to be defined.

**Worktree picker is one hover deep.** Inline trailing actions on each worktree row. Clicking the row launches Grid, with Dock as a compact trailing button on the same row. The inline pin pattern in `AgentTrayButton.tsx` is the in-codebase precedent. Alt-click for Dock can stay as an accelerator, but it shouldn't be the only path.

## Constraints

- **No accent colour on the saved-preset marker.** Per CLAUDE.md the accent is a scarce resource for one load-bearing signal per view. The marker is informational, not promotional.
- **Keyboard a11y.** The chevron must remain Tab-focusable. Reveal-on-hover is not acceptable, since it creates an invisible focusable Tab stop and kills first-time discovery.
- **`null` vs `undefined` in `agent.launch`.** They aren't interchangeable: `null` forces an explicit Default, `undefined` lets launcher resolution apply. Don't conflate them in the tooltip-resolution path or the chevron's "Default" item.
- **Tooltip leak.** If the primary and chevron tooltips are split, opening the dropdown must dismiss the primary tooltip, otherwise it'll hang over the menu.

## Evidence

The problem surfaces all live in one file plus an adjacent one:

- [`src/components/Layout/AgentButton.tsx`](https://github.com/daintreehq/daintree/blob/3632617fdb6ac1ed851ae8e2d20de91b662aa4b3/src/components/Layout/AgentButton.tsx) — split-button render and both branches of the right-click menu. The 3-deep worktree picker is duplicated across the with-presets and no-presets branches; worth refactoring into one component or it'll drift.
- [`src/components/Layout/AgentTrayButton.tsx`](https://github.com/daintreehq/daintree/blob/3632617fdb6ac1ed851ae8e2d20de91b662aa4b3/src/components/Layout/AgentTrayButton.tsx) — adjacent surface with the inline pin precedent and the `SplitLaunchItem` that needs the unified indicator. `AgentRow` will need `savedPresetId` plumbed through to render the marker.

Consolidates #5901 and #5902.

gregpriday:
## Summary

- Fixes 5 interconnected UX problems with the toolbar agent button (active preset visibility, chevron affordance, inconsistent saved-preset indicators, no preset management entry point, and a 3-deep worktree picker) from #5903, which consolidates #5901 and #5902
- Critical correctness fix: the primary button click no longer forwards an explicit `presetId` to `agent.launch` — it defers to `useAgentLauncher`'s stale-fallback resolution path, which handles deleted/sanitised presets gracefully
- 21 unit tests in `AgentButton.test.tsx` (14 existing kept, 6 new scenarios, 1 stale-preset regression guard)

Resolves #5903

## Changes

- **Tooltip names the active preset.** Shows `Start Claude Code · <preset name> (⌥⌘C)` scoped to the current worktree. Falls back to "Default" when the saved preset has been deleted or sanitised away
- **Chevron is a proper affordance.** Wider hit target (~24px), persistent when presets exist, with its own tooltip. Dropdown includes a "Manage Presets..." footer item that deep-links to the preset editor via the existing `app.settings.openTab` action
- **Unified Check indicator.** Single trailing check icon with muted text colour (no accent per CLAUDE.md) across the chevron dropdown, right-click submenu, and `SplitLaunchItem` in `AgentTrayButton` — replaces the three inconsistent treatments (bold text, `" ✓"` suffix, nothing)
- **"Manage Presets..." in right-click menu.** Added to both the with-presets and no-presets branches of the context menu
- **Worktree picker is 1 hover deep.** Inline trailing Dock button on each worktree row replaces the 3-level flyout. Grid is the row click; Dock is the trailing button on hover
- **No `presetId` forwarded on primary click.** `resolvedPresetId` was previously passed into `agent.launch` which bypassed the launcher's stale-preset handling. Primary click now passes no `presetId`, letting the launcher resolve it via its own fallback logic
- Includes a small pre-existing prettier fix to `Toolbar.tsx` (format drift on develop that would have blocked CI)

## Testing

- 21 unit tests covering: tooltip label resolution (active/default/deleted preset), chevron presence/absence, Check indicator rendering across surfaces, Manage Presets entry, stale-preset regression guard, worktree picker structure
- `npm run check` passes (0 errors, warnings are pre-existing on develop)

### DIFF
--- a/src/components/Layout/AgentButton.tsx
+++ b/src/components/Layout/AgentButton.tsx
@@ -1,10 +1,11 @@
-import { useMemo, useRef, type PointerEvent as ReactPointerEvent } from "react";
+import { useMemo, useRef, useState, type PointerEvent as ReactPointerEvent } from "react";
 import { Button } from "@/components/ui/button";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { ShortcutRevealChip } from "@/components/ui/ShortcutRevealChip";
 import {
   DropdownMenu,
   DropdownMenuContent,
+  DropdownMenuItem,
   DropdownMenuLabel,
   DropdownMenuRadioGroup,
   DropdownMenuRadioItem,
@@ -151,6 +152,35 @@ export function AgentButton({
   const panelsById = usePanelStore(useShallow((s) => s.panelsById));
   const panelIds = usePanelStore(useShallow((s) => s.panelIds));
   const activeWorktreeId = useWorktreeSelectionStore((s) => s.activeWorktreeId);
+
+  // Radix Tooltip reopens on focus restoration. When the chevron's
+  // DropdownMenu closes, Radix returns focus to the chevron trigger and the
+  // tooltip would reopen on top of the freshly-launched action's surfaces.
+  // Gate both halves' tooltips on controlled state and hold suppression open
+  // until the next genuine pointer hover. Same pattern as AgentTrayButton.
+  const [primaryTooltipOpen, setPrimaryTooltipOpen] = useState(false);
+  const [chevronTooltipOpen, setChevronTooltipOpen] = useState(false);
+  const isRestoringFocusRef = useRef(false);
+
+  const handlePrimaryTooltipOpenChange = (open: boolean) => {
+    if (open && isRestoringFocusRef.current) return;
+    setPrimaryTooltipOpen(open);
+  };
+
+  const handleChevronTooltipOpenChange = (open: boolean) => {
+    if (open && isRestoringFocusRef.current) return;
+    setChevronTooltipOpen(open);
+  };
+
+  const suppressTooltipsDuringFocusRestore = () => {
+    setPrimaryTooltipOpen(false);
+    setChevronTooltipOpen(false);
+    isRestoringFocusRef.current = true;
+  };
+
+  const clearFocusRestoreSuppression = () => {
+    isRestoringFocusRef.current = false;
+  };
 
   const activeSession = useMemo(() => {
     const states: (AgentState | undefined)[] = [];
@@ -301,7 +331,7 @@ export function AgentButton({
     return (
       <ContextMenu>
         <ContextMenuTrigger asChild>
-          <Tooltip>
+          <Tooltip open={primaryTooltipOpen} onOpenChange={handlePrimaryTooltipOpenChange}>
             <TooltipTrigger asChild>
               <span className="inline-flex">
                 <Button
@@ -310,6 +340,7 @@ export function AgentButton({
                   onClick={handleClick}
                   disabled={isLoading}
                   data-toolbar-item={dataToolbarItem}
+                  onPointerEnter={clearFocusRestoreSuppression}
                   className={cn(
                     "toolbar-agent-button text-daintree-text transition-colors relative",
                     isReady &&
@@ -326,7 +357,7 @@ export function AgentButton({
             <TooltipContent side="bottom">{tooltip}</TooltipContent>
           </Tooltip>
         </ContextMenuTrigger>
-        <ContextMenuContent>
+        <ContextMenuContent className="max-h-[var(--radix-context-menu-content-available-height)] overflow-y-auto">
           <ContextMenuItem
             disabled={!isReady}
             onSelect={() =>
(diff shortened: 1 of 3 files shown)


gregpriday:
Heads up — autonomous merge pipeline ran the recent batch. Verdict was READY on review, but rebase against the latest develop hits a structural conflict on `AgentButton.tsx`: this PR re-introduces per-component `<TooltipProvider>` wrappers in the same regions where #5909 just removed them in favor of a single root provider. The two intents can't be merged mechanically — needs a manual call on whether the controlled tooltip suppression state from this PR should be preserved without the redundant providers.

Same conflict shape blocks #5905 and #5908. Suggested resolution path: rebase this PR by hand, dropping the per-component `TooltipProvider` wrappers and keeping the controlled suppression refs/state, then close #5905 and #5908 as superseded.

Deferred — no auto-retry.

gregpriday:
Update — the constituent PRs both landed via the autonomous pipeline after rebase agents handled the conflicts:

- #5908 merged (right-click context menu, Manage Presets entry, flattened worktree picker)
- #5905 merged (active preset surfaced in tooltip, separate launch/chevron tooltips, unified saved-preset indicator)

This PR's intent is now in develop. Likely safe to close as superseded — though worth a quick diff against develop to confirm nothing in this PR was lost during the per-PR rebases (the rebase agents did keep the bare `<Tooltip>` architecture from #5909 and dropped the per-component `<TooltipProvider>` wrappers in the process).

PR #6072, opened 2026-04-28
feat(hooks): add useLoadingState for deferred spinner and slow-call signals

### LINKED ISSUE #6030: Add deferred-loading and slow-call primitives
Loading states across panels suffer from two related issues that compound across the app.

First, spinners and skeletons render immediately on every load, even when the underlying call resolves in well under 100ms. The brief flash reads as a glitch rather than progress — design-system guidance treats sub-200ms loaders as net-negative. Several panels (sidebar refreshes, settings tabs, GitHub dropdowns) flicker visibly on routine state changes.

Second, when an async call exceeds typical latency, there's no acknowledgment that anything is wrong. The spinner spins indefinitely, the user has no signal that the operation is slow rather than stuck, and there's no way to cancel.

Both gaps are downstream of missing primitives:

1. **Deferred-loading helper** — render the loading indicator only after a configurable delay (default ~200ms). If the call resolves before the delay elapses, no indicator ever appears.
2. **Slow-call signal** — track an `isSlow` flag separately from `isPending` so panels can layer a "Taking longer than usual" message and a Cancel affordance over the existing loader after a configurable threshold (typical bands: 3s notice, 10s error).

These are small utilities, but they unblock the visible-polish work in BrowserPane, DevPreviewPane, agent CLI probes, and any future panel that wraps a slow IPC call. The existing `Spinner` already respects `prefers-reduced-motion`; the new behavior should preserve that.

gregpriday:
## Summary

- Adds `useLoadingState(isPending, deferDelay?, slowThreshold?, overdueThreshold?)` — a flat-params hook that returns `{ showSpinner, isSlow, isOverdue }`. The spinner is gated behind a defer window (default 200ms) so fast loads don't flash a spinner at all.
- Wires it into `BrowserPane` and `DevPreviewPane`, replacing the ad-hoc spinner logic in both with a consistent primitive. DevPreviewPane also suppresses the "Starting / Installing / Restarting" label for the same 200ms window — sub-200ms transitions don't need a label.
- `isSlow` and `isOverdue` are exposed on the return value for future consumers; the "Taking longer than usual" copy and cancel UI are out of scope here since they need per-component cancellation chains that don't exist yet.

Resolves #6030

## Changes

- `src/hooks/useLoadingState.ts` — three independent `setTimeout` calls in one `useEffect`; flags reset at the top of the effect so threshold changes mid-cycle don't leave stale state
- `src/hooks/__tests__/useLoadingState.test.tsx` — 15 tests covering deferred onset, threshold boundaries, mid-cycle threshold changes, prior-cycle flag leakage, and unmount cleanup
- `src/hooks/index.ts` — barrel export added
- `src/components/Browser/BrowserPane.tsx` — spinner overlay gated on `showSpinner`
- `src/components/DevPreview/DevPreviewPane.tsx` — dev-server spinner and status label gated on `showDevServerSpinner`

**Naming note:** `isOverdue` was chosen over the issue's `isError` to avoid collision with the existing `isError` / `status === "error"` semantics already in both components.

## Testing

- 15/15 unit tests pass with fake timers covering all boundary conditions
- `npm run check` passes: typecheck clean, lint ratchet at -2 warnings, format unchanged

### DIFF
--- a/src/components/Browser/BrowserPane.tsx
+++ b/src/components/Browser/BrowserPane.tsx
@@ -29,6 +29,7 @@ import { useProjectSettingsStore } from "@/store/projectSettingsStore";
 import { useProjectSettings } from "@/hooks/useProjectSettings";
 import { useUrlHistoryStore } from "@/store/urlHistoryStore";
 import { useFindInPage } from "@/hooks/useFindInPage";
+import { useLoadingState } from "@/hooks/useLoadingState";
 import { logError } from "@/utils/logger";
 
 export interface BrowserPaneProps extends BasePanelProps {
@@ -131,6 +132,7 @@ export function BrowserPane({
   });
 
   const [isLoading, setIsLoading] = useState(true);
+  const { showSpinner } = useLoadingState(isLoading);
   const [loadError, setLoadError] = useState<string | null>(null);
   const [blockedNav, setBlockedNav] = useState<{
     url: string;
@@ -1001,7 +1003,7 @@ export function BrowserPane({
             )}
             <div className="relative flex-1 min-h-0">
               {isDragging && <div className="absolute inset-0 z-10 bg-transparent" />}
-              {isLoading && (
+              {showSpinner && (
                 <div className="absolute inset-0 flex items-center justify-center bg-daintree-bg z-10">
                   <Spinner size="2xl" className="text-status-info" />
                 </div>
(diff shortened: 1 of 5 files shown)


gregpriday:
Holding this one for a follow-up. Merged #6107 (`fix(browser): improve error handling and slow-load UX for Browser and DevPreview`) shipped a competing slow-load story for the same two surfaces — overlay-based, with timeout/retry/cancel UX baked in.

The `useLoadingState` hook itself is cleanly built (15 tests, sound timer hygiene, three independent thresholds), and there's a real use for it as a standalone primitive. But the BrowserPane / DevPreviewPane integration in this PR now duplicates state management with #6107's overlay machinery.

Two ways to land this:
1. Drop the BrowserPane/DevPreviewPane changes from this PR. Keep just the hook + its tests + the barrel export. That gives us the primitive without the integration conflict.
2. Refactor #6107's overlay to consume `useLoadingState` as the source of truth for `showSpinner`/`isSlow`/`isOverdue`. Bigger change but unifies the two systems.

Either path needs a fresh diff from you. Deferring until then.

PR #6103, opened 2026-04-28
feat(ipc): type-brand handler returns to forbid {ok|success} envelope keys

### LINKED ISSUE #6020: Type-brand the IPC envelope to forbid handler returns of {ok|success: false}
The global `ipcMain.handle` wrapper in `electron/setup/security.ts` already catches throws and returns `{__daintreeIpcEnvelope: true, ok: true|false, data|error}`. The bug: ~14 handler files *also* return their own `{ok: false, error}` or `{success: false, error}` shapes from handler logic — these get wrapped INSIDE the success envelope, producing `{__envelope: true, ok: true, data: {ok: false, error: "…"}}`. The renderer's `_unwrappingInvoke` sees `ok: true` and returns `data` without checking the inner discriminator. Failures are silently swallowed.

49 such return statements across these files: `clipboard.ts`, `git-write.ts`, `webview.ts`, `logs.ts`, `appTheme.ts`, `gemini.ts`, `ai.ts`, `files.ts`, `terminal/io.ts`, `keybinding.ts`, `terminalConfig.ts`, `projectCrud/crud.ts`, `projectCrud/gitClone.ts`, `projectCrud/gitInit.ts`. Field naming is also inconsistent (`ok` vs `success`).

The fix is type-level: constrain handler return types so they cannot include keys that overlap with the envelope discriminator. A handler that tries to return `{ok: false, …}` should fail typecheck with a message pointing the developer at `throw new AppError(...)` instead.

Permalink to the wrapper that auto-envelopes everything: https://github.com/daintreehq/daintree/blob/c8b745df552059de6f93e47f0deb4efa0ac66110/electron/setup/security.ts#L21-L59

Permalink to the renderer-side unwrapper that misses inner failures: https://github.com/daintreehq/daintree/blob/c8b745df552059de6f93e47f0deb4efa0ac66110/electron/preload.cts#L491-L498

Permalink to one example of an inner-envelope return that the type-brand should reject: https://github.com/daintreehq/daintree/blob/c8b745df552059de6f93e47f0deb4efa0ac66110/electron/ipc/handlers/clipboard.ts#L55-L84

This is purely a type-system change — no runtime behavior changes. The compiler error list it produces is the migration backlog for the follow-up issue that introduces `AppError` and rewrites those handlers.

gregpriday:
## Summary

- Adds a compile-time constraint that bans handler return types containing `ok` or `success` keys — the inner-envelope antipattern that leaks discriminator shape into the transport layer. Handlers that want to signal failure should throw an `AppError`, not return a failure-shaped value.
- Pure type-level change, no runtime behaviour touched. CI stays green via `@ts-expect-error` markers at 19 known violation sites. Each marker is tagged `See #6020` so the follow-up migration can enumerate them with a single grep.
- The `ForbidIpcEnvelopeKeys<T>` constraint is distributive over unions, so `{success: boolean} | null` can't slip through either (that edge case was caught in a Codex review pass and fixed before this landed).

Resolves #6020

## Changes

- `shared/types/ipc/errors.ts` — `ForbidIpcEnvelopeKeys<T>` and `IpcHandlerEnvelopeViolation` (branded error type whose property name carries the remediation hint inline in tsc output)
- `electron/ipc/define.ts`, `electron/ipc/utils.ts` — `PlainHandler`, `ContextHandler`, `typedHandle`, `typedHandleWithContext` return positions wired to `SafeResult<K>`
- 19 handler files — `@ts-expect-error // See #6020` at each violation site (the migration backlog)
- `shared/types/__tests__/ipcEnvelopeConstraint.test.ts` — 8 type-level vitest cases covering primitives, safe objects, banned keys, full discriminated unions, and the `null/undefined` union edge case
- `electron/ipc/__tests__/define.test.ts` — switched test channels to envelope-safe ones so the test suite doesn't rely on violation sites

## Testing

`npm run check` clean (typecheck, channels, lint:ratchet, format). `npm test` passes — 14 type-level assertions + define.test.ts.

### DIFF
--- a/electron/ipc/define.ts
+++ b/electron/ipc/define.ts
@@ -8,20 +8,27 @@ import {
 import { ValidationError } from "./validationError.js";
 import type { IpcContext } from "./types.js";
 import type { IpcInvokeMap } from "../types/index.js";
+import type { ForbidIpcEnvelopeKeys } from "../../shared/types/ipc/errors.js";
 
 export { ValidationError };
 
 type Channel = keyof IpcInvokeMap;
 type Cleanup = () => void;
 
+// Reject handler return types that overlap with the envelope discriminator.
+// security.ts auto-wraps every handler return in {ok: true, data}; a handler
+// returning {ok: false, ...} would be silently nested inside a success
+// envelope. This constraint forces such handlers to throw instead.
+type SafeResult<K extends Channel> = ForbidIpcEnvelopeKeys<IpcInvokeMap[K]["result"]>;
+
 type PlainHandler<K extends Channel> = (
   ...args: IpcInvokeMap[K]["args"]
-) => Promise<IpcInvokeMap[K]["result"]> | IpcInvokeMap[K]["result"];
+) => Promise<SafeResult<K>> | SafeResult<K>;
 
 type ContextHandler<K extends Channel> = (
   ctx: IpcContext,
   ...args: IpcInvokeMap[K]["args"]
-) => Promise<IpcInvokeMap[K]["result"]> | IpcInvokeMap[K]["result"];
+) => Promise<SafeResult<K>> | SafeResult<K>;
 
 type ValidatedPayloadHandler<K extends Channel, S extends z.ZodTypeAny> = (
   payload: z.output<S>
--- a/electron/ipc/handlers/agentCli.ts
+++ b/electron/ipc/handlers/agentCli.ts
@@ -185,7 +185,10 @@ export function registerAgentCliHandlers(deps: HandlerDependencies): () => void 
       }
     });
   };
-  handlers.push(typedHandleWithContext(CHANNELS.SETUP_AGENT_INSTALL, handleSetupAgentInstall));
+  handlers.push(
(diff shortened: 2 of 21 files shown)


gregpriday:
Deferring this one to a follow-up rebase pass. It overlaps structurally with #6097 (IPC validation wrapper): both rewrite `PlainHandler`/`ContextHandler` return types in `electron/ipc/define.ts`, and `typedHandle`/`typedHandleWithContext` in `electron/ipc/utils.ts`. After #6097 lands, the `ForbidIpcEnvelopeKeys` brand also needs to be extended to the new `typedHandleValidated`/`typedHandleWithContextValidated` signatures — otherwise every namespace migrated to `opValidated()` (slashCommands, systemShell, files, terminal/lifecycle, terminal/snapshots) bypasses the brand check and can quietly return `{ok: ...}` envelopes again.

Also: PR body says "19 known violation sites" tagged with `@ts-expect-error`, but actual count is 30. Update the body when rebasing.

Plan: land #6097, rebase this PR, extend the constraint to the validated-handler types in the same pass, then re-run the merge pipeline.

PR #6312, opened 2026-05-01
refactor(panels): tighten types in panel kind registry

### LINKED ISSUE #6293: Tighten types in panel kind registry (drop any casts, satisfies, variance cast comment)
The panel kind registry in `src/panels/registry.tsx` carries three small type-safety drags that React 19 + TypeScript 5.5 make removable.

**1. `(props: any)` casts in lazy wrappers.** `BrowserPaneWrapper` and `DevPreviewPaneWrapper` accept `props: any` because the older `lazy()` typings didn't propagate prop types through the `LazyExoticComponent` boundary. React 19's `lazy<T>()` returns `LazyExoticComponent<T> = ExoticComponent<ComponentPropsWithRef<T>>`, so `React.ComponentProps<typeof LazyBrowserPane>` now extracts the underlying prop type cleanly. The wrappers can be typed against the actual lazy-imported component instead of `any`.

See [`src/panels/registry.tsx`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/panels/registry.tsx#L45-L72).

**2. `PANEL_KIND_DEFINITION_REGISTRY` widens via `Record<string, …>`.** The registry is typed as `Record<string, PanelKindDefinition>`, which means indexing with a known kind discards the narrow component-prop type. TS 5.5's constant-indexed-access narrowing combined with `satisfies` lets us keep per-entry types narrow while still validating the shape of every entry. Pattern:

```ts
const PANEL_KIND_DEFINITION_REGISTRY = {
  terminal: { ...requirePanelKindConfig(\"terminal\"), component: TerminalPane },
  browser: { ...requirePanelKindConfig(\"browser\"), component: BrowserPaneWrapper },
  \"dev-preview\": { ...requirePanelKindConfig(\"dev-preview\"), component: DevPreviewPaneWrapper },
} satisfies Record<string, PanelKindDefinition>;
```

See [`src/panels/registry.tsx`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/panels/registry.tsx#L128-L132).

**3. The variance cast comment understates the trap.** `initBuiltInPanelKinds` casts each per-kind serializer/createDefaults pair to the wide `PanelKindConfig['serialize']` shape. The current comment correctly explains it's contravariance, but the obvious-looking \"fix\" — switching to method syntax to get bivariant parameters — is *unsafe widening* per the TS handbook + `microsoft/TypeScript#40192`. It would let a caller hand a wider `PanelInstance` into a strictly typed serializer.

The cast is correct; what's missing is a sentence calling out method syntax as the unsafe trap so the next reader doesn't try it.

See [`src/panels/registry.tsx`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/panels/registry.tsx#L104-L118).

**Constraints**
- Don't change the runtime shape of the registry — extension consumers depend on `getPanelKindDefinition(kind)` and `registerPanelKindDefinition`.
- The `as PanelKindConfig['serialize']` cast must remain. The change is documentation + ergonomic typing, not a structural rewrite.

gregpriday:
## Summary

- Replaced `props: any` in `BrowserPaneWrapper` and `DevPreviewPaneWrapper` with `ComponentProps<typeof LazyBrowserPane>` / `ComponentProps<typeof LazyDevPreviewPane>` — drops both `eslint-disable` suppressions alongside
- Expanded the variance cast comment in `initBuiltInPanelKinds` to explicitly call out method syntax as an unsafe bivariance trap, not just a style choice
- Added a comment on `PANEL_KIND_DEFINITION_REGISTRY` explaining why the `Record<string, PanelKindDefinition>` annotation has to stay: `registerPanelKindDefinition` mutates the registry by dynamic string key at runtime, and `satisfies` strips the index signature TypeScript needs to accept that

Resolves #6293

## Changes

**`src/panels/registry.tsx`** only — no runtime behaviour changes, no shape changes to the registry.

The one item from the issue that wasn't adopted: switching the registry to `satisfies Record<string, PanelKindDefinition>`. That strips the index signature, which breaks the `registerPanelKindDefinition` mutation site with TS7017. The comment in the file documents the trade-off so the next reader doesn't have to rediscover it.

## Testing

- Typecheck clean
- Lint warning count down by 2 (the two suppressed `any` casts removed)
- 32 panel tests passing

### DIFF
--- a/src/panels/registry.tsx
+++ b/src/panels/registry.tsx
@@ -1,4 +1,4 @@
-import { Suspense, lazy, type ComponentType } from "react";
+import { Suspense, lazy, type ComponentProps, type ComponentType } from "react";
 import type { PanelKindConfig } from "@shared/config/panelKindRegistry";
 import { getPanelKindConfig } from "@shared/config/panelKindRegistry";
 import type { PtyPanelData, BrowserPanelData, DevPreviewPanelData } from "@shared/types/panel";
@@ -53,8 +53,7 @@ const LazyDevPreviewPane = lazy(() =>
 // correct componentName attribution on chunk-load failures. The per-panel
 // boundary in GridPanel catches render errors; this wrapper catches import
 // failures with proper attribution — the two boundaries serve different roles.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function BrowserPaneWrapper(props: any) {
+function BrowserPaneWrapper(props: ComponentProps<typeof LazyBrowserPane>) {
   return (
     <ErrorBoundary variant="component" componentName="BrowserPane">
       <Suspense fallback={<BrowserPaneSkeleton />}>
@@ -64,8 +63,7 @@ function BrowserPaneWrapper(props: any) {
   );
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function DevPreviewPaneWrapper(props: any) {
+function DevPreviewPaneWrapper(props: ComponentProps<typeof LazyDevPreviewPane>) {
   return (
     <ErrorBoundary variant="component" componentName="DevPreviewPane">
       <Suspense fallback={<BrowserPaneSkeleton label="Loading dev preview panel" />}>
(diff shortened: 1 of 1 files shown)


gregpriday:
**CI status note** — `check` and `test` are failing on this branch, but the failures are inherited from `develop` (currently also red on its own CI run for `e6d768494`). The 6 ESLint `no-console` errors are all in files this PR does not touch:

- `src/components/Browser/BrowserPane.tsx:193,199,201`
- `src/store/slices/panelRegistry/core.ts:689,908`
- `src/store/slices/panelRegistry/trashActions.ts:113`

They were added as intentional temporary debug `console.error` calls in `e6d768494` ("fix(e2e): harden fault-mode tests and fix dialog/browser regressions"). This PR's diff is a single file (`src/panels/registry.tsx`, +11/-5) and locally passes typecheck, lint (baseline -2), format, and the 32 panel-registry tests.

CI will go green once the develop-side lint cleanup lands.

PR #6324, opened 2026-05-01
feat(panel-kinds): emit registry events on plugin register/unregister

### LINKED ISSUE #6306: Emit registry events when panel kinds register or unregister at runtime
`registerPanelKind` and `unregisterPluginPanelKinds` mutate the `PANEL_KIND_REGISTRY` map silently. When a plugin contributing a panel kind loads or reloads mid-session, nothing in the renderer learns about it.

This matters because we already do most of the work to handle the missing-kind case gracefully:

- Saved snapshots referencing an unregistered kind survive hydration. `src/utils/stateHydration/index.ts` recognizes pluginId-bearing or dotted-kind panels and creates them anyway.
- A `PluginMissingPanel` placeholder component exists and is wired through `ContentPanel` — it shows a clean \"Plugin unavailable — re-enable to restore\" UI.

The gap: **once a plugin re-registers its kind, live placeholder slots don't swap to the real component.** The user has to reload the window. VS Code, Theia, and Obsidian all expose registry-change events specifically so placeholders can hot-swap.

Add `onDidRegisterPanelKind` / `onDidUnregisterPanelKind` event emitters on the panel kind registry and have the renderer's panel resolution path subscribe — when a previously missing kind appears, the placeholder slot rerenders against the now-resolvable definition. Same path covers plugin uninstall/reload during a session.

**Why it matters**

Plugin authors iterating on a panel kind have to fully reload the window to see their changes propagate to existing panel slots. Users whose plugin temporarily failed to load see a permanent placeholder until they reload, even after the plugin recovers. The placeholder UI was built precisely for this scenario; the missing piece is the swap-on-recovery signal.

**Related**
- [`shared/config/panelKindRegistry.ts`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/shared/config/panelKindRegistry.ts#L108-L132) — current silent-mutation register/unregister
- [`src/components/Panel/PluginMissingPanel.tsx`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/components/Panel/PluginMissingPanel.tsx) — placeholder that should hot-swap
- [`src/utils/stateHydration/index.ts`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/utils/stateHydration/index.ts#L680-L703) — hydration path that already preserves unresolvable kinds

**Constraints**
- The registry currently lives in the shared layer (`shared/config/`) so it can run in main and renderer. The event emitter has to be friendly to both — Node `EventEmitter` is fine, or a tiny in-house listener set.
- Built-in panel kinds get registered once at startup; the events are for plugin-driven mutations only. Don't emit for the bootstrap registration.
- A swap-on-register path needs to handle the case where the placeholder's `extensionState` should be threaded into the newly resolved panel — the placeholder already renders against `pluginId` + `kind` from the existing `PanelInstance`, so the swap is just a rerender, not a state migration.

gregpriday:
## Summary

- The shared panel kind registry mutated silently when plugins registered or unregistered kinds at runtime -- the renderer had no way to know something changed, so placeholder slots stayed stuck until a full window reload.
- Adds `onDidRegisterPanelKind` and `onDidUnregisterPanelKind` event listeners to the shared registry, a `usePluginPanelKinds` hook in the renderer that subscribes and maintains a live map, and a `PluginService` in main with IPC plumbing to broadcast registration changes across the process boundary.
- `DockedPanel` and `GridPanel` now resolve plugin kinds from the live map, so placeholder slots hot-swap to the real component the moment a plugin re-registers.

Resolves #6306

## Changes

- `shared/config/panelKindRegistry.ts` -- `onDidRegisterPanelKind` / `onDidUnregisterPanelKind` event emitters; stale definition cleared on `hasPty` flip; events fire only for plugin-driven mutations, not bootstrap registrations
- `src/hooks/usePluginPanelKinds.ts` -- new hook; subscribes to registry events via IPC and keeps a live `Map` of plugin panel kind definitions
- `src/components/Terminal/DockedPanel.tsx` + `GridPanel.tsx` -- pull plugin kinds from `usePluginPanelKinds` live map instead of static snapshot
- `electron/services/PluginService.ts` -- new service; listens to registry events and broadcasts over IPC; cancels disposed broadcasts
- `electron/ipc/` -- new channel + handler wiring for plugin kind events
- `src/panels/registry.tsx` -- integrates `usePluginPanelKinds` into the panel kind resolution path

## Testing

- Unit tests for `panelKindRegistry` event emitters (register, unregister, hasPty flip, no-emit on built-ins)
- Unit tests for `usePluginPanelKinds` hook (subscribe, hot-swap on re-register, cleanup)
- Unit tests for `PluginService` IPC handler and broadcast cancellation

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -611,6 +611,7 @@ export const CHANNELS = {
   PLUGIN_ACTIONS_GET: "plugin:actions-get",
   PLUGIN_ACTIONS_REGISTER: "plugin:actions-register",
   PLUGIN_ACTIONS_UNREGISTER: "plugin:actions-unregister",
+  PLUGIN_PANEL_KINDS_GET: "plugin:panel-kinds-get",
 
   // Config reload channels
   APP_RELOAD_CONFIG: "app:reload-config",
--- a/electron/ipc/handlers/__tests__/plugin.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/plugin.handlers.test.ts
@@ -57,7 +57,7 @@ beforeEach(() => {
 describe("registerPluginHandlers", () => {
   it("registers handlers for all plugin channels", () => {
     registerPluginHandlers();
-    expect(mockIpcMainHandle).toHaveBeenCalledTimes(8);
+    expect(mockIpcMainHandle).toHaveBeenCalledTimes(9);
     expect(mockIpcMainHandle).toHaveBeenCalledWith("plugin:list", expect.any(Function));
     expect(mockIpcMainHandle).toHaveBeenCalledWith("plugin:invoke", expect.any(Function));
     expect(mockIpcMainHandle).toHaveBeenCalledWith("plugin:toolbar-buttons", expect.any(Function));
@@ -72,6 +72,7 @@ describe("registerPluginHandlers", () => {
       "plugin:actions-unregister",
       expect.any(Function)
     );
+    expect(mockIpcMainHandle).toHaveBeenCalledWith("plugin:panel-kinds-get", expect.any(Function));
   });
 
   it("throws before registering any handler when invoked before enforceIpcSenderValidation", () => {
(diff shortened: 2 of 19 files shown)


gregpriday:
Heads-up: the failing `check` and `test` jobs are inherited from develop, not introduced by this PR. The latest develop commit (e6d768494, "fix(e2e): harden fault-mode tests…") shipped temporary debug `console.error` calls in `BrowserPane.tsx`, `panelRegistry/core.ts`, and `trashActions.ts` (lint failures), plus a `webview.loadURL().catch()` call that throws against the test webview mock in `BrowserPane`/`DevPreviewPane`.

Develop's own CI run on that SHA fails identically (https://github.com/daintreehq/daintree/actions/runs/25202088964). None of the failing files are in this branch's diff — the panel-kind registry tests, plugin handler tests, and PluginService tests added/touched here all pass locally and inside CI. CI will go green for #6324 automatically once develop is fixed and this branch is rebased.

Reviewers can verify by checking the diff: `git diff develop...HEAD --stat` shows only registry, hook, IPC plumbing, and test files.

PR #6326, opened 2026-05-01
fix(icons): forward className to bare child in BrandMark

### LINKED ISSUE #6290: Dock launch menu: agent icons have no gap before name
In the dock's "+" launch menu, agent rows under "Launch agent" render with the brand icon flush against the agent name — no visible gap. The "Launch panel" rows (Terminal, Browser) and "Launch recipe" rows render with the expected spacing.

All rows in [`DockLaunchMenuItems.tsx`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/components/Layout/DockLaunchMenuItems.tsx#L54-L107) use the same `w-3.5 h-3.5 mr-2` icon classes, but only the brand icons (e.g. [`ClaudeIcon`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/components/icons/brands/ClaudeIcon.tsx), [`CodexIcon`](https://github.com/daintreehq/daintree/blob/540019a071ec6989302a18e2601fe314942fc937/src/components/icons/brands/CodexIcon.tsx)) collapse the gap. The lucide icons and the custom `Workflow` icon on the same menu render correctly.

Agent rows should have the same icon-to-label spacing as the rest of the menu.

gregpriday:
## Summary

- `BrandMark.tsx` was discarding `className` on bare SVG children — `cloneElement` spread props onto a wrapper `div` but never passed them through to the child element, so spacing classes like `mr-2` had no effect
- This caused brand icons in the dock launch menu to render flush against the agent name, with no visible gap
- Fix merges the incoming `className` into the child element directly

Resolves #6290

## Changes

- `src/components/icons/BrandMark.tsx` — merge `className` prop into the bare child via `cloneElement` instead of spreading onto the wrapper
- `src/components/icons/__tests__/BrandMark.test.tsx` — 5 new tests covering `className` forwarding and merging behaviour

## Testing

Unit tests cover: className forwarded when absent on child, merged correctly when child already has its own className, no-op when BrandMark receives no className, and wrapper-children path (unaffected by the change).

### DIFF
--- a/src/components/icons/BrandMark.tsx
+++ b/src/components/icons/BrandMark.tsx
@@ -1,4 +1,4 @@
-import type { ReactElement } from "react";
+import { cloneElement, type ReactElement } from "react";
 import { cn } from "@/lib/utils";
 import { resolveBrandChip } from "@/lib/brandIcon";
 import { useActiveAppScheme } from "@/hooks/useActiveAppScheme";
@@ -7,7 +7,7 @@ interface BrandMarkProps {
   brandColor?: string;
   size?: number;
   className?: string;
-  children: ReactElement;
+  children: ReactElement<{ className?: string }>;
 }
 
 const SIZE_CLASS_REGEX = /\b(?:size-|w-|h-)/;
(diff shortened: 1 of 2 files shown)


gregpriday:
Heads up: the `test` job is failing on a pre-existing `McpServerService` assertion (`sets openWorldHint true for network-bound categories and false for local ones`) — the same test also fails on develop at b476a1636 after the MCP allowlist landed in 5234f55fa. The tool IDs the test uses (`github.listPRs`, `browser.navigate`) aren't in the new `MCP_TOOL_ALLOWLIST`, so they get filtered out and `ghTool` resolves to `undefined`. Unrelated to this PR's icon change — should be fixed in a separate MCP-scoped PR.

PR #6553, opened 2026-05-02
feat(mcp): enforce four-tier tool authorization at dispatch

### LINKED ISSUE #6517: Enforce MCP tool authorization at dispatch time, not listTools
The local MCP server's `MCP_TOOL_ALLOWLIST` filters which tools are advertised to clients via `listTools`, but the dispatch path doesn't enforce it — any client that already knows an action ID can invoke it. The code comment at [McpServerService.ts:36-41](https://github.com/daintreehq/daintree/blob/64e87d9601ed5b632966dcfcb6248c34cc2bf344/electron/services/McpServerService.ts#L36-L41) is explicit about this. That makes the allowlist descriptive, not enforcing.

This is fine when there's exactly one trust tier ("the user runs this tool, the user is the user"). It breaks the moment we want different tiers of access for different connections — which we want, both for the help assistant and for project worktree agents that should have a narrower surface than power-user external clients.

Tier model to ship:

- **`workbench`** — read-only Daintree introspection (project state, worktree list, file search, terminal output, copy-tree). Default for project agents that opt in.
- **`action`** — adds non-destructive mutations (create worktree, run recipe, inject terminal text, panel manipulation). Default for the help assistant.
- **`system`** — adds destructive operations (delete worktree, send raw terminal commands, git mutations, `agent.launch`). Only when "skip permissions" is explicitly enabled, or when an external client has explicitly opted in.
- **`external`** — keeps today's curated allowlist behavior for backward compatibility with the existing user-facing external server.

Each connection authenticates with a token whose tier is recorded server-side (in-memory map). Dispatch consults that map before forwarding the action — reject with a clear MCP error envelope if the tier doesn't grant the requested action ID.

Important: the existing single-key external server doesn't disappear. This adds a layer; it doesn't replace `apiKey`. The external bearer maps to tier `external` for now.

Walk the action manifest to draw the tier boundaries — there are ~258 actions; the right cuts depend on category and `danger`. Don't just inherit `MCP_TOOL_ALLOWLIST` as-is for every tier — it's already too permissive for `workbench`.

Done when:
- Dispatch rejects unknown-tier actions with an MCP error, before bouncing IPC to the renderer.
- Each tier's allowlist is defined and unit-tested.
- The existing single-key external server continues working under the `external` tier.
- The token-to-tier map is in-memory only (no persistence yet — that comes with help sessions).

gregpriday:
## Summary

The MCP server's `MCP_TOOL_ALLOWLIST` was filtering `listTools` responses but not enforcing anything at dispatch time — a client that already knew an action ID could call it unconditionally. This PR closes that gap and introduces a proper four-tier authorization model.

- Adds `McpTier` (`workbench` | `action` | `system` | `external`) with cumulative allowlist Sets; replaces `isAuthorized` with `resolveAuthTier`, which resolves a session's tier at connect time and gates both `listTools` and `callTool` on membership.
- Binds `POST /messages` to the originating SSE session's tier (defense-in-depth against tier-confusion via leaked `sessionId`); returns a clear `"not registered"` error if an action ID isn't in the manifest, before any widening logic runs.
- Adds in-memory `mintToken`/`revokeToken` for future internal clients (help assistant, fleet-agent terminals) — tokens are scoped to a tier and cleared on `stop()`. Backward compat preserved: the existing single-key external server maps to the `external` tier unchanged.

Resolves #6517

## Test plan

```
npx vitest run electron/services/__tests__/McpServerService.test.ts
```

36 tests passing (up from 25, with the prior permissive curated-mode dispatch test flipped to assert rejection, plus 11 new tier-authorization tests, 2 invariant tests, and 3 regression tests).

### DIFF
--- a/electron/services/McpServerService.ts
+++ b/electron/services/McpServerService.ts
@@ -463,6 +463,7 @@ export class McpServerService {
       }
     }
     this.sessions.clear();
+    this.tokenTiers.clear();
 
     for (const cleanup of this.cleanupListeners) {
       cleanup();
@@ -705,7 +706,19 @@ export class McpServerService {
         };
       }
 
-      if (!this.isActionDispatchAllowed(actionId, manifestEntry?.danger, tier)) {
+      if (manifestEntry === undefined) {
+        return {
+          content: [
+            {
+              type: "text" as const,
+              text: `Action '${actionId}' is not registered.`,
+            },
+          ],
+          isError: true,
+        };
+      }
+
+      if (!this.isActionDispatchAllowed(actionId, manifestEntry.danger, tier)) {
         return {
           content: [
             {
@@ -821,10 +834,13 @@ export class McpServerService {
    * of the listTools surface so that even a caller that already knows an
    * action ID is rejected when the tier doesn't grant it. `restricted`
    * actions are blocked unconditionally regardless of `fullToolSurface`.
+   * Caller is responsible for verifying the action is in the manifest;
+   * passing an unknown ID with `fullToolSurface=true` would otherwise reach
+   * `dispatchAction` for the renderer to reject.
    */
   private isActionDispatchAllowed(
     actionId: string,
-    danger: ActionManifestEntry["danger"] | undefined,
+    danger: ActionManifestEntry["danger"],
     tier: McpTier
   ): boolean {
     if (danger === "restricted") return false;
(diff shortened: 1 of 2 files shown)


gregpriday:
Closing — built against a pre-#6574/#6569 `McpServerService` shape that no longer exists: `isAuthorized()` now resolves both global-key and pane-token branches, and #6569's audit hook wraps every dispatch. Re-doing fresh once three decisions are settled: (1) tier mapping for pane tokens (likely `workbench`), (2) `resolveAuthTier()` replaces `isAuthorized` with tier-keyed allowlist, (3) audit-record composition records `tier` and `allowed|denied|not-registered` outcome. Issue #6517 stays open.

PR #6563, opened 2026-05-02
feat(mcp): migrate API key storage to safeStorage with Linux fallback

### LINKED ISSUE #6519: Drop persistent MCP key from electron-store; add rotate button
The MCP API key currently lives in two places: `electron-store` (plaintext) and `~/.daintree/mcp.json` (0600, also plaintext — required by the discovery contract external clients depend on). The original framing for this issue was "encrypt with `safeStorage`," but that only protects the `electron-store` copy. Anyone who can read one location can almost certainly read the other, so encrypting half the surface is theatre — and it costs a Keychain prompt on macOS, which we're not willing to spend.

The real fix is to stop double-storing. The discovery file is already the source of truth for external clients; the `electron-store` copy is redundant.

Proposed:

- Drop `apiKey` (and any encrypted variant) from the `electron-store` schema.
- The discovery file at `~/.daintree/mcp.json` becomes the sole persistence. It already lives at 0600.
- On launch, read from the discovery file; generate a fresh key + write the file if it's missing or unreadable.
- Migration: drop the legacy `apiKey` field from the store on first run after upgrade.
- Add a "Rotate API key" button in MCP settings that regenerates the key and rewrites the discovery file. Real user-facing hygiene knob (post-laptop-loan, post-incident, casual rotation) without pretending to encrypt anything.

Things this explicitly does NOT do:

- Use `safeStorage` / Keychain on macOS — see closed PR #6563 for why.
- Move the discovery file location — external clients (Claude Code, etc.) depend on `~/.daintree/mcp.json`.
- Encrypt the discovery file — same reason.

Done when:

- The store schema no longer carries the bearer token in any form.
- The migration cleanly removes the legacy field on existing installs without breaking auth.
- A "Rotate API key" button exists in settings and works while the server is running (rewrites the discovery file, no restart needed).
- External clients pick up the new key on their next request after a rotation.

Evidence:

- Current store schema: [electron/store.ts:187-192](https://github.com/daintreehq/daintree/blob/develop/electron/store.ts)
- Discovery file write: [McpServerService.ts](https://github.com/daintreehq/daintree/blob/develop/electron/services/McpServerService.ts)

gregpriday:
## Summary

- Migrates the MCP server API key from plaintext `electron-store` to Electron `safeStorage` (Keychain on macOS, DPAPI on Windows, libsecret on Linux when available).
- Migration 021 runs on first launch: encrypts any existing plaintext key into the new `apiKeyEncrypted` field and removes the legacy `apiKey` field. On Linux without a secret store, it drops the key cleanly so the auto-keygen path can recover rather than leaving plaintext behind.
- The decrypted key is cached per session in a tagged-union (`unread` / `absent` / `ok` / `undecryptable`) so `isAuthorized` can correctly deny access when a key is configured but unreadable, rather than failing open.

Resolves #6519

## Changes

- `electron/store.ts` — new `apiKeyEncrypted` field (base64 `Buffer`), legacy `apiKey` field removed from schema.
- `electron/services/McpServerService.ts` — `resolveDecryptedKey()` returns tagged-union state; `isAuthorized` denies on `undecryptable`; `start()` catches encryption failures from auto-keygen and proceeds open-access rather than crashing; `getStatus()` exposes `encryptionBackend` (`'keychain'` | `'basic_text'` | `'unavailable'`).
- `electron/services/migrations/021-mcp-api-key-safestorage.ts` — new migration with both-fields edge case handling.
- `src/components/Settings/McpServerSettingsTab.tsx` — warning banner when `encryptionBackend` is `basic_text` or `unavailable`; silence is the positive signal.

## Testing

38 test cases pass. New coverage includes:

- `safeStorage` unavailable at startup (open-access path, no crash)
- Mid-session auth-bypass denial on `undecryptable` key
- `encryptionBackend` branches including a `process.platform` stub for the Linux `basic_text` path
- Corrupted-ciphertext recovery
- Migration with both legacy and new fields present

### DIFF
--- a/electron/services/McpServerService.ts
+++ b/electron/services/McpServerService.ts
@@ -1,4 +1,4 @@
-import { ipcMain } from "electron";
+import { ipcMain, safeStorage } from "electron";
 import type { WindowRegistry } from "../window/WindowRegistry.js";
 import http from "node:http";
 import net from "node:net";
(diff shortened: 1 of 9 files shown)


gregpriday:
Closing — going to live with plaintext storage in `electron-store`.

`safeStorage` on macOS triggers a Keychain permission prompt on first encrypt/decrypt, and that surface isn't worth the protection it buys here. The MCP server only binds to `127.0.0.1`, the API key isn't a high-value secret on its own, and the prompt is the kind of thing that erodes trust faster than the encryption builds it.

Not the right tradeoff. Closing this and #6519.

PR #6581, opened 2026-05-02
chore(terminal): remove obsolete xterm 6 DECRQM workarounds and add regression guard

### LINKED ISSUE #6502: Clean up obsolete xterm 6 workarounds in terminal pipeline
Two workarounds in the per-pane terminal render pipeline are now redundant.

**DECRQM consume-and-discard CSI handlers.** xterm.js 6.0's `requestMode` handler throws `ReferenceError: i is not defined` when Vite re-minifies the already-minified `@xterm/xterm` bundle — the crash terminates the parser and drops all subsequent terminal output for that pane. Two CSI handlers in `TerminalParserHandler` (one for `CSI ? Pm $ p`, one for `CSI Pm $ p`) consume DECRQM queries to short-circuit the buggy path before the parser reaches it. That defense is now obsolete: `vite.config.ts` disables `esbuild.minifyIdentifiers` to fix the bug at the bundler layer. The parser handlers now pay an O(N) handler-chain walk on every CSI sequence ending in `p` for protection the bundler config already provides.

- Handlers: https://github.com/daintreehq/daintree/blob/758fd55c1cb093046b9dc767d49afc2abd4db27e/src/services/terminal/TerminalParserHandler.ts#L105-L121
- Bundler fix: https://github.com/daintreehq/daintree/blob/758fd55c1cb093046b9dc767d49afc2abd4db27e/vite.config.ts#L215-L221

**Private `_core._renderService.dimensions` cast for cell metrics.** `getXtermCellDimensions()` in `TerminalResizeController` reads cell width/height through a structural narrow on `_core._renderService.dimensions.css.cell`, with a comment claiming "xterm.js 6.0 does not expose a public API for per-cell pixel dimensions." That claim is stale: xterm 6's public typings expose `readonly dimensions: IRenderDimensions | undefined` with the same `css.cell` shape, and `@xterm/addon-fit`'s own `proposeDimensions()` reads the public path. Keeping the private cast carries a real future-version risk — xterm could rename `_core` or `_renderService` internally at any time and break the dimension read silently.

- Private cast and stale comment: https://github.com/daintreehq/daintree/blob/758fd55c1cb093046b9dc767d49afc2abd4db27e/src/services/terminal/TerminalResizeController.ts#L25-L59

**Constraint.** Removing the parser handlers leaves the project relying on `esbuild.minifyIdentifiers: false` as the only line of defense against the parser-killing crash. If that bundler option is ever dropped or overridden, the crash returns. The fix needs a regression check that the bundler config (or its effect on the bundled `requestMode` output) is in place — the existing compiler-bailout-report and renderer-bundle-size plugins in `vite.config.ts` are precedents for build-time invariant checks.

gregpriday:
## Summary
- Removes two DECRQM CSI handlers in `TerminalParserHandler` that became redundant after the bundler config in `vite.config.ts` started preventing the underlying minification bug at the source.
- Updates the `getXtermCellDimensions` comment to match what the installed `@xterm/xterm` typings actually expose -- xterm 6 still has no public cell-dimension API.
- Adds `xtermMinifyIdentifiersGuardPlugin`, a build-time invariant that fails the build if `esbuild.minifyIdentifiers` is accidentally turned back on.

Resolves #6502

## Changes
- **`TerminalParserHandler.ts`** -- Removed the `CSI ? Pm $ p` and `CSI Pm $ p` DECRQM consume-and-discard handlers.
- **`TerminalResizeController.ts`** -- Simplified the `getXtermCellDimensions` JSDoc; old wording was speculative, now confirmed against the installed typings.
- **`vite.config.ts`** -- Added `xtermMinifyIdentifiersGuardPlugin` that throws during `configResolved` if `esbuild.minifyIdentifiers` isn't `false`.
- **`TerminalParserHandler.test.ts`** -- Removed the DECRQM blocker test case; updated the CSI handler count assertion from 5 to 3.

## Testing
- Existing terminal parser tests pass with the updated handler count.
- Build-time guard verified by temporarily removing the config and confirming the build fails.

### DIFF
--- a/src/services/terminal/TerminalParserHandler.ts
+++ b/src/services/terminal/TerminalParserHandler.ts
@@ -102,24 +102,6 @@ export class TerminalParserHandler {
     );
     this.disposables.push(altScreenSetHandler);
 
-    // Block DECRQM (Request Mode) queries: CSI ? Pm $ p
-    // xterm.js 6.0 has a bug in its minified requestMode handler that throws
-    // "i is not defined", crashing the parser and dropping all subsequent output.
-    // OpenCode (Bubble Tea) sends these during startup for capability detection.
-    // Silently consuming them is safe — the app doesn't need DECRQM responses.
-    const decrqmHandler = terminal.parser.registerCsiHandler(
-      { prefix: "?", intermediates: "$", final: "p" },
-      () => true // Consume and discard
-    );
-    this.disposables.push(decrqmHandler);
-
-    // Also block the non-private DECRQM variant: CSI Pm $ p
-    const decrqmNonPrivateHandler = terminal.parser.registerCsiHandler(
-      { intermediates: "$", final: "p" },
-      () => true
-    );
-    this.disposables.push(decrqmNonPrivateHandler);
-
     // Block mouse reporting mode toggles (enables programs to capture mouse events).
     // We block this for agent terminals to avoid surprising interactions inside the app.
     const mouseModeParams = new Set([1000, 1002, 1003, 1005, 1006, 1015]);
--- a/src/services/terminal/TerminalResizeController.ts
+++ b/src/services/terminal/TerminalResizeController.ts
@@ -25,15 +25,17 @@ interface XtermCoreRenderDimensions {
 /**
  * Returns the CSS pixel dimensions of a single terminal cell.
  *
- * xterm.js 6.0 does not expose a public API for per-cell pixel dimensions.
- * The official @xterm/addon-fit (0.11) accesses the same private path
- * (`_core._renderService.dimensions`) — see its proposeDimensions() source.
- * Upstream tracking: xtermjs/xterm.js#702 (closed; no public API in 6.0).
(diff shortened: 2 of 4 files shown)


parsakhaz:
fwiw i think the guard here may be checking a workaround that is not sufficient in Vite 5/6.

i re-hit this in Pane with opencode and tested the emitted chunks. `esbuild: { minifyIdentifiers: false }` alone did not reliably prevent the broken `requestMode` output. the bad shape i saw was from downleveling xtermjs 6's `r ||= {}` path into an undeclared assignment like `void 0 || (r = {})`.

what worked in the matrix:

- `build.target: 'es2021'` or higher
- `build.minify: 'terser'`

what did not work by itself:

- `esbuild.minifyIdentifiers: false`
- `build.target: 'modules'` / `es2020`

so i would probably guard the emitted bundle shape directly, or guard that the build target is at least es2021, instead of only guarding `minifyIdentifiers`. we added a post-build emitted-chunk check in Pane here: https://github.com/dcouple/Pane/pull/176

PR #6906, opened 2026-05-06
fix(setup): respect window focus in CLI poll and health check

### LINKED ISSUE #6869: Wizard CLI polling and health check ignore window focus changes
The agent-setup wizard's CLI-availability poll (`useAgentSetupPoll`) gates on `document.hidden`, but in Chromium 146 `document.hidden` only flips on minimize or full occlusion — not on OS-level focus loss. When the user opens a side-by-side terminal to install an agent CLI, the wizard keeps polling at full rate (every 3s) while the user is in another app, draining CPU and battery without doing useful work.

When the user *returns* from that terminal, two related friction points show up: the wizard waits up to ~3s before noticing the new binary, and the system-requirements section (`useSystemHealthCheck`) doesn't auto-recheck at all — the user has to click the manual "Re-check" button to pick up a freshly-installed Node or Git.

The renderer-side `window.addEventListener("focus", ...)` signal does fire on actual OS focus changes and is already the established pattern in this codebase — six hooks already use it (`useAgentLauncher`, `useGitHubResourceListSWR`, `useFleetRibbonFlashes`, `useWindowNotifications`, `useReEntrySummary`, `TerminalReflowController`). The wizard's two polling surfaces should hook into the same signal.

## Evidence

- Polling logic gated on `document.hidden`: [`useAgentSetupPoll.ts`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Setup/useAgentSetupPoll.ts)
- Health check that only runs on mount: [`useSystemHealthCheck.ts`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Setup/useSystemHealthCheck.ts)
- Established renderer focus pattern reference: [`useAgentLauncher.ts:171`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/hooks/useAgentLauncher.ts#L171)

## Constraints

- Multi-window scenarios via `ProjectViewManager` should not fire spurious focus events when the user switches between two project windows in the same app
- The main-process `WindowFocusThrottle` (PR #4838) handles main-process polling services; this is renderer-only and shouldn't duplicate that infrastructure

gregpriday:
## Summary

- `useAgentSetupPoll` was gating on `document.hidden`, which only flips on minimize or full occlusion — not on OS-level focus loss. Switched to `window` `focus`/`blur` events to correctly detect when the user has stepped away to a side-by-side terminal.
- Polling now backs off when the window loses focus and immediately retries on focus regain when the last health check failed, so the wizard reacts quickly when the user returns with a freshly-installed CLI.
- `useSystemHealthCheck` now re-runs on window focus, not just on mount — no more manual "Re-check" click needed after installing Node or Git.

Resolves #6869

## Changes

- `src/components/Setup/useAgentSetupPoll.ts` — replace `document.hidden` gate with `focus`/`blur` listeners; add immediate retry on focus after failed check
- `src/components/Setup/useSystemHealthCheck.ts` — add `focus` listener to re-run health check on window regain
- `src/components/Setup/__tests__/useAgentSetupPoll.test.ts` — tests covering focus-aware polling behavior
- `src/components/Setup/__tests__/useSystemHealthCheck.test.ts` — tests covering focus-triggered re-check

## Testing

Both hooks have unit tests covering the new focus-aware paths. Manually verified that switching away and back to the window while the wizard is open triggers the expected polling and health-check behavior.

### DIFF
--- a/src/components/Setup/__tests__/useAgentSetupPoll.test.ts
+++ b/src/components/Setup/__tests__/useAgentSetupPoll.test.ts
@@ -25,6 +25,14 @@ function setHidden(hidden: boolean) {
   document.dispatchEvent(new Event("visibilitychange"));
 }
 
+function dispatchFocus() {
+  window.dispatchEvent(new Event("focus"));
+}
+
+function dispatchBlur() {
+  window.dispatchEvent(new Event("blur"));
+}
+
 describe("useAgentSetupPoll", () => {
   beforeEach(() => {
     vi.useFakeTimers();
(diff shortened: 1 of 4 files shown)


gregpriday:

PR #6907, opened 2026-05-06
fix(error-boundary): make Report Issue deeplink reliable for real crashes

### LINKED ISSUE #6884: Make Report-Issue deeplink reliable for crashes with real stacks
The boundary's "Report Issue" deeplink will fail with HTTP 414 on any non-trivial component crash, and the user-visible incident ID isn't searchable in Sentry.

`handleReport` in [`ErrorBoundary.tsx`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/ErrorBoundary/ErrorBoundary.tsx#L129-L158) builds `github.com/.../issues/new?body=...` with the full `error.stack`, full `errorInfo.componentStack`, and a JSON-stringified context inline. GitHub's Nginx layer hard-caps the URL at 8,192 chars (414 URI Too Long beyond that). React component stacks at the 122 boundary wrap sites are routinely 3–15KB, so a real crash report opens to an error page instead of a prefilled issue.

VS Code's pattern for the same problem (`src/vs/workbench/contrib/issue/`) is to budget the URL well under 8KB, truncate stacks/logs with a placeholder, and copy the full payload to the clipboard before opening the URL so the user can paste the rest into the issue body. For genuinely over-budget reports, drop to a clipboard-only flow with a toast explaining what happened.

Separately: the boundary generates a fresh `correlationId` via `crypto.randomUUID` (line 54) and discards Sentry's `eventId` returned synchronously by `captureException`. The `correlationId` is what the user sees as "Error ID", but engineers can only search Sentry by `eventId`, so a user pasting the displayed ID into a support chat doesn't lead anywhere useful. `captureRendererException` in [`rendererSentry.ts`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/utils/rendererSentry.ts#L74) is currently typed `void` and would need to expose the `eventId` so the boundary can display it as the incident ID.

gregpriday:
## Summary

- Caps the GitHub issue URL at 8 KB to avoid HTTP 414 on large component stacks, truncating with a `[... truncated ...]` placeholder so the structure is still readable
- Copies the full error payload to the clipboard before opening the URL; falls back to clipboard-only with a toast when the report is still over-budget after truncation
- Wires up the Sentry `eventId` returned by `captureException` so the Error ID shown to the user is the same one engineers can search for in Sentry

Resolves #6884

## Changes

- `ErrorBoundary.tsx`: `handleReport` now builds the URL under a hard byte cap, copies full payload first, and shows a clipboard-only toast when needed
- `rendererSentry.ts`: `captureRendererException` now returns the Sentry `eventId` instead of `void`
- Tests updated to cover URL budget logic, truncation, clipboard fallback, and the `eventId` plumbing

## Testing

- Unit tests for the URL budget, truncation placeholder, clipboard-then-open flow, and clipboard-only fallback
- Verified `captureRendererException` return type and `eventId` propagation through the boundary

### DIFF
--- a/src/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary/ErrorBoundary.tsx
@@ -35,7 +35,25 @@ interface ErrorBoundaryState {
 const ISSUE_URL_BUDGET = 7200;
 const STACK_TOP_LINES = 15;
 const STACK_BOTTOM_LINES = 5;
+const TITLE_MESSAGE_CAP = 200;
 const TRUNCATION_PLACEHOLDER = "\n… (middle truncated — full report copied to clipboard) …\n";
+
+// `encodeURIComponent` throws `URIError: URI malformed` on lone (unpaired)
+// surrogates, which can appear in errors originating from native modules,
+// WASM, or mis-decoded byte streams. Replace any unpaired surrogate with the
+// Unicode replacement char so encoding always succeeds inside an error path
+// where throwing again would lose the report entirely.
+function sanitizeForUriComponent(value: string): string {
+  return value.replace(
+    /[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g,
+    "�"
+  );
+}
+
+function capMessage(message: string, cap = TITLE_MESSAGE_CAP): string {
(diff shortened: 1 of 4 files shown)


gregpriday:

PR #6915, opened 2026-05-06
fix(settings): remove double-announcing role="alert" from inline field errors

### LINKED ISSUE #6883: Settings form primitives need a11y polish on field-error semantics
Two complementary a11y issues on the shared settings form primitives.

**1. `role="alert"` on inline field errors interrupts typing.** The W3C ARIA Authoring Practices Guide treats `role="alert"` as an assertive live region that interrupts the user's current announcement, and explicitly warns it should be used sparingly because it can be intrusive during active typing. Today every field-error `<p>` in the shared primitives carries it:

```typescript
{error && (
  <p id={errorId} role="alert" className="text-xs text-status-error">
    {error}
  </p>
)}
```

This appears in `SettingsInput`, `SettingsTextarea`, `SettingsSelect`, `SettingsCheckbox`, and `SettingsChoicebox`. The error is also wired through `aria-describedby` on each input, and the input carries `aria-invalid={error ? true : undefined}` — so screen readers will announce the error on focus or pause anyway. The `role="alert"` is double-announcing during typing and competing with the user's own keystroke feedback.

`TerminalAppearanceTab` and `PortalSettingsTab` have a few ad-hoc field-error blocks with the same pattern and should be tidied along the way.

The form-level error banners in `SettingsDialog` (project load failures, autosave failures) and in `EnvironmentSettingsTab` (save failures) are different — those are recovery-required interruptions that arrive on async events the user isn't actively engaged with. Those should keep `role="alert"`.

**2. Raw env-var inputs in `EnvironmentSettingsTab` don't mark themselves invalid.** The shared primitives all do `aria-invalid={error ? true : undefined}` correctly, and `EnvVarEditor` already follows the same pattern:

```typescript
aria-invalid={isEmptyKey || isDuplicate ? "true" : undefined}
```

But the raw `<input>` elements in [`EnvironmentSettingsTab`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Settings/EnvironmentSettingsTab.tsx#L260-L320) (the env-var name and value fields) render an error `<p>` below the row when `rowErrors[envVar.id]` is set, but never mark the input itself as invalid. Screen reader users see the error visually but don't hear the field announce as invalid on focus.

Evidence:

- [`role="alert"` in `SettingsInput`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Settings/SettingsInput.tsx#L97-L101) — same pattern repeats across the four other primitives
- [`EnvVarEditor` showing the right `aria-invalid` pattern](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Settings/EnvVarEditor.tsx#L260-L270)

W3C also notes that `aria-invalid="false"` is equivalent to omitting the attribute and that some legacy screen readers announce "invalid: false" — the existing `error ? true : undefined` pattern is correct and worth keeping consistent across all input surfaces.

gregpriday:
## Summary

- Removed `role="alert"` from inline field-error `<p>` elements in five shared settings primitives (`SettingsInput`, `SettingsTextarea`, `SettingsSelect`, `SettingsCheckbox`, `SettingsChoicebox`) and three ad-hoc blocks (`TerminalAppearanceTab`, `PortalSettingsTab`). The `aria-describedby` + `aria-invalid` pairing already covers screen reader announcement — the live region was double-announcing on every keystroke.
- Wired `aria-invalid` and `aria-describedby` onto the raw env-var key `<input>` in `EnvironmentSettingsTab`, and gave the row error `<p>` a stable `id`. Kept `role="alert"` on async save-error banners — those are legitimate interruptions, not inline validation feedback.
- Fixed `SettingsCheckbox` to omit `aria-invalid` entirely when there's no error (was emitting `"false"`, which some screen readers announce as "invalid: false").

Resolves #6883

## Changes

- `SettingsInput`, `SettingsTextarea`, `SettingsSelect`, `SettingsCheckbox`, `SettingsChoicebox`: drop `role="alert"` from field-error paragraphs
- `TerminalAppearanceTab`, `PortalSettingsTab`: same cleanup on ad-hoc field-error blocks
- `EnvironmentSettingsTab`: add `aria-invalid` + `aria-describedby` to the key input, stable error `id`, `role="alert"` on `saveError`
- `SettingsCheckbox`: omit `aria-invalid` attribute when no error present
- `SettingsFormPrimitives.test.tsx`: updated 4 assertions from `getByRole("alert")` to `getByText()` + `queryByRole("alert")` nullity checks; added `SettingsCheckbox` no-error regression
- `EnvironmentSettingsTab.test.tsx`: new tests covering key input ARIA wiring, duplicate-key validation, and `saveError` live region

## Testing

- Unit tests updated and passing — `SettingsFormPrimitives.test.tsx` and the new `EnvironmentSettingsTab.test.tsx` cover the ARIA attribute changes and regression cases
- Typecheck, lint, and format all pass


gregpriday:

PR #6917, opened 2026-05-06
fix(settings): flush unsaved edits on view detach (#6875)

### LINKED ISSUE #6875: Settings dialog can lose unsaved edits when project view detaches
The settings dialog has two save paths that can silently drop in-flight edits when the renderer is torn down.

`EnvironmentSettingsTab` tracks an explicit `isDirty` snapshot with Save/Discard buttons but never flushes on dialog close — the dialog's `handleClose` only awaits `projectForm.flush()`. If the user types a new env-var key/value and clicks the dialog X before pressing Save, the row is silently dropped.

`useProjectSettingsForm` debounces autosave and exposes `flush()`, which is wired to dialog close — but only via React's unmount path. That breaks under WebContentsView detach in Electron 41. The codebase already documents this exact quirk in `HelpPanel.tsx`:

> In Electron 41, beforeunload does not fire on WebContentsView detach, but visibilitychange does — this covers both project switches and window unload.

PR #5009 fixed the same failure class for terminal panels (webContents was being destroyed before the debounced 500ms IPC flush completed, so the snapshot never landed). The settings dialog hits the same path on project switch — edit a project setting, switch projects via the picker, the project view evicts before the autosave debounce fires.

Evidence:

- [`useProjectSettingsForm.flush()`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/hooks/useProjectSettingsForm.ts#L442-L450) — exists, but only entry point is `SettingsDialog.handleClose`
- [`SettingsDialog.handleClose`](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/Settings/SettingsDialog.tsx#L340-L348) — no per-tab dirty-state flush
- [`HelpPanel.tsx` Electron 41 quirk comment](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/components/HelpPanel/HelpPanel.tsx#L231-L235)
- [`panel/resource.ts` existing beforeunload pattern for terminal persistence](https://github.com/daintreehq/daintree/blob/ef36c6caffd5d920a8883a814ecc33635b32323b/src/store/listeners/panel/resource.ts#L31-L36)
- `.lessons/_index.jsonl` PR 5009 — prior incident with same failure class

Data loss bugs that survived a previous postmortem deserve to be closed for good. The codebase has the right patterns in-house already (`visibilitychange` listener, synchronous-payload IPC); it's a question of plumbing them into the settings dialog's save paths.

gregpriday:
## Summary

- The settings dialog had two save paths that could silently drop unsaved edits when a WebContentsView was detached. `useProjectSettingsForm` was cancelling its debounce on close instead of flushing, and `EnvironmentSettingsTab` had no flush path at all outside of an explicit Save click.
- Root cause: Electron 41 doesn't fire `beforeunload` on WebContentsView detach, but `visibilitychange` does. `HelpPanel.tsx` already documents this quirk and uses the same pattern.
- Fixed by wiring a `visibilitychange` listener into `SettingsDialog` that flushes both the project form and the env tab via `Promise.allSettled`, and exposing a `flushRef` on `EnvironmentSettingsTab` so the dialog can trigger that flush without an explicit Save click.

Resolves #6875

## Changes

- `useProjectSettingsForm.ts`: flush the pending debounce on close instead of cancelling it; `flush()` skips a redundant persist when the before/after snapshots are identical.
- `SettingsDialog.tsx`: adds a `visibilitychange` listener (mirrors `HelpPanel.tsx`) that flushes `projectForm` and `envFlushRef` via `Promise.allSettled`; `handleBeforeClose` and `handleClose` also flush `envFlushRef`.
- `EnvironmentSettingsTab.tsx`: exposes a `flushRef` prop; callers can trigger a flush of dirty rows on close or detach without requiring a Save click. Empty-key drafts and invalid keys are silently dropped on flush, matching the existing `envVarsToRecord` sanitize-on-save policy.
- `settingsTabRegistry.tsx`: adds `needsFlushRef` flag on the environment entry; `LazyTabContent` forwards `envFlushRef` as `flushRef` when the flag is set.

## Testing

- `useProjectSettingsForm.test.tsx`: replaced "cancels pending debounce on close" (was asserting the old broken behavior) with "flushes pending debounce on close (#6875)"; added "flush() is a no-op when nothing has changed".
- `EnvironmentSettingsTab.test.tsx`: added 3 tests covering `flushRef` saves dirty rows, silently drops empty-key drafts, and is a no-op when not dirty.
- 48/48 tests pass across the 3 affected test files.


gregpriday:
CI failed post-rebase (Doherty 400ms timing test from #6904). Closing and returning issue to work pool — likely a timing flake; reattempt after #6904's test is stabilized.

PR #6954, opened 2026-05-06
feat(assistant): hibernate idle Daintree Assistant by capturing Claude resume code

### LINKED ISSUE #6949: Hibernate idle Daintree Assistant by capturing the Claude resume code
Pivoting this away from the original plan (hibernation timer + chat-scrollback rehydration + a vague "auto-relaunch into the same session" step) toward reusing the per-agent resume infrastructure we already ship. The whole feature collapses down to: capture the resume code on hibernation, persist it, replay it on reopen.

The Daintree Assistant currently only supports Claude Code, so this issue is scoped to Claude. Multi-agent support comes later, agent-by-agent, and the same machinery covers it.

## Infrastructure already in place

- `shared/config/agentRegistry.ts` defines an `AgentResume` discriminated union with kinds `session-id`, `rolling-history`, `named-target`, `project-scoped`.
- `shared/config/agents/claude.ts` declares Claude as `session-id` kind with `args: (sessionId) => ["--resume", sessionId]` and `sessionIdPattern: "claude --resume ([\\w-]+)"`.
- `electron/services/pty/TerminalGracefulShutdown.ts` already sends the agent's `quitCommand` or `shutdownKeySequence`, runs `sessionIdPattern` against buffered output, and stores the captured ID on `terminal.agentSessionId`.
- `shared/types/agentSettings.ts` exports `buildResumeCommand(agentId, sessionId, launchFlags?)`, which returns the full resume command (for Claude, `claude --resume abc-123`).

## What's missing for the assistant

- The assistant's PTY is spawned with `ephemeral: true` via the `agent.launch` action in `src/components/HelpPanel/HelpPanel.tsx`, so it never flows through `TerminalGracefulShutdown`.
- The `visibilitychange` handler at `HelpPanel.tsx:234-286` calls `removePanel()` directly, bypassing the graceful path.
- `src/store/helpPanelStore.ts` doesn't persist `agentSessionId` across restarts. `partialize` (lines 85-90) only saves `isOpen`, `width`, `preferredAgentId`, `introDismissed`.
- Help terminals are excluded from project rehydration in `electron/ipc/handlers/terminal/snapshots.ts:188`.

## Proposed work

1. After `N` minutes of the panel being hidden continuously (default 30, configurable under `helpAssistant` settings), route the assistant's PTY through `TerminalGracefulShutdown` instead of `removePanel()`. Capture the Claude session ID into `helpPanelStore` and persist it.
2. On reopen, if a stored session ID exists, build the launch command via `buildResumeCommand("claude", sessionId)` and spawn through the same `agent.launch` flow.
3. If capture fails (agent crashed mid-shutdown, regex didn't match), fall back to a fresh launch and surface a one-line ambient banner.
4. Keep the existing `+ New session` button at `HelpPanel.tsx:826-835` and have it also clear the persisted session ID.
5. If the agent is `working` when the timer fires, defer the kill until it goes idle. Don't interrupt mid-turn.

**Acceptance:** an idle hidden assistant returns to baseline memory within ~5 minutes of the timer firing, and reopening reconnects to the prior Claude session without the user re-pasting context.

## Design questions worth answering before implementing

- Hibernation timer fixed, or driven by `ResourceProfileService`? Probably fixed for v1; revisit if memory pressure across worktrees becomes the dominant signal.
- Should the timer pause when another window or project is foregrounded? Probably not. A hidden assistant in window A while you're working in window B is still not engaged.
- One-line ambient banner on reopen when a resume happened, or silent? Probably banner. The user should know the agent restarted under the hood.
- For the eventual multi-agent case, the same `AgentResume` machinery covers it, but `rolling-history` agents have different semantics and will need per-agent UX thought when they come online.

## Out of scope

- Adapting non-Claude agents now. Add support agent-by-agent as the assistant gains them.
- Hibernating active grid agent terminals; different registry, different lifecycle, separate problem.
- Snapshotting in-flight tool calls or partial assistant turns.

gregpriday:
## Summary

- After a configurable idle period (default 30 min; options: 0/15/30/60/120) of the panel being hidden and the agent being idle, the assistant PTY is gracefully shut down via the existing `TerminalGracefulShutdown` infrastructure to capture the Claude `--resume` token. On reopen, `buildResumeCommand` spawns a resumed PTY directly — no context lost.
- A tier-1 ambient banner ("Resumed your previous session.") confirms the resume and auto-dismisses after 10 seconds. Hibernation is deferred while the agent is working/waiting/directing, skipped when the OS is suspended, and the `+ New session` button clears the persisted token so the just-discarded chat isn't auto-resumed.
- `helpPanelStore` bumped to v3 with sanitized rehydration — persists a `Record<projectId, { sessionId, cwd, agentId }>`. Custom launch flags thread through `buildResumeCommand` and `agentLaunchFlags` so user CLI flags (e.g. `--model`) survive the round-trip.

Resolves #6949

## Changes

- `HelpPanel.tsx` — hibernate timer, graceful shutdown path, resume-spawn logic, ambient banner, OS-suspend guard, race-condition fix (panel reopened during `gracefulKill` no longer tears down the visible terminal once IPC resolves)
- `helpPanelStore.ts` — v3 store with `hibernateData` record, `idleHibernateMinutes` setting, sanitized `partialize`/rehydration
- `DaintreeAssistantSettingsTab.tsx` — idle hibernate minutes select (0/15/30/60/120)
- `helpAssistant.ts` IPC handler + `shared/types/ipc/api.ts` — `getHibernateData` / `setHibernateData` channels
- 13 hibernation component tests, 6 store tests, 6 settings handler tests (654 unit tests total, all passing)

## Testing

- Unit tests: 654 passing, including 25 new tests covering hibernation timer, defer-while-working, OS-suspend guard, race fix, resume-spawn path, store v3 migration, and settings handler
- Typecheck: clean
- Lint: clean

### DIFF
--- a/electron/ipc/handlers/__tests__/helpAssistant.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/helpAssistant.handlers.test.ts
@@ -215,7 +215,9 @@ describe("registerHelpAssistantHandlers", () => {
   });
 
   it("rejects an out-of-range stored idleHibernateMinutes and falls back to default", async () => {
-    storeMock.get.mockReturnValue({ idleHibernateMinutes: 999 });
+    storeMock.get.mockReturnValue({
+      idleHibernateMinutes: 999,
+    } as unknown as Partial<HelpAssistantSettings>);
     registerHelpAssistantHandlers();
     const handler = ipcMainMock._handlers.get(GET_CHANNEL)!;
 
--- a/electron/ipc/handlers/__tests__/helpAssistant.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/helpAssistant.handlers.test.ts
@@ -58,6 +58,7 @@ describe("registerHelpAssistantHandlers", () => {
       skipPermissions: false,
       auditRetention: 7,
       customArgs: "",
+      idleHibernateMinutes: 30,
     });
   });
 
@@ -73,6 +74,7 @@ describe("registerHelpAssistantHandlers", () => {
       skipPermissions: true,
       auditRetention: 30,
       customArgs: "",
+      idleHibernateMinutes: 30,
     });
   });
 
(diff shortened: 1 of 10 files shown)


gregpriday:
Closing — will redo this issue from scratch after #6955 lands. Reverting branch artifacts so the next attempt starts clean.

PR #7174, opened 2026-05-07
fix(quarantine): tidy quarantine prefix list and add per-file reap logging

### LINKED ISSUE #7106: Tidy quarantine cleanup prefix list and add reap logging
A few small polish items in `electron/services/projectQuarantineCleanup.ts`:

1. **Dead `workflows.json.corrupted` prefix.** `QUARANTINE_PREFIXES` includes `workflows.json.corrupted.` and `workflows.json.corrupted`, but no producer for that file exists anywhere in `electron/`, `shared/`, or `src/`. The corresponding test entry can go too.

2. **Redundant no-dot prefix variants.** Every prefix appears twice, once with a trailing dot and once without. Every actual producer writes with the dot-suffixed form (`${filePath}.corrupted.${Date.now()}`), so the no-dot variants are unnecessary and they accidentally match neighbours like `state.json.corruptedXYZ`. Producers verified at `ProjectStateManager.ts:219`, `ProjectFileStore.ts:36`, `ProjectSettingsManager.ts:246`, `GlobalFileStore.ts:35`, `electron/store.ts:432`.

3. **No per-file forensic trail.** The cleanup has a silent `catch { continue }` on the inner `readdir(projectDir)` ([`projectQuarantineCleanup.ts:42-45`](https://github.com/daintreehq/daintree/blob/d35c5c3882206e6e36114476da7ae9bf78e987e6/electron/services/projectQuarantineCleanup.ts#L42-L45)) and the only deletion log is the aggregate "Cleaned up N files" at the end. When a user reports a cleanup-related bug there's no breadcrumb. Logging needs to avoid the absolute `filePath` since it includes the user's home directory; project ID, filename, and computed age are enough.

Pre-release polish.

gregpriday:
## Summary
- Removes the dead `workflows.json.corrupted` prefix and drops redundant no-dot variants from `QUARANTINE_PREFIXES`
- Adds per-file `projectQuarantine.reaped` logging so cleanup deletions leave a forensic trail
- Tightens the age comparison to compute `ageMs` once and reuse it

Resolves #7106

## Changes
- `electron/services/projectQuarantineCleanup.ts` — trimmed prefix list from 8 entries to 3, added per-reap logging, cleaned up age computation
- `electron/services/__tests__/projectQuarantineCleanup.test.ts` — updated existing assertions, added spy-based test for per-file log output

## Testing
- All existing quarantine unit tests pass (adjusted for the reduced prefix count)
- New test verifies `logInfo` is called with `projectQuarantine.reaped` and the correct payload shape

### DIFF
--- a/electron/services/__tests__/projectQuarantineCleanup.test.ts
+++ b/electron/services/__tests__/projectQuarantineCleanup.test.ts
@@ -1,8 +1,9 @@
-import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import fs from "fs/promises";
 import path from "path";
 import os from "os";
 import { cleanupQuarantinedProjectFiles } from "../projectQuarantineCleanup.js";
+import * as logger from "../../utils/logger.js";
 
 const THIRTY_ONE_DAYS_MS = 31 * 24 * 60 * 60 * 1000;
 const TWENTY_NINE_DAYS_MS = 29 * 24 * 60 * 60 * 1000;
@@ -14,7 +15,6 @@ const QUARANTINE_FILES = [
   "state.json.corrupted.1234567890",
   "settings.json.corrupted.1234567890",
   "recipes.json.corrupted.1234567890",
-  "workflows.json.corrupted.1234567890",
 ];
 
 let tmpDir: string;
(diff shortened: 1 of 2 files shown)


gregpriday:
This PR was branched before #7159 merged earlier today. Its diff cleanly applies against the pre-#7159 file shape, but on top of current `develop` it reverts most of #7159's hardening:

- Drops `cleanupGlobalQuarantineFiles` and `cleanupUserDataRootQuarantineFiles` — loses sweep of `${userData}/global/recipes.json.corrupted.*` and `${userData}/config.json.corrupted.*`
- Drops the `state.json.future-v` downgrade prefix
- Reverts `fs.lstat` → `fs.stat` (symlink-follow regression on age check)
- Removes the `Math.max(0, now - mtimeMs)` clock-skew clamp
- Drops ~21 tests covering the new sweeps and isolated-failure path

A rebase onto current `develop` needs three deliberate calls:

1. Keep or drop `workflows.json.corrupted.*` (issue #7106 says drop because no producer — verify before deleting)
2. Keep or drop the no-dot variants (#7106 says drop because every producer writes the dot-suffixed form)
3. Port the per-file `projectQuarantine.reaped` logging into the shared `sweepDirectoryForPrefixes` helper so it covers all three sweep paths, not just per-project

The age-comparison "tighten ageMs once" change in this PR is already in develop via #7159, so the rebased scope reduces to prefix-list trim + helper logging.

Deferring autonomously since the rebase needs human judgment on (1) and (2).

gregpriday:
Closing — needs rebase + integration with #7159 (already merged). Reopen on a fresh branch off current develop.

PR #7178, opened 2026-05-07
chore(scratch): polish store lifecycle and platform hygiene

### LINKED ISSUE #7122: Scratch-store lifecycle and platform-hygiene polish
A basket of small polish items in the scratch-store subsystem before the release. Each is XS effort, all confined to [`electron/services/ScratchStore.ts`](https://github.com/daintreehq/daintree/blob/49d7251e659e94f0deebcb900b0fd2b1021334ad/electron/services/ScratchStore.ts), [`electron/services/scratchStorePaths.ts`](https://github.com/daintreehq/daintree/blob/49d7251e659e94f0deebcb900b0fd2b1021334ad/electron/services/scratchStorePaths.ts), [`electron/services/ScratchCleanupService.ts`](https://github.com/daintreehq/daintree/blob/49d7251e659e94f0deebcb900b0fd2b1021334ad/electron/services/ScratchCleanupService.ts), and [`electron/ipc/handlers/scratch/index.ts`](https://github.com/daintreehq/daintree/blob/49d7251e659e94f0deebcb900b0fd2b1021334ad/electron/ipc/handlers/scratch/index.ts). The bigger delete-ordering rework is tracked separately.

## Dead `scratchStateFilePath` helper

`scratchStorePaths.ts` exports a `scratchStateFilePath` helper and a `STATE_FILENAME = "state.json"` constant. No production code reads or writes a per-scratch `state.json` — only `__tests__/scratchStorePaths.test.ts` references the helper. Likely a vestige of a JSON-store design that was superseded by the SQLite migration.

## `SCRATCH_TTL_MS` re-export alias

`ScratchCleanupService.ts` re-exports `SCRATCH_CLEANUP_TTL_MS` as `SCRATCH_TTL_MS`. The only consumer is the cleanup test, which should import `SCRATCH_CLEANUP_TTL_MS` from `@shared/config/scratchCleanup` directly.

## Redundant `existsSync` guards

Several `existsSync` checks sit in front of `fs.mkdir(..., { recursive: true })` and `fs.rm(..., { force: true })` calls in `ScratchStore` and `ScratchCleanupService`. Both options already make the underlying calls idempotent on missing or existing paths, so the guards are dead weight — and slightly racy.

## Duplicate root mkdir in `createScratch`

`ScratchStore.createScratch` re-creates the scratch root directory on every call. `ScratchStore.initialize` already does this once at boot via `electron/window/windowServices.ts:326`, and the inner per-scratch `fs.mkdir(..., { recursive: true })` would create the parent anyway.

## Case-insensitive `UUID_V4_REGEX`

`UUID_V4_REGEX` in `scratchStorePaths.ts` carries the `/i` flag and accepts uppercase IDs. `crypto.randomUUID()` always emits lowercase, so the canonical generator only ever produces matching IDs. On case-insensitive filesystems (default APFS, NTFS), an uppercase ID points at the same on-disk directory as its lowercase form but a different SQLite row — split-brain class of bug if a non-canonical caller ever appears. Drop the flag and let validation fail loudly at the boundary.

## Missing `fs.rm` retries on Windows

The two `fs.rm` calls in `ScratchStore.removeScratch` and `ScratchCleanupService.runScratchCleanup` only pass `{ recursive: true, force: true }`. On Windows, Search and antivirus scanners briefly lock directories after modification, which surfaces as transient `EPERM` / `EBUSY`. Node's `fs.rm` has supported `maxRetries` and `retryDelay` well within the project's Node 22+ baseline; passing them costs nothing on the success path.

## No-fsync gap in save-as-project

The `fs.cp` in the save-as-project IPC handler does not fsync the destination tree before resolving. A power loss between IPC resolution and the next OS flush can leave the destination partial. The original scratch is preserved as a recovery path (already documented inline), but the durability gap deserves a one-line comment so the failure-tolerance contract is visible to future readers.

## Unhandled `db.insert` failure in `createScratch`

`createScratch` runs `fs.mkdir` then `db.insert(...).run()` with no try/catch. A UUIDv4 collision is essentially impossible, but a `SQLITE_BUSY` past `busy_timeout` or a disk-full mid-insert leaves an empty orphaned directory with no DB row. Wrapping the insert and `fs.rm`-ing the freshly-created directory on failure mirrors the rollback discipline of the delete path.

gregpriday:
## Summary
- Removes dead code, redundant guards, and duplicate work from the scratch store subsystem
- Hardens Windows resilience with `fs.rm` retry options and tightens UUID validation
- Adds rollback discipline to scratch creation so a failed DB insert doesn't leave orphaned directories

Resolves #7122

## Changes
- Remove dead `scratchStateFilePath` helper and `STATE_FILENAME` constant
- Remove `SCRATCH_TTL_MS` re-export alias; import `SCRATCH_CLEANUP_TTL_MS` directly
- Remove redundant `existsSync` guards before `mkdir({ recursive: true })` and `rm({ force: true })`
- Remove duplicate root `mkdir` from `createScratch` (already done in `initialize`)
- Drop case-insensitive flag from `UUID_V4_REGEX` to reject non-canonical IDs
- Add `maxRetries` and `retryDelay` to `fs.rm` calls for Windows lock resilience
- Wrap `db.insert` in `createScratch` with directory rollback on failure
- Document fsync gap in save-as-project with recovery-path comment

## Testing
- Updated existing unit tests for `ScratchCleanupService` and `scratchStorePaths`
- Added new `ScratchStore` tests covering create and remove lifecycle

### DIFF
--- a/electron/ipc/handlers/scratch/index.ts
+++ b/electron/ipc/handlers/scratch/index.ts
@@ -203,6 +203,8 @@ export function registerScratchHandlers(deps: HandlerDependencies): () => void {
     }
 
     try {
+      // No fsync: destination may be partial after a power loss before OS flush.
+      // The original scratch is preserved (copy-not-move), so the user can re-save.
       await fs.cp(scratch.path, destinationPath, {
         recursive: true,
         preserveTimestamps: true,
--- a/electron/services/ScratchCleanupService.ts
+++ b/electron/services/ScratchCleanupService.ts
@@ -8,23 +8,21 @@
  * (per `app_state.currentScratchId`) is always excluded so an actively-open
  * workspace can never disappear under the user.
  *
- * Tombstoning is one-way — orphaned directories left by a failed `fs.rm`
- * stay on disk (logged, not retried). Accepting that orphan rate is
- * deliberate: the alternative (re-sweeping tombstoned rows) would require a
- * second query and could surprise users who manually re-create folders at
- * the same path.
+ * Tombstoning is one-way — orphaned directories left by a persistently
+ * failing `fs.rm` stay on disk (logged). `fs.rm` retries transient locks
+ * (`maxRetries` / `retryDelay`), but permanent failures are not re-queued.
+ * Accepting that orphan rate is deliberate: the alternative (re-sweeping
+ * tombstoned rows) would require a second query and could surprise users
+ * who manually re-create folders at the same path.
  *
  * Mirrors the fire-and-forget pattern of `initializeTrashedPidCleanup`:
  * called once at app boot, never awaited, never throws — a cleanup failure
  * must not block startup.
  */
 import fs from "fs/promises";
-import { existsSync } from "fs";
 import { scratchStore as defaultScratchStore } from "./ScratchStore.js";
 import { logError, logInfo } from "../utils/logger.js";
 import { SCRATCH_CLEANUP_TTL_MS } from "../../shared/config/scratchCleanup.js";
-
-export { SCRATCH_CLEANUP_TTL_MS as SCRATCH_TTL_MS } from "../../shared/config/scratchCleanup.js";
 
 export interface ScratchCleanupResult {
   /** Total rows examined as candidates (predate cutoff, not yet tombstoned). */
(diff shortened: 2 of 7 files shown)


gregpriday:
PR #7150 (tombstone-first delete ordering) merged earlier today and touches the same regions of `ScratchCleanupService.ts` (candidate loop) and `ScratchStore.ts` (`removeScratch` directory cleanup) that this PR rewrites. GitHub reports `CONFLICTING` and the textual conflict is real even though the semantic intent of both PRs aligns — both polish the scratch subsystem in the same direction.

Deferring autonomously. Rebase against `develop` and reconcile the cleanup-service loop conflict; the rest of this PR (dead-code removal, redundant `existsSync` removal, Windows retry options on `fs.rm`, `db.insert` rollback) applies cleanly on top.

gregpriday:
Closing — needs rebase + integration with #7150 (already merged). Reopen on a fresh branch off current develop.

PR #7324, opened 2026-05-07
perf(worktree): tighten filter and store pipeline

### LINKED ISSUE #7221: Tighten worktree filter and store pipeline
Six small cleanups in the worktree filter and store pipeline that should land together as one polish PR. None of them changes user-visible behavior except #1, which makes `feature-2` sort before `feature-10`.

## 1. `localeCompare` calls without options

Five callsites in [`src/lib/worktreeFilters.ts:257-272`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/lib/worktreeFilters.ts#L257-L272) (lines 257, 263, 269, 272) and [`src/hooks/useWorktrees.ts:59`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/hooks/useWorktrees.ts#L59) call `a.name.localeCompare(b.name)` with no options. Two consequences: a fresh internal collator is allocated per call, and `feature-2` sorts after `feature-10`.

A module-scoped `Intl.Collator` with `{ numeric: true, sensitivity: "base" }` fixes both. Keep this scoped to the worktree pipeline — the codebase has ~30 other `localeCompare` callsites that aren't part of this issue.

## 2. `applySnapshot` always rebuilds the worktrees Map

[`createWorktreeStore.ts:65-76`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/store/createWorktreeStore.ts#L65-L76) replaces the worktrees Map on every snapshot past the `version <= get().version` check, even when every `WorktreeSnapshot` is value-equal to the prior version. The version short-circuit only catches stale messages, not no-op refreshes (user-triggered `worktree.refresh`, periodic resyncs).

Identity-bust on the Map cascades through every selector that reads it: `useWorktrees` rebuilds `normalizedMap`, `derivedMetaMap` reruns, the `SidebarContent` filter pipeline reruns. The existing [`snapshotsEqual` helper](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/store/createWorktreeStore.ts#L165-L211) is field-by-field cheap and is already the basis for the `applyUpdate` short-circuit at line 82 — `applySnapshot` should use it the same way.

## 3. `WorktreeFilterPopover` re-renders on every filter change

[`WorktreeFilterPopover.tsx:147-167`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Worktree/WorktreeFilterPopover.tsx#L147-L167) destructures `useWorktreeFilterStore()` with no selector and no `useShallow` — 16 fields pulled out of one call. Every filter change re-renders the entire popover including all five chip-grid sections.

This is the canonical Zustand 5 anti-pattern; the v5 docs flag the no-selector destructure as triggering React's "result of getSnapshot should be cached to avoid an infinite loop" guard. `useShallow` is already used elsewhere in the codebase ([`WorktreeCard.tsx:2`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Worktree/WorktreeCard.tsx#L2) imports it).

## 4. Stale-pin/collapse cleanup runs N persistence writes

[`SidebarContent.tsx:200-206`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Sidebar/SidebarContent.tsx#L200-L206) calls `unpinWorktree(id)` and `expandWorktree(id)` once per stale id via `forEach`. Each call fires its own `_projectStore.setState` and triggers its own persist write. The adjacent `setManualOrder(cleaned)` cleanup at [lines 209-216](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Sidebar/SidebarContent.tsx#L209-L216) is already batched into one call. The asymmetry should go.

## 5. Duplicate 200ms debounce in popover and search bar

[`WorktreeFilterPopover.tsx:179-198`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Worktree/WorktreeFilterPopover.tsx#L179-L198) and [`WorktreeSidebarSearchBar.tsx:47-67`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/components/Worktree/WorktreeSidebarSearchBar.tsx#L47-L67) each implement the same 200ms `setTimeout` + `clearTimeout` debounce around `setQuery`.

The codebase already exports a value-debounce hook from [`src/hooks/useDebounce.ts`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/hooks/useDebounce.ts) (re-exported from the hooks barrel). Both files can drop the inline implementation.

## 6. `buildSearchableText` is dead code

[`buildSearchableText` in `worktreeFilters.ts:75-86`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/lib/worktreeFilters.ts#L75-L86) has zero production callers — confirmed via grep across `src/`, `shared/`, and `electron/`. The only references are in [`worktreeFilters.test.ts`](https://github.com/daintreehq/daintree/blob/7b98b3171f44856a50834c41c09de84b0e4d5118/src/lib/__tests__/worktreeFilters.test.ts#L148-L200) where 11 test cases exercise the function. Drop the function and the corresponding `describe("buildSearchableText", …)` block.

gregpriday:
## Summary

- Replace five bare `localeCompare` calls across `worktreeFilters.ts` and `useWorktrees.ts` with a module-scoped `Intl.Collator` (`{ numeric: true, sensitivity: "base" }`) so `feature-2` sorts before `feature-10` as expected
- Add a value-equality guard to `applySnapshot` in `createWorktreeStore.ts` that skips rebuilding the worktrees `Map` when every snapshot compares equal to its prior entry — mirrors the existing guard on `applyUpdate`
- Add a `cleanupStaleWorktrees` facade action to `worktreeFilterStore` so `SidebarContent`'s stale cleanup becomes a single store write instead of N `forEach` `setState` calls; wrap the bare `useWorktreeFilterStore()` call in `WorktreeFilterPopover` with `useShallow` to stop unrelated facade-store changes from triggering re-renders

Resolves #7221

## Changes

- `src/lib/worktreeFilters.ts` — module-scoped collator replaces inline `localeCompare` calls
- `src/hooks/useWorktrees.ts` — same collator applied to the sort in `useWorktrees`
- `src/store/createWorktreeStore.ts` — snapshot equality guard on `applySnapshot`; includes `output` field in `lifecycleStatusEqual` so the guard stays correct if any future surface reads it
- `src/store/worktreeFilterStore.ts` — `cleanupStaleWorktrees` facade action
- `src/components/Sidebar/SidebarContent.tsx` — uses facade action; fixes pre-existing bug where the cleanup effect wiped persisted pins/collapses on the empty pre-hydration render
- `src/components/Worktree/WorktreeFilterPopover.tsx` — `useShallow` wrapping
- `src/lib/__tests__/worktreeFilters.test.ts` — deletes dead `buildSearchableText` export and its 11 tests; adds numeric sort assertion (`feature-2 < feature-10`)

## Testing

vitest: 146 tests pass. `tsc --noEmit` clean. Lint ratchet improved by 8 warnings. Production build passes.

### DIFF
--- a/src/components/Sidebar/SidebarContent.tsx
+++ b/src/components/Sidebar/SidebarContent.tsx
@@ -194,8 +194,11 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
 
   const setManualOrder = useWorktreeFilterStore((state) => state.setManualOrder);
 
-  // Clean up stale pinned and collapsed worktrees
+  // Clean up stale pinned and collapsed worktrees. Skip when worktrees is
+  // empty — that's the pre-hydration state, not a signal that every persisted
+  // pin/collapse is stale.
   useEffect(() => {
+    if (worktrees.length === 0) return;
     const existingIds = new Set(worktrees.map((w) => w.id));
     cleanupStaleWorktrees(existingIds);
   }, [worktrees, cleanupStaleWorktrees]);
--- a/src/components/Sidebar/SidebarContent.tsx
+++ b/src/components/Sidebar/SidebarContent.tsx
@@ -151,9 +151,7 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
   );
   const clearAllFilters = useWorktreeFilterStore((state) => state.clearAll);
   const hasActiveFilters = useWorktreeFilterStore((state) => state.hasActiveFilters);
-  const unpinWorktree = useWorktreeFilterStore((state) => state.unpinWorktree);
-  const collapsedWorktrees = useWorktreeFilterStore((state) => state.collapsedWorktrees);
-  const expandWorktree = useWorktreeFilterStore((state) => state.expandWorktree);
+  const cleanupStaleWorktrees = useWorktreeFilterStore((state) => state.cleanupStaleWorktrees);
   const setQuickStateFilter = useWorktreeFilterStore((state) => state.setQuickStateFilter);
 
   // Terminal store for derived metadata
@@ -199,11 +197,8 @@ function SidebarContent({ onOpenOverview }: SidebarContentProps) {
   // Clean up stale pinned and collapsed worktrees
   useEffect(() => {
     const existingIds = new Set(worktrees.map((w) => w.id));
-    const stalePins = pinnedWorktrees.filter((id) => !existingIds.has(id));
-    stalePins.forEach((id) => unpinWorktree(id));
-    const staleCollapsed = collapsedWorktrees.filter((id) => !existingIds.has(id));
-    staleCollapsed.forEach((id) => expandWorktree(id));
-  }, [worktrees, pinnedWorktrees, unpinWorktree, collapsedWorktrees, expandWorktree]);
+    cleanupStaleWorktrees(existingIds);
+  }, [worktrees, cleanupStaleWorktrees]);
 
   // Clean up stale manual order entries
   useEffect(() => {
--- a/src/components/Worktree/WorktreeFilterPopover.tsx
+++ b/src/components/Worktree/WorktreeFilterPopover.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useState, useRef } from "react";
 import { Filter, X, ChevronDown } from "lucide-react";
+import { useShallow } from "zustand/react/shallow";
 import { cn } from "@/lib/utils";
 import { Button } from "@/components/ui/button";
 import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover";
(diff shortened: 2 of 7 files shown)


gregpriday:
Closing per request.

PR #7344, opened 2026-05-07
fix(updater): disable electron-updater on all Windows builds

### LINKED ISSUE #7199: Disable electron-updater on Windows and gate updater UI for Store builds
With Daintree shipping through the Microsoft Store on Windows, the Store handles Windows updates end-to-end. We need to disable `electron-updater` on Windows entirely and hide the UI surfaces that assume it's available.

See [`docs/distribution/microsoft-store.md`](docs/distribution/microsoft-store.md) for the full migration context.

The existing Windows gate in `electron/services/AutoUpdaterService.ts` only fires for portable builds (it checks `process.env.PORTABLE_EXECUTABLE_FILE`). Widen it so all `process.platform === "win32"` runs return early — the Microsoft Store handles Windows update distribution, so there's no scenario where electron-updater should run on Windows after this lands.

A few UI surfaces also assume the updater is running. They should hide on Windows or display "Updates managed by Microsoft Store":

- `electron/menu.ts` — "Check for Updates…" menu items in the Daintree menu (macOS) and the Help menu (non-macOS)
- `src/components/Settings/GeneralTab.tsx` — the stable / nightly channel selector
- `src/hooks/useUpdateListener.tsx` and `src/hooks/useMainProcessToastListener.ts` — update-available toast actions

A small helper in `shared/config/` (e.g. `isWindowsStoreBuild()`) keeps the gate logic in one place rather than scattering `process.platform === "win32"` checks across the renderer.

macOS and Linux update flows stay unchanged — the channel-switching code, the R2 update feed, and the rest of the auto-updater plumbing remain meaningful for those platforms.

Sequencing: PR #7196 is currently in flight on `AutoUpdaterService.ts` for issue #7139. This work rebases on top of that PR after it merges to avoid conflicts on the gate block.

gregpriday:
## Summary
- Disables electron-updater on all Windows builds, not just portable builds. A shared `isElectronUpdaterSupported()` helper in `shared/config/platform.ts` gates everywhere.
- The update channel settings section in `GeneralTab` is hidden on Windows, and the updater hook skips event subscriptions when unsupported.
- Tests updated across the board for the new gate -- platform helper, the updater service, and the UI hook.

Resolves #7199

## Changes
- New `isElectronUpdaterSupported()` helper in `shared/config/platform.ts` with tests
- `AutoUpdaterService` uses the helper instead of checking `PORTABLE_EXECUTABLE_FILE`
- `preload.cts` exposes `update.isSupported` to the renderer
- Menu hides "Check for Updates" on Windows
- `GeneralTab` conditionally renders the Update Channel section and skips channel/last-check fetches
- `useUpdateListener` skips event subscriptions when `isSupported` is false

## Testing
- Unit tests for `isElectronUpdaterSupported` covering all platforms
- Updated `AutoUpdaterService` tests for the Windows gate
- New `useUpdateListener` tests verifying no subscriptions or notifications when unsupported

### DIFF
--- a/electron/menu.ts
+++ b/electron/menu.ts
@@ -454,7 +454,7 @@ export function createApplicationMenu(
             await shell.openExternal("https://github.com/daintreehq/daintree");
           },
         },
-        ...(process.platform !== "darwin" && app.isPackaged
+        ...(process.platform !== "darwin" && process.platform !== "win32" && app.isPackaged
           ? [
               { type: "separator" as const },
               {
--- a/electron/preload.cts
+++ b/electron/preload.cts
@@ -11,6 +11,7 @@
 
 import { contextBridge, ipcRenderer, webFrame, webUtils } from "electron";
 import { isTrustedRendererUrl } from "../shared/utils/trustedRenderer.js";
+import { isElectronUpdaterSupported } from "../shared/config/platform.js";
 import { isIpcEnvelope } from "../shared/types/ipc/errors.js";
 import { deserializeError } from "../shared/utils/ipcErrorSerialization.js";
 import type { AppErrorCode } from "../shared/types/appError.js";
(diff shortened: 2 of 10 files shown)


gregpriday:
Closing per request.

PR #7476, opened 2026-05-08
fix(terminal): stagger WebGL context allocation to prevent bulk-create flash

### LINKED ISSUE #7467: Agent terminals flash blank during bulk worktree creation
When bulk-creating worktrees, some agent terminals show a blank canvas for a few seconds before the content appears. The window is consistent — long enough to be jarring, short enough to look like a glitch.

The cause is Chromium's per-renderer-process limit on simultaneous WebGL contexts (around 16). When too many xterm WebGL contexts are constructed in the same animation frame, older contexts get evicted. The eviction is partially silent: `gl.isContextLost()` flips to `true` synchronously, but the `webglcontextlost` DOM event dispatches asynchronously. On top of that, the `WebglRenderer` in `@xterm/addon-webgl` 0.19.0 calls `e.preventDefault()` on `webglcontextlost` and waits 3000ms for `webglcontextrestored` before firing its own `onContextLoss` emitter. So a canvas can sit blank for at least three seconds before any consumer is notified.

```ts
// node_modules/@xterm/addon-webgl/src/WebglRenderer.ts:110-121
this._register(addDisposableListener(this._canvas, 'webglcontextlost', (e) => {
  console.log('webglcontextlost event received');
  e.preventDefault();
  this._contextRestorationTimeout = setTimeout(() => {
    this._contextRestorationTimeout = undefined;
    console.warn('webgl context not restored; firing onContextLoss');
    this._onContextLoss.fire(e);
  }, 3000 /* ms */);
}));
```

`TerminalWebGLManager` already pools addons with an LRU cap of 12 and trips a circuit breaker after 3 `onContextLoss` events in 60s. The breaker is mistuned for burst eviction: each event is individually delayed 3000ms by the timer above, so a tight burst that evicts five contexts trickles in over 3-6s and may not coalesce into a trip. The 12-context cap also doesn't account for non-terminal WebGL consumers in the renderer, so the reserved 4-context headroom isn't actually enforced.

```ts
// src/services/terminal/TerminalWebGLManager.ts:13-24
// Chromium caps active WebGL contexts at 16 per renderer process.
// Reserve 4 slots for potential non-terminal WebGL consumers...
private static _maxContexts = 12;

// Circuit breaker: if N genuine context-loss events occur within W ms,
// disable WebGL for the rest of the session...
private static readonly LOSS_THRESHOLD = 3;
private static readonly LOSS_WINDOW_MS = 60_000;
```

The proximate trigger is bulk creation. `BulkCreateWorktreeDialog` uses `p-queue` with `QUEUE_CONCURRENCY=3`, and each worktree's `spawnPanelsFromRecipe` spawns its agent panels sequentially in a `for ... of` loop. React batches the resulting attach landings into one or two animation frames, so up to ~9 `attach()` calls fire close together — and each one calls `webGLManager.ensureContext` synchronously inside `TerminalInstanceService.attach`:

```ts
// src/services/terminal/TerminalInstanceService.ts:895-906
if (!managed.isOpened) {
  ...
  managed.terminal.open(managed.hostElement);
  managed.isOpened = true;
  if (managed.runtimeAgentId && isWebGLEligibleTier(managed.lastAppliedTier)) {
    this.webGLManager.ensureContext(id, managed);
  }
}
```

### Reference files

- [`src/services/terminal/TerminalWebGLManager.ts`](https://github.com/daintreehq/daintree/blob/c39aced0fc457542f7a5eb4339e9be9ce52cbf6a/src/services/terminal/TerminalWebGLManager.ts)
- [`src/services/terminal/TerminalInstanceService.ts#L895-L906`](https://github.com/daintreehq/daintree/blob/c39aced0fc457542f7a5eb4339e9be9ce52cbf6a/src/services/terminal/TerminalInstanceService.ts#L895-L906)
- [`src/components/Worktree/panelSpawning.ts#L44`](https://github.com/daintreehq/daintree/blob/c39aced0fc457542f7a5eb4339e9be9ce52cbf6a/src/components/Worktree/panelSpawning.ts#L44)
- [`src/components/GitHub/bulkCreateUtils.ts#L16`](https://github.com/daintreehq/daintree/blob/c39aced0fc457542f7a5eb4339e9be9ce52cbf6a/src/components/GitHub/bulkCreateUtils.ts#L16)
- `node_modules/@xterm/addon-webgl/src/WebglRenderer.ts:110-121` (vendored, see snippet above)

### Related work

- #6163 introduced the circuit breaker
- #6801 and #6802 fixed adjacent WebGL flicker on tier transitions and tab/project switches
- #5836 added off-screen renderer disposal
- #4961 fixed content disappearing during renderer switching

### Reproducing

Open the bulk-create-worktree dialog with 3-5 issues selected, where the worktree recipe spawns 1-2 agent panels per worktree. Watch the new agent terminals — a subset land with a blank xterm canvas that fills in after several seconds. More likely to reproduce when the renderer already has other live WebGL contexts (browser previews, GPU-backed UI surfaces).

### Constraints

- The fallback path (DOM renderer, since xterm 6.0 removed the canvas addon) must keep working.
- The existing 3-loss/60s circuit breaker is load-bearing for persistent GPU faults (M-series Macs on external displays at fractional scaling) and shouldn't be removed.
- Whatever the fix is, focused/visible terminals should still get the WebGL renderer when capacity allows.

gregpriday:
## Summary

- Bulk-creating worktrees triggers near-simultaneous `ensureContext` calls that can exceed Chromium's 16-context-per-renderer cap synchronously, evicting older contexts and leaving agent terminals blank for ~3 seconds before the upstream `WebglRenderer` fires `onContextLoss`.
- `TerminalWebGLManager` now queues `ensureContext` requests and drains them in batches of 3 per `setTimeout(0)` macrotask, so a burst of up to 9 attach calls can never overflow the cap in a single event-loop turn.
- The time-based loss cluster is now anchored on the first event in a burst rather than rolled forward on each new event. The upstream addon delays `onContextLoss` by 3000ms per eviction, so a burst that evicts five contexts trickles callbacks in over 3-6s — rolling the anchor caused them to coalesce outside a single window and miss the circuit breaker. Anchoring on start preserves breaker sensitivity to sustained near-threshold faults.

Resolves #7467

## Changes

- `src/services/terminal/TerminalWebGLManager.ts` — async allocation queue with batch draining; fixed cluster anchor logic
- `src/services/terminal/__tests__/TerminalWebGLManager.test.ts` — 401 lines of new targeted tests covering queue batching, drain ordering, cluster anchoring, and breaker behaviour under burst eviction
- `src/services/terminal/__tests__/TerminalInstanceService.adversarial.test.ts`, `webglLease.test.ts`, `webglVisibility.test.ts` — 6-8 additional assertions per suite covering the staggered allocation path

## Testing

87 targeted tests added across `TerminalWebGLManager` and the three `TerminalInstanceService` suites (adversarial, webglLease, webglVisibility). Full `src/` suite: 10,641 tests pass. Typecheck, lint, format, and build all clean.

### DIFF
--- a/src/services/terminal/TerminalWebGLManager.ts
+++ b/src/services/terminal/TerminalWebGLManager.ts
@@ -42,7 +42,8 @@ export class TerminalWebGLManager {
   }
 
   static setMaxContexts(n: number): void {
-    TerminalWebGLManager._maxContexts = Math.max(1, n);
+    if (!Number.isFinite(n) || n < 1) return;
+    TerminalWebGLManager._maxContexts = Math.floor(n);
   }
 
   private pool = new Map<string, WebGLEntry>();
@@ -56,10 +57,13 @@ export class TerminalWebGLManager {
   // event-loop ticks (see CONTEXTS_PER_DRAIN). Insertion order is the drain order.
   private pending = new Map<string, ManagedTerminal>();
   private drainScheduled = false;
-  // Timestamp of the most recent recorded loss. Used together with
-  // LOSS_CLUSTER_MS to collapse clustered upstream loss events that all
-  // belong to the same burst-overflow wave.
-  private lastLossAt: number | null = null;
+  // Anchor for the active loss-cluster window. Losses arriving within
+  // LOSS_CLUSTER_MS of the cluster start collapse to a single timestamp;
+  // the next loss outside that window opens a new cluster. Anchoring on
+  // start (not rolling on each loss) ensures sustained near-threshold
+  // faults still trip the breaker — a 400ms-interval fault closes one
+  // cluster every ~500ms and accumulates timestamps normally.
+  private clusterStartAt: number | null = null;
 
   setHardwareAvailable(available: boolean): void {
     this.hardwareAvailable = available;
@@ -215,16 +219,18 @@ export class TerminalWebGLManager {
 
     // Cluster collapse: when bulk-creation overflows Chromium's 16-context
     // cap, the upstream addon's 3000ms timer fires for every evicted addon
-    // at once. Treat losses arriving within LOSS_CLUSTER_MS of the previous
-    // one as the same wave, recording at most one timestamp per cluster.
+    // at once. Collapse all losses arriving within LOSS_CLUSTER_MS of the
+    // cluster start to a single timestamp; the next loss outside that
+    // window opens a new cluster. The anchor is the cluster start (not the
+    // last loss) so sustained near-threshold faults can't extend the window
+    // indefinitely.
     if (
-      this.lastLossAt !== null &&
-      now - this.lastLossAt < TerminalWebGLManager.LOSS_CLUSTER_MS
+      this.clusterStartAt !== null &&
+      now - this.clusterStartAt < TerminalWebGLManager.LOSS_CLUSTER_MS
     ) {
-      this.lastLossAt = now;
       return;
     }
-    this.lastLossAt = now;
+    this.clusterStartAt = now;
 
     this.lossTimestamps = this.lossTimestamps.filter(
       (t) => now - t < TerminalWebGLManager.LOSS_WINDOW_MS
--- a/src/services/terminal/TerminalWebGLManager.ts
+++ b/src/services/terminal/TerminalWebGLManager.ts
@@ -23,6 +23,20 @@ export class TerminalWebGLManager {
   private static readonly LOSS_THRESHOLD = 3;
   private static readonly LOSS_WINDOW_MS = 60_000;
 
+  // When the bulk-creation path overflows Chromium's 16-context cap, the
+  // upstream addon's 3000ms timer fires for every evicted addon at once —
+  // so loss events arrive in a tight cluster (~ms apart). Collapse a cluster
+  // of clustered losses into a single timestamp so the wave doesn't itself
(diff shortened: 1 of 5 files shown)


gregpriday:
Tried to auto-rebase against `develop` after #7474 (lazy-load `@xterm/addon-webgl`) landed, but the conflict in `TerminalWebGLManager.ts` is structural rather than mechanical:

- #7474 reshaped `ensureContext` around a `pendingEnsures` / `flushPendingEnsures` flow that replays queued attaches once `loadWebglAddon()` resolves.
- This PR reshapes the same path around a `pending` queue that drains `CONTEXTS_PER_DRAIN` per `setTimeout(0)` macrotask.

Composing the two needs a design call:
1. Do load-resolve replays go through the staggered drain (correct under burst but adds a macrotask hop), or fire synchronously (defeats the stagger when N items queued during load)?
2. The auto-merged test file mixes `vi.useFakeTimers()` (this PR's drain tests) with `flushDynamicImport()` (#7474's real-timer lazy-load tests) — those are mutually incompatible without further surgery.

Punted to the author so the integration semantics are picked deliberately. Once rebased, /prs:merge will pick it back up automatically.

gregpriday:
Closing — structural conflict with #7474 needs design call (see prior comment). Will re-open / re-do after rebase.

PR #7479, opened 2026-05-08
fix(actions): ignore stale focused panel when duplicating

### LINKED ISSUE #7477: Cmd+T duplicates a trashed terminal that's still marked as focused
Cmd+T (Duplicate Panel) duplicates a panel that's just been moved to the trash dock, instead of doing nothing. Repro: focus a Claude Code terminal, close it (it goes to trash), press Cmd+T — a duplicate of the trashed terminal gets spawned. Expected: Cmd+T does nothing, since there's no live "selected" terminal to duplicate.

Two pieces fit together to cause it:

`trashPanel` flips the panel's `location` to `"trash"` but never touches `state.focusedId`. The just-trashed panel keeps the focus pointer pointing at it.

[`src/store/slices/panelRegistry/trashActions.ts#L69-L116`](https://github.com/daintreehq/daintree/blob/5488e80b34677ebb4bcfb4a483d5868da5bebb70/src/store/slices/panelRegistry/trashActions.ts#L69-L116)

The `terminal.duplicate` handler then reads `state.focusedId` without checking whether that panel is in the trash. It computes a `nonTrashed` list right above this line and never consults it for the focused-id case:

```ts
const nonTrashed = state.panelIds
  .map((id) => state.panelsById[id])
  .filter((t) => t && t.location !== "trash");
const targetId =
  terminalId ?? state.focusedId ?? (nonTrashed.length === 1 ? nonTrashed[0]!.id : undefined);
```

[`src/services/actions/definitions/terminalSpawnActions.ts#L51-L77`](https://github.com/daintreehq/daintree/blob/5488e80b34677ebb4bcfb4a483d5868da5bebb70/src/services/actions/definitions/terminalSpawnActions.ts#L51-L77)

If `targetId` resolves to a trashed panel, the handler falls through to the duplicate path. The fallback at L78–108 (which would create a fresh terminal when nothing's available to duplicate) is never reached.

## Followup observation

The reporter noted that pressing Cmd+T a second time does **not** create another Claude Code terminal — i.e. the behavior on press #2 differs from press #1. The duplicate action sets `location` to `"grid"` when the source is trashed (L65–66), so the first press should produce a grid panel that then becomes focused, and the second press should duplicate that one. Worth verifying during fix that the new panel actually surfaces in the grid and that `buildPanelDuplicateOptions` isn't carrying trash-related state into the copy.

## Desired behavior

When the focused panel is in the trash, Cmd+T should treat focus as empty — either do nothing, or fall through to the existing "no panels" path that creates a fresh terminal from `lastClosedConfig` or defaults.

gregpriday:
## Summary

- `terminal.duplicate` was reading `state.focusedId` without checking whether that panel was still live. When a trash op left `focusedId` pointing at a trashed panel (IPC race or just normal close-then-Cmd+T), the action duplicated the trashed panel rather than falling through to the fresh-terminal path.
- Fix: derive `liveFocusedId` by checking `panelsById[focusedId]?.location !== "trash"` before using it as the target. The explicit `terminalId` path (context-menu / panel-header dispatches) is unchanged.
- 5 new adversarial tests: stale focus + 0 live panels, stale focus + 1 live panel, dangling `focusedId` (id missing from `panelsById`), stale focus + `lastClosedConfig` fallthrough, stale focus + multiple live panels (no-op).

Resolves #7477

## Changes

- `src/services/actions/definitions/terminalSpawnActions.ts` — guard `focusedId` against trashed location before resolving `targetId`
- `src/services/actions/__tests__/actionDefinitions.adversarial.test.ts` — 5 new cases covering the above edge conditions

## Testing

- 45/45 unit tests pass on the adversarial spec
- Typecheck, lint, format, and build all pass

### DIFF
--- a/src/services/actions/__tests__/actionDefinitions.adversarial.test.ts
+++ b/src/services/actions/__tests__/actionDefinitions.adversarial.test.ts
@@ -775,6 +775,54 @@ describe("terminal action hardening", () => {
     expect(addPanel).toHaveBeenCalledWith(expect.objectContaining({ title: "Live One (copy)" }));
   });
 
+  it("ignores dangling focusedId pointing at a missing panel and falls back to the lone live terminal", async () => {
+    const actions = buildRegistry(registerTerminalActions);
+    const duplicate = actions.get("terminal.duplicate")!();
+    const addPanel = vi.fn().mockResolvedValue("copy-id");
+
+    usePanelStore.setState({
+      panelsById: { "term-live": createTerminal({ id: "term-live", title: "Live One" }) },
+      panelIds: ["term-live"],
+      focusedId: "gone",
+      addPanel,
+    } as never);
+
+    await duplicate.run(undefined, {} as never);
+
+    expect(addPanel).toHaveBeenCalledWith(expect.objectContaining({ title: "Live One (copy)" }));
+  });
+
+  it("falls through to lastClosedConfig snapshot when focusedId is stale-trashed and no live panels exist", async () => {
+    const actions = buildRegistry(registerTerminalActions);
+    const duplicate = actions.get("terminal.duplicate")!();
+    const addPanel = vi.fn().mockResolvedValue("new-id");
+
+    usePanelStore.setState({
+      panelsById: {
+        "term-trash": createTerminal({ id: "term-trash", location: "trash" }),
+      },
+      panelIds: ["term-trash"],
+      focusedId: "term-trash",
+      lastClosedConfig: {
+        kind: "terminal",
+        type: "terminal",
+        cwd: "/projects/app",
+        worktreeId: "wt-1",
+      },
+      addPanel,
+    } as never);
+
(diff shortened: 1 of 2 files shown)


gregpriday:
CI `test` job is red, but the failure is in `src/components/Worktree/WorktreeCard/__tests__/WorktreeTerminalSection.test.tsx` (15 failed tests, all `TypeError: children is not a function` in `SortableWorktreeTerminal`). That test file is unrelated to this PR — the diff here only touches `src/services/actions/definitions/terminalSpawnActions.ts` and `src/services/actions/__tests__/actionDefinitions.adversarial.test.ts`.

Confirmed pre-existing on `develop`: the most recent develop CI run (25570086167, sha `60c77790a`) has the same 15 failures in the same file. Needs a separate fix.

The new adversarial tests in this PR (45 tests, including 5 new) pass locally and would have run as part of the same `vitest` invocation.

PR #7620, opened 2026-05-09
fix(e2e): guard against null return from getCurrentProject in switchToProject

### LINKED ISSUE #7619: fix(e2e): null pointer crash in switchToProject when no project is loaded
## Bug

In `e2e/full/core-project-switch-race.spec.ts`, the helper `getCurrentProject()` is typed as `Promise<ProjectInfo>` but `electron.project.getCurrent()` returns `null` when no project is loaded (welcome screen active or project view not yet initialised).

The `switchToProject` helper immediately dereferences `current.name`:
```typescript
const current = await getCurrentProject(page);
if (current.name === projectName) return page; // TypeError when current is null
```

This crashes the entire `beforeAll` for the race-conditions suite, blocking all tests.

## Error

```
TypeError: Cannot read properties of null (reading 'name')
at e2e/full/core-project-switch-race.spec.ts:53
```

## Platform

All platforms (Ubuntu, macOS, Windows) — the null return is not platform-specific.

## Fix

```typescript
// Return type
async function getCurrentProject(page): Promise<ProjectInfo | null>

// Guard
if (current?.name === projectName) return page;
```

## Verification

```bash
npx playwright test e2e/full/core-project-switch-race.spec.ts
```

The `beforeAll` should no longer crash with a TypeError when the project switcher returns null.

## Seen in

`.last-run.json` failure: `e2e/full/core-project-switch-race.spec.ts >> Core: Project Switch Race Conditions >> delayed spawn assigns terminal to originating project`

dev-en-m:
## Summary

- `getCurrentProject()` was typed as `Promise<ProjectInfo>` but the underlying IPC call returns `null` when no project is loaded
- `switchToProject` read `current.name` without a null check, crashing `beforeAll` and blocking all tests in the race-conditions suite
- Fix: widen return type to `ProjectInfo | null`, add optional chaining (`current?.name`)

Closes #7619

## Test plan

- [ ] Run `npx playwright test e2e/full/core-project-switch-race.spec.ts` and confirm no TypeError in beforeAll
- [ ] Confirm `switchToProject` proceeds to the palette flow when `current` is null (no project loaded)

### DIFF
--- a/e2e/full/core-project-switch-race.spec.ts
+++ b/e2e/full/core-project-switch-race.spec.ts
@@ -127,6 +127,7 @@ test.describe.serial("Core: Project Switch Race Conditions", () => {
 
     // Capture Project A's ID
     const projectA = await getCurrentProject(ctx.window);
+    expect(projectA).not.toBeNull();
 
     // Open a terminal in Project A to confirm normal flow works
     await openTerminal(ctx.window);
@@ -164,7 +165,7 @@ test.describe.serial("Core: Project Switch Race Conditions", () => {
     const withProject = activeTerminals.filter((t: TerminalInfo) => t.projectId !== undefined);
     expect(withProject.length).toBeGreaterThanOrEqual(1);
     for (const t of withProject) {
-      expect(t.projectId).toBe(projectA.id);
+      expect(t.projectId).toBe(projectA!.id);
     }
   });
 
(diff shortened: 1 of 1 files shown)


gregpriday (approved):
Good catch on the null check gap in `switchToProject` — the widened `ProjectInfo | null` return type was a real crash path. I added `expect(projectA).not.toBeNull()` guards and a non-null assertion (`projectA!.id`) to the spec so the updated return type can't cause a `TypeError` at the assertion loops either.

Thanks for the contribution. Approving.

PR #7625, opened 2026-05-09
fix(e2e): gate refreshActiveWindow on sidebar DOM attachment

### LINKED ISSUE #7624: fix(e2e): refreshActiveWindow returns before currentProject loads — systemic [data-worktree-branch] failures
## Summary

`refreshActiveWindow` in `e2e/helpers/launch.ts` returns before the project is
fully loaded into the view, causing widespread test failures across `e2e/core/`
and `e2e/full/`.

## Root cause

`AppLayout` only mounts `<Sidebar>` when `currentProject != null`:

```tsx
// src/components/Layout/AppLayout.tsx:448
{currentProject != null && (
  <Sidebar ... isVisible={showSidebar} />
)}
```

`currentProject` is set asynchronously after the project view's React app loads
and the IPC `loadCurrentProject` call resolves. Between mount and resolution the
sidebar — and all `[data-worktree-branch]` cards inside it — are absent from the
DOM.

`refreshActiveWindow` waited for those worktree selectors and fell back to:

```ts
.locator("text=Loading worktrees...")
  .waitFor({ state: "hidden", timeout: refreshTimeout })
```

The `Skeleton` component renders that label as `<span class="sr-only">Loading
worktrees</span>` (no "...", screen-reader-only). The locator never matched, so
`waitFor({ state: "hidden" })` resolved immediately — causing `refreshActiveWindow`
to return while `currentProject` was still `null`.

## Impact

- **`e2e/core/core-persistence.spec.ts`** — release-gating; project switcher shows
  "Open project" after `openAndOnboardProject`
- **`e2e/core/core-worktree-lifecycle.spec.ts`** — release-gating; `[data-worktree-branch]`
  not visible after onboarding
- Dozens of `e2e/full/` tests with the same root failure

## Fix

Add an `aside[aria-label="Sidebar"]` attached-wait after the Toggle Sidebar check.
Since `<Sidebar>` is only mounted when `currentProject != null`, this gates
`refreshActiveWindow` on the project actually being loaded. Also remove the broken
`text=Loading worktrees...` fallback (replace with no-op catch).

## Platform

Linux (Ubuntu) — Ubuntu x86_64 CI runner. Same logic runs on macOS and Windows.

## Verification

- Run `npx playwright test e2e/core/core-persistence.spec.ts` — "previously onboarded project appears after restart" should pass
- Run `npx playwright test e2e/core/core-worktree-lifecycle.spec.ts` — "main worktree card is visible and selected" should pass
- Run `npx playwright test e2e/full/core-terminal-panels.spec.ts` — "worktree dashboard appears with at least one card" should pass

dev-en-m:
Fixes #7624

## Summary

- `AppLayout` only mounts `<Sidebar>` when `currentProject != null`; without a
  gate, `refreshActiveWindow` returned before the IPC `loadCurrentProject` call
  resolved
- The previous `text=Loading worktrees...` fallback was a no-op: `Skeleton`
  renders that text as sr-only without "..." so the locator never matched and
  `waitFor({ state: "hidden" })` resolved immediately
- Add `aside[aria-label="Sidebar"]` attached-wait between the Toggle Sidebar
  check and the worktree poll — sidebar mounts only when `currentProject != null`,
  so this reliably blocks until the project is loaded
- Remove the broken fallback (replace with no-op catch)

## Affected tests (sample)

- `e2e/core/core-persistence.spec.ts` — project switcher showed "Open project"
- `e2e/core/core-worktree-lifecycle.spec.ts` — `[data-worktree-branch]` not found
- `e2e/full/core-terminal-panels.spec.ts` — worktree dashboard not appearing
- Dozens more `e2e/full/` specs with the same root failure

## Test plan

- [ ] `npx playwright test e2e/core/core-persistence.spec.ts` — "previously onboarded project appears after restart" passes
- [ ] `npx playwright test e2e/core/core-worktree-lifecycle.spec.ts` — "main worktree card is visible and selected" passes
- [ ] `npx playwright test e2e/full/core-terminal-panels.spec.ts` — "worktree dashboard appears with at least one card" passes

### DIFF
--- a/e2e/helpers/launch.ts
+++ b/e2e/helpers/launch.ts
@@ -407,12 +407,9 @@ export async function refreshActiveWindow(app: ElectronApplication, oldPage?: Pa
     .locator('[aria-label="Toggle Sidebar"]')
     .waitFor({ state: "visible", timeout: refreshTimeout });
 
-  // Wait for the sidebar element to appear in the DOM. AppLayout only mounts
-  // <Sidebar> when currentProject != null, so this guards against returning
-  // before the IPC loadCurrentProject call resolves. Without this gate the
-  // worktree wait below times out (no sidebar = no worktree cards), and the
-  // previous "Loading worktrees..." fallback was a no-op because that text is
-  // sr-only in the Skeleton component and was never found by the locator.
+  // <Sidebar> mounts only after currentProject hydrates — best-effort gate
+  // before the worktree poll. Use attached not visible: a gesture-hidden
+  // sidebar is aria-hidden/inert but still in the DOM.
   await newWindow
     .locator('aside[aria-label="Sidebar"]')
     .waitFor({ state: "attached", timeout: refreshTimeout })
--- a/e2e/helpers/launch.ts
+++ b/e2e/helpers/launch.ts
@@ -407,25 +407,28 @@ export async function refreshActiveWindow(app: ElectronApplication, oldPage?: Pa
     .locator('[aria-label="Toggle Sidebar"]')
     .waitFor({ state: "visible", timeout: refreshTimeout });
 
+  // Wait for the sidebar element to appear in the DOM. AppLayout only mounts
+  // <Sidebar> when currentProject != null, so this guards against returning
+  // before the IPC loadCurrentProject call resolves. Without this gate the
+  // worktree wait below times out (no sidebar = no worktree cards), and the
+  // previous "Loading worktrees..." fallback was a no-op because that text is
+  // sr-only in the Skeleton component and was never found by the locator.
+  await newWindow
+    .locator('aside[aria-label="Sidebar"]')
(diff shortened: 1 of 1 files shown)


gregpriday (approved):
Thanks for tracking this down. Gating on DOM attachment rather than visibility is the right call — it's a subtle distinction but it matters for timing.

I tightened up the guard slightly and added a comment explaining the `attached` vs `visible` difference so it's clear why we're not just checking visibility. Everything else looks good.

gregpriday (approved):
Good find. The missing gate on sidebar DOM attachment was a subtle one -- `refreshActiveWindow` returning early before the project loaded is exactly the kind of timing issue that's hard to spot without knowing how `AppLayout` conditionally mounts `<Sidebar>`.

I trimmed the comment block per the minimal-comments rule, rebased onto `develop`, and everything looks solid. Thanks for digging this out.

PR #7649, opened 2026-05-10
fix(e2e,resource): terminal selection lost in context menu test + provision no-op skips connect command sync

dev-en-m:
## Summary

Two nightly E2E failures fixed (Ubuntu x86-64, full suite run):

**1. `core-terminal-context-menu` — Copy always disabled after selectAll**

- `openTerminalContextMenu` sends `page.keyboard.press("Escape")` at the start of each attempt before right-clicking
- xterm.js 6.0 calls `clearSelection()` synchronously in `_keyDown` for any key that isn't browser-handled (Escape is forwarded to the PTY — it's not browser-handled)
- This destroyed any prior `selectAll()` call before the `contextmenu` event could read the selection via `terminal.getSelection()`
- Fix: add `preserveSelection` option to `openTerminalContextMenu` that skips the Escape; the "Copy is enabled after selecting text" test passes `{ preserveSelection: true }`

**2. `core-worktree-resource-lifecycle` — `{{branch}}`/`{{worktree_path}}` substitution ignored**

- When re-provisioning a resource that is already in a ready state, `ResourceActionExecutor.execute` correctly calls `applyResourceConfigToMonitor` (which updates `monitor.resourceConnectCommand` with the new substituted command from the on-disk config), but then returns early without calling `this.ctx.emitUpdate(monitor)`
- The renderer's Zustand store never received the updated connect command; "Connect to Resource" used the stale original command stored from the previous provision
- Fix: call `this.ctx.emitUpdate(monitor)` before the early return so the renderer always has the current connect command after any provision call

## Test plan

- [x] `npx playwright test e2e/full/terminal/core-terminal-context-menu.spec.ts --project=full-terminal` → 10 passed (Ubuntu)
- [x] `npx playwright test e2e/full/worktree/core-worktree-resource-lifecycle.spec.ts --project=full-worktree` → 14 passed (Ubuntu)
- [x] `npm run typecheck` → passes

### DIFF
--- a/e2e/full/terminal/core-terminal-context-menu.spec.ts
+++ b/e2e/full/terminal/core-terminal-context-menu.spec.ts
@@ -134,7 +134,7 @@ test.describe.serial("Core: Terminal Context Menu", () => {
       const { window } = ctx;
 
       await selectAllTerminalText(panel);
-      await openTerminalContextMenu(panel);
+      await openTerminalContextMenu(panel, { preserveSelection: true });
 
       const copyItem = window.getByRole("menuitem", { name: "Copy" });
       await expect(copyItem).toBeVisible({ timeout: T_SHORT });
--- a/e2e/helpers/terminal.ts
+++ b/e2e/helpers/terminal.ts
@@ -320,7 +320,10 @@ async function dispatchXtermContextMenu(xterm: Locator): Promise<boolean> {
   });
 }
 
-export async function openTerminalContextMenu(panelLocator: Locator): Promise<void> {
+export async function openTerminalContextMenu(
+  panelLocator: Locator,
+  { preserveSelection = false }: { preserveSelection?: boolean } = {}
+): Promise<void> {
   const page = panelLocator.page();
   const menu = page.locator(SEL.contextMenu.content);
   const xterm = panelLocator.locator(SEL.terminal.xtermRows);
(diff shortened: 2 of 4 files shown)


gregpriday (approved):
Good catch on both of these — glad you tracked them down.

I added two things on top of the original fix:

- A regression test asserting that `emitUpdate` fires in the no-op provision path and that the emitted snapshot carries the substituted `resourceConnectCommand`. The original fix corrected the behavior but left no coverage for it.
- A guard that re-fetches the monitor after the `loadConfig` await before calling `emitUpdate`. Without it, a worktree removed mid-provision would cause a ghost resurrection — the original monitor reference is still in scope but the monitor is gone.

Happy with how it all came together. Nice find.

gregpriday:
Two additional commits pushed since the original PR:

**`test(resource)`: cover `emitUpdate` broadcast in no-op provision path**
Extends the existing no-op test with an assertion that the worktree-update event fires, plus a new test asserting the emitted snapshot carries the substituted `resourceConnectCommand`. The original fix corrected the behavior but there was no coverage ensuring it stays correct.

**`fix(resource)`: guard no-op provision `emitUpdate` against removed monitor**
Re-fetches the monitor after the `loadConfig` await before calling `emitUpdate`. If a worktree is removed mid-provision, the stale reference would be resurrected into the live state. Includes a regression test covering that race.

Issue #7668, opened 2026-05-11
Extract shared toolbar button metadata to fix Plug/Puzzle drift


gregpriday:
The agent-tray entry renders with `Plug` in the toolbar but `Puzzle` in the customization settings tab. Same item, two icons. The cause is two parallel metadata maps that have drifted:

- `OVERFLOW_MENU_META` in `src/components/Layout/Toolbar.tsx:118-134`
- `BUTTON_METADATA` in `src/components/Settings/ToolbarSettingsTab.tsx:65-131`

Both declare label + icon for the same `ToolbarButtonId` set. They've diverged on `agent-tray`, and there's no structural protection against the same thing happening to the next item.

## Fix

Extract a single `TOOLBAR_BUTTON_METADATA` map to a new shared module (e.g. `src/components/Layout/toolbarButtonMetadata.ts`) with one record per `ToolbarButtonId`:

```ts
type ToolbarButtonMetadata = {
  label: string;
  icon: LucideIcon;
  description: string; // long-form for the customization UI
};
```

`OVERFLOW_MENU_META` in `Toolbar.tsx` and `BUTTON_METADATA` in `ToolbarSettingsTab.tsx` both read from this single source.

## Scope

In scope:

- The shared metadata file
- Migrating both call-sites to consume it
- Picking one canonical icon for `agent-tray`. Recommend `Plug`, since `AgentTrayButton.tsx:613` already renders `Plug` for the live button.
- A regression test that asserts the toolbar render and the settings-tab render for `agent-tray` (and any other built-in) come from the same source

Out of scope:

- The full Command/Placement/Widget registry redesign (tracked in #7666)
- Pin-as-universal UX changes
- `agentSettingsStore` / `toolbarPreferencesStore` consolidation
- Drag-region placeholder refactor

## Constraints

- Agent metadata stays derived from `BUILT_IN_AGENT_IDS` + `getAgentConfig` exactly as today.
- Plugin-contributed buttons (`usePluginToolbarButtons` / `shared/config/toolbarButtonRegistry.ts`) stay on their current path. This change doesn't touch them.
- No store migration required. This only affects rendering, not persisted state.

Issue #7673, opened 2026-05-11
Agent pinned state freezes at first availability snapshot


gregpriday:
`normalizeAgentSelection` in [`src/store/agentSettingsStore.ts:30-57`](https://github.com/daintreehq/daintree/blob/fcd8e47991d8a3bf349e7eadb83f19593df08698/src/store/agentSettingsStore.ts#L30-L57) seeds the persisted `pinned` boolean on a per-agent entry from a snapshot of CLI availability the first time `hasRealData` lands. The value is never re-derived after availability changes. The persisted preference freezes at first observation, and subsequent installs and uninstalls don't move it.

The render gate in `src/components/Layout/Toolbar.tsx` masks the symptom for the toolbar itself. Agent buttons are drawn through `isAgentInstalled(...) && isAgentPinned(...)`, so uninstalled-but-still-pinned rows are hidden at draw time. The settings surface doesn't gate the same way. [`isEffectivelyVisible`](https://github.com/daintreehq/daintree/blob/fcd8e47991d8a3bf349e7eadb83f19593df08698/src/components/Settings/ToolbarSettingsTab.tsx#L56-L63) in `ToolbarSettingsTab.tsx` reads only `pinned`, with no availability check, and shows uninstalled agents as enabled. The two surfaces disagree.

This is the lazy-seed-from-async-default pattern. Mainstream extensible platforms route around it the same way: persist explicit user intent only, derive visibility per render.

- Chrome MV3: `chrome.action.getUserSettings()` returns the user's intent as `isOnToolbar`. Toolbar visibility is derived per render.
- VS Code: extension enablement is stored separately from system availability (`engines.vscode`, activation conditions). The persisted preference is never rewritten by an availability flip.
- JetBrains: only negative intent is persisted in `disabled_plugins.txt`. A missing plugin binary just fails to load silently.
- OpenFeature: visibility is evaluated through a `FeatureProvider` with an `EvaluationContext` on each call. The result is transient.

## Why a straight fix won't do

The doc comment in `normalizeAgentSelection` references #5111 and #5158, the 0.7.x upgrade contract. An upgrader whose persisted entry exists without `pinned` must still see their installed agents in the toolbar after the fix, otherwise the toolbar collapses on relaunch and the AgentSetupWizard auto-opens. Removing the lazy seed without a migration breaks that contract.

## Proposed direction

[...]

PR #7738, opened 2026-05-11
feat(toolbar): unify toolbar customization with the agent tray

### LINKED ISSUE #7666: Unify toolbar customization with the agent tray
The toolbar customization settings and the agent tray are two parallel systems describing overlapping things, and the seams are starting to show.

## Concrete problems

**Icon mismatch.** The agent tray renders with a `Plug` icon in the toolbar, but the customization settings render the same row with a `Puzzle` icon. Same item, two identities.

- Toolbar rendering: [`Toolbar.tsx:147`](https://github.com/daintreehq/daintree/blob/88e295a07b833e7b56aa8b949b0f05334e344fae/src/components/Layout/Toolbar.tsx#L147) — `Plug`
- Settings rendering: [`ToolbarSettingsTab.tsx:68`](https://github.com/daintreehq/daintree/blob/88e295a07b833e7b56aa8b949b0f05334e344fae/src/components/Settings/ToolbarSettingsTab.tsx#L68) — `Puzzle`

That's the symptom, not the disease. There are two registries of toolbar items with no single source of truth, so they drift.

**Conceptual overlap.** The settings tab lets users toggle items in/out of the toolbar, but for agents the tray already does this — pinning an agent surfaces it in the toolbar. The two paths are persisted separately:

- Non-agent button visibility: `useToolbarPreferencesStore.layout.hiddenButtons`
- Agent visibility: `useAgentSettingsStore.agents[id].pinned`

The settings tab papers over the split via `isEffectivelyVisible()` ([`ToolbarSettingsTab.tsx:56-63`](https://github.com/daintreehq/daintree/blob/88e295a07b833e7b56aa8b949b0f05334e344fae/src/components/Settings/ToolbarSettingsTab.tsx#L56-L63)), and the toolbar duplicates the same logic at render time ([`Toolbar.tsx:655-685`](https://github.com/daintreehq/daintree/blob/88e295a07b833e7b56aa8b949b0f05334e344fae/src/components/Layout/Toolbar.tsx#L655-L685)). Adding a new toolbar item means touching both layers and remembering which store owns its visibility.

The user-facing result is that "enable item in the toolbar" in settings doesn't really make sense anymore — the agent tray is already the place where you decide what's in the toolbar (for agents), and the rest of the toolbar items are a small fixed set.

## Desired direction

Rather than just patching the icon, the goal is to collapse the two systems into one. The implementing agent should investigate the current architecture and propose a unified model before writing code. Some directions worth weighing:

- **Pinning as the universal model.** Treat the agent tray as the canonical "all available items" view, and let pin/unpin be the only mechanism for putting things in the toolbar — including non-agent items. The "enable in toolbar" checkboxes in settings go away.
- **Single registry.** One source of truth (label, icon, availability, default-pinned, plugin contributions) that both the tray and the toolbar consume. The icon mismatch becomes structurally impossible.
- **Agent tray as one section.** Treat the entire agent tray as one logical section in the toolbar. Whatever the customization UI currently exposes "after" the agent tray section probably folds into the same registry.

## Constraints

- Existing user customizations (pinned agents, hidden buttons) must migrate cleanly. Don't reset users' toolbars on upgrade.
- Plugin-contributed toolbar buttons (`usePluginToolbarButtons`) must keep working in whatever model emerges.

gregpriday:
## Summary

- Two parallel visibility systems (`useToolbarPreferencesStore.hiddenButtons` for non-agent buttons, `useAgentSettingsStore.pinned` for agents) are collapsed into a single source of truth
- `toolbarButtonMetadata.ts` canonicalizes labels, icons, and default visibility for all toolbar items — making icon drift like the `Plug`/`Puzzle` mismatch structurally impossible
- `useUnifiedToolbarVisibility` replaces the duplicated `isEffectivelyVisible` logic that lived independently in both `Toolbar.tsx` and `ToolbarSettingsTab.tsx`

Resolves #7666. Also folds in #7668 (agent-tray icon drift).

## Changes

- New `src/components/Layout/toolbarButtonMetadata.ts` — central registry (label, icon, default visibility) for all toolbar items
- New `src/hooks/useUnifiedToolbarVisibility.ts` — unified visibility hook consumed by both the toolbar and the settings tab
- `ToolbarSettingsTab.tsx` simplified significantly; no longer duplicates visibility logic
- `Toolbar.tsx` visibility logic pulled out in favour of the hook
- Existing user customizations (pinned agents, hidden buttons) migrate cleanly — no store shape changes

## Testing

- Updated unit tests for `ToolbarSettingsTab`, `useUnifiedToolbarVisibility`, and the agent registry integration
- Verified pinned-agent and hidden-button state round-trips correctly after the refactor
- Confirmed plugin-contributed toolbar buttons still work through the unified model

### DIFF
--- a/src/components/Layout/Toolbar.tsx
+++ b/src/components/Layout/Toolbar.tsx
@@ -2,9 +2,6 @@ import { useRef, useState, useEffect, useLayoutEffect, useMemo, useCallback } fr
 import type React from "react";
 import { Button } from "@/components/ui/button";
 import {
-  SlidersHorizontal,
-  SquareTerminal,
-  AlertCircle,
   GitCommit,
   GitPullRequest,
   CircleDot,
@@ -12,12 +9,9 @@ import {
   PanelLeftClose,
   Check,
   ChevronsUpDown,
-  Globe,
   MonitorPlay,
-  Bell,
   Ellipsis,
   GitBranch,
-  Plug,
 } from "lucide-react";
 import { Spinner } from "@/components/ui/Spinner";
 import { Folders, McpServerIcon } from "@/components/icons";
@@ -68,6 +62,7 @@ import { useOverflowBadgeSeverity, type OverflowBadgeSeverity } from "./useOverf
 
 import { BUILT_IN_AGENT_IDS, type BuiltInAgentId } from "@shared/config/agentIds";
 import { getAgentConfig } from "@/config/agents";
+import { TOOLBAR_BUTTON_METADATA } from "./toolbarButtonMetadata";
 
 const AGENT_TOOLBAR_IDS = new Set<ToolbarButtonId>([
   "agent-tray",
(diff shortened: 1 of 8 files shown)


gregpriday:
Closing — structural conflict with #7673's tri-state toolbar visibility. The unified `useUnifiedToolbarVisibility` hook needs to consume `cliAvailabilityStore` and switch to `isAgentToolbarVisible` (which now takes an availability param). Reopen after reconciling.

PR #7765, opened 2026-05-11
refactor(ipc): codegen IpcInvokeMap from defineIpcNamespace blocks

### LINKED ISSUE #7684: Codegen IpcInvokeMap from defineIpcNamespace blocks
[`shared/types/ipc/maps.ts`](https://github.com/daintreehq/daintree/blob/fcd8e47991d8a3bf349e7eadb83f19593df08698/shared/types/ipc/maps.ts) is 2,668 lines of hand-maintained `IpcInvokeMap` entries that mirror what already exists in the `defineIpcNamespace` blocks under `electron/ipc/handlers/`. Every payload added to a handler also has to be added to `maps.ts` by hand, which is the single largest source of drift between handlers and types.

Replace it with a codegen pass.

`scripts/codegen/ipc-map.mjs` walks the namespace files with `ts-morph`, extracts each `op()` / `opValidated()` call's channel string and handler signature, and emits `shared/types/ipc/generated.ts`. `maps.ts` reduces to a re-export of `generated.ts` plus the few non-codegen-able types that already live there (`ChecklistItemId`, `OnboardingState`, `HelpAssistantTier`, `MainProcessToastPayload`).

For `opValidated` handlers, the script runs the Zod schema through `zod-to-ts@^2`. That package's peer dep is `zod ^4.0.0`, which matches the project's `zod 4.3.6`. `ts-morph` and `zod-to-ts` both need to be added to `devDependencies`.

Wire `npm run check:ipc` to run the codegen, and `npm run check:ipc -- --check` to fail CI on diff — mirroring the existing `format:check` / `check:help` pattern. Fold the script into the aggregate `npm run check`. Once `electron/ipc/channels.ts` is gone (handled by #7680), `scripts/ci/check-channel-drift.mjs` retires; the codegen-diff replaces it.

## Preconditions before the first PR

Two things to confirm:

1. **ts-morph spike on return types.** `op()` handlers return `Promise<SafeResult<K>>` where `SafeResult<K>` is `ForbidIpcEnvelopeKeys` applied to `IpcInvokeMap[K]['result']` (see [`electron/ipc/define.ts`](https://github.com/daintreehq/daintree/blob/fcd8e47991d8a3bf349e7eadb83f19593df08698/electron/ipc/define.ts)). The codegen needs the underlying type, not the wrapped form. Worth a short ts-morph prototype against the 5 migrated namespaces to confirm `TypeChecker` resolves through the wrappers cleanly. Failure mode is everything coming out as `unknown`.

2. **IpcEventMap scope decision.** The main→renderer push event side uses `typedSend` / `sendToRenderer*` and has its own type map. The routing semantics are different (multi-window targeting per lesson #4641), and events don't currently live in `defineIpcNamespace` blocks. Either (a) extend the codegen to handle events with a different AST shape, or (b) leave `IpcEventMap` hand-maintained for now and revisit. Pick before the first PR ships.

## Out of scope

Collapsing `electron/preload.cts` to a single invoker and the renderer-side typed Proxy. That's a separate design call — the rename volume is large and the back-compat-shim-vs-codemod question deserves its own issue.

Depends on #7680 reaching far enough that the codegen pilot has a critical mass of namespaces to walk; can start in pilot mode against the 5 already-migrated namespaces (`clipboard`, `slashCommands`, `globalEnv`, `accessibility`, `help`) before #7680 finishes.

gregpriday:
## Summary

- Adds `scripts/codegen/ipc-map.mjs` — a `ts-morph` walker that extracts channel entries from `defineIpcNamespace` blocks across the migrated handler files and emits `shared/types/ipc/generated.ts`
- Removes 352 lines of hand-maintained duplicate entries from `shared/types/ipc/maps.ts`; `IpcInvokeMap` now extends `GeneratedIpcInvokeMap` and the file is down to the types that can't be derived from handler signatures
- Adds `npm run codegen:ipc` (write) and `npm run check:ipc` (diff-only, fails on drift); `check:ipc` is wired into the `check` umbrella so CI catches stale generated output

Resolves #7684.

## Changes

- `scripts/codegen/ipc-map.mjs` — ts-morph walker; handles `op`, `opValidated` (ctx-stripping), inline fn bodies, named-tuple args, and `Promise<T>` unwrapping
- `shared/types/ipc/generated.ts` — 73 channel entries across 12 namespaces, auto-generated (316 lines)
- `shared/types/ipc/maps.ts` — 352-line deletion; 5 unused imports removed
- `electron/ipc/__tests__/channelDrift.test.ts` — `extractInvokeMapKeys` updated to parse both `maps.ts` and `generated.ts` via a shared `extractKeysFromBlock` helper
- `scripts/codegen/__tests__/` — 6 fixture files, 12 unit tests covering happy path and defensive guards (duplicate channel detection, `opValidated` arity rejection, `any`-guard)
- `package.json` + `package-lock.json` — `ts-morph` added as devDep; `codegen:ipc` and `check:ipc` scripts added

## Testing

- 12 new unit tests pass (`scripts/codegen/__tests__/ipc-map.test.ts`)
- `channelDrift.test.ts` updated and passing
- `npm run check:ipc` runs clean against the committed `generated.ts`
- `npm run check` exits clean (typecheck, lint, format, ipc-drift)


gregpriday:
CI status: `build` (typecheck + check:ipc + lint:ratchet + format:check) passes. The `check` job's `ProjectViewManager.eviction.test.ts > does not mutate maxCachedViews — user limit returns when pressure subsides` test times out at 15s — same failure is on develop's latest run (25662654621), so pre-existing on the base branch and unrelated to the codegen changes here (which only touch IPC types + a build-time script).

PR #7958, opened 2026-05-14
fix(updater): detect MSIX builds via process.windowsStore, not platform

### LINKED ISSUE #7940: Build-type detection treats every Windows build as a Store build
`isWindowsStoreBuild` in `shared/config/distribution.ts` returns true any time the platform is `win32`. It treats running on Windows as the same thing as being a Microsoft Store build.

Because of that, `electron-updater` never runs on any Windows build.

That's wrong. A non-Store Windows build gets misreported as a Store build, and the auto-updater stays disabled on a build that should be updating itself. The check needs to reflect whether the build is packaged as MSIX, not just which OS it's running on. `process.windowsStore` is the property that signals MSIX packaging — the main process has direct access to it, the renderer doesn't.

```ts
export function isWindowsStoreBuild(platform = getRuntimePlatform()): boolean {
  return platform === "win32";
}
```

[`shared/config/distribution.ts`](https://github.com/daintreehq/daintree/blob/846518b34f4ba522808ca77678cb79c50d579072/shared/config/distribution.ts#L25-L27)

The predicate is used in 7 places across the main process and the renderer: the 3 early-returns in `AutoUpdaterService`, the "Check for Updates…" guard in the application menu, and 3 renderer hooks and components (`GeneralTab`, `useUpdateListener`, `useMainProcessToastListener`). All of them currently key off the platform check, so the rename also needs the main-process value bridged to the renderer.

This is a latent bug today — no non-Store Windows build ships yet — but it's also the prerequisite for adding one.

gregpriday:
## Summary

- `isWindowsStoreBuild()` in `shared/config/distribution.ts` was returning `true` for any `win32` platform, not just MSIX-packaged builds — meaning `electron-updater` was silently disabled on non-Store Windows builds
- Fixed by reading `process.windowsStore` (the authoritative MSIX signal) and bridging it to the renderer as a synchronous boolean via `contextBridge`, so all 7 call sites across main and renderer get the correct value without any interface changes
- Strengthened tests to cover all 3 main-process call sites in `AutoUpdaterService` independently and the renderer bridge/`process` precedence

Resolves #7940

## Changes

- `shared/config/distribution.ts` — `isWindowsStoreBuild()` now reads `process.windowsStore` with a `win32` platform guard as a fallback; accepts an optional override for testability
- `shared/types/ipc/api.ts` — adds `windowsStore: boolean` to the preload API surface
- `electron/preload.cts` — bridges `process.windowsStore` to the renderer
- `shared/config/__tests__/distribution.test.ts` — comprehensive coverage of the new logic including `process.windowsStore` precedence
- `electron/services/__tests__/AutoUpdaterService.test.ts` — isolated tests for all 3 guard call sites

## Testing

Unit tests cover the corrected `isWindowsStoreBuild()` behaviour, `process.windowsStore` precedence over platform, and all `AutoUpdaterService` guard paths. No call sites needed updating — the function signature is unchanged.

### DIFF
--- a/electron/__tests__/menu.test.ts
+++ b/electron/__tests__/menu.test.ts
@@ -289,15 +289,26 @@ describe("createApplicationMenu", () => {
 
 describe("update menu lifecycle", () => {
   const originalPlatform = process.platform;
+  const originalWindowsStore = (
+    process as Partial<NodeJS.Process & { windowsStore?: boolean }> & NodeJS.Process
+  ).windowsStore;
 
   beforeEach(() => {
     vi.clearAllMocks();
     capturedTemplate = [];
     autoUpdaterServiceMock.__setMenuState("idle");
+    Object.defineProperty(process, "windowsStore", {
+      value: originalWindowsStore,
+      configurable: true,
+    });
   });
 
   afterEach(() => {
     Object.defineProperty(process, "platform", { value: originalPlatform, configurable: true });
+    Object.defineProperty(process, "windowsStore", {
+      value: originalWindowsStore,
+      configurable: true,
+    });
     Object.defineProperty(app, "isPackaged", { value: false, configurable: true });
   });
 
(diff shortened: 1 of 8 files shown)


gregpriday:
Closing as superseded by #7963 (NSIS Windows installer), which already landed the MSIX detection fix on develop using a different signature: `isWindowsStoreBuild(windowsStore?: boolean)` reads `process.windowsStore` + the explicit param, with the renderer routed through `useDistributionStore` populated by `HydrateResult.isWindowsStore`.

The renderer-bridge precedence approach in this PR conflicts with develop's signature and would require a structural rewrite of the additive bits (preload bridge + AutoUpdater hardening tests) to reconcile. The hardening tests for the Windows Store guard are still worth porting — happy to take those as a small follow-up if needed.

Unassigning #7940 so it can come back into the work pool with fresh context now that #7963 has shipped.

PR #8212, opened 2026-05-17
refactor(toolbar): extract shared helper for toolbar visibility dispatch

gregpriday:
## Summary

- `BUILT_IN_AGENT_ID_SET` was defined in three places (`ToolbarSettingsTab.tsx`, `toolbarButtonMetadata.ts`, `useOverflowBadgeSeverity.ts`). It now lives in `shared/utils/agentPinned.ts` next to its sibling resolvers and the local copies are gone.
- New `src/lib/toolbarVisibilityDispatch.ts` handles the agent-vs-non-agent branch in one place: agent IDs route to `setAgentPinned` (preserving tri-state semantics from #7673), non-agent IDs to `toggleButtonVisibility`, with an optional `explicitPinned` for unpin actions.
- `handleToggleLeft`/`handleToggleRight` in `ToolbarSettingsTab` collapsed into a single `handleToggle`; `AgentButton.handleUnpinFromToolbar` updated to use the same helper. Third call site is ready for the follow-up.

Resolves #8181

## Changes

- Export `BUILT_IN_AGENT_ID_SET` from `shared/utils/agentPinned.ts`, remove three local copies
- Add `src/lib/toolbarVisibilityDispatch.ts` with the unified dispatch helper
- Collapse `handleToggleLeft`/`handleToggleRight` into `handleToggle` in `ToolbarSettingsTab`
- Unify `AgentButton.handleUnpinFromToolbar` to the helper
- Switch `useOverflowBadgeSeverity` to `isBuiltInAgentId` from `@shared/config/agentIds`
- Rename "Reset to Defaults" to "Reset toolbar" per verb-noun microcopy rule

## Testing

93 unit tests pass across the 4 affected files. Added 2 new tests covering the unpin path and tri-state semantics. ESLint and tsc clean.

PR #8218, opened 2026-05-17
feat(agent-tray): polish discovery dot, pinned icon, MRU, and TTL backstop

### LINKED ISSUE #8177: Polish AgentTrayButton: discovery, pinned-row, and MRU signals
The agent tray dropdown (`AgentTrayButton`, the Plug-icon toolbar button opening a Radix DropdownMenu with a row for every built-in agent) has a few polish gaps. None are huge on their own; together they tighten up the discovery, pinned-row, and frecency signals on this surface.

## 1. NEW indicator visual mismatch

Each launch row shows the NEW indicator as an outlined pill. The trigger button on the same surface already uses a small `status-info` dot for the equivalent "new agents detected" signal. The two treatments don't match.

Switch the per-row indicator to the same dot pattern, and add an `aria-label` so the screen-reader signal the pill currently carries (the literal word "New") doesn't get dropped on the swap.

## 2. NEW decay fires on the wrong event

`markAgentsSeen` is dispatched for every launchable agent the moment the dropdown opens. A user who opens the tray to launch one agent burns the NEW signal on every other one in the same gesture.

Decay should be per-launch: clear the NEW state only for the agent the user actually engaged with.

## 3. No backstop on the NEW signal

A user who never launches a freshly-installed agent gets a permanent dot. Worth a TTL backstop alongside the per-launch decay above.

One way: add an `availabilityFirstSeen` field to `OnboardingState` (bump `schemaVersion`) and clear the signal when the timestamp ages past roughly two weeks. The exact window is a judgment call at PR time.

## 4. Pinned-and-in-tray redundancy is unexplained

When an agent is pinned to the toolbar AND appears in the tray, the tray row's filled Pin icon is always on at full opacity. It reads as an active control rather than a state marker, which makes the toolbar/tray duplication look unintentional.

Dialling the pinned-state Pin to a quieter opacity (or otherwise muting it) is enough to communicate "this is mirrored on the toolbar" without the heavier escalation of a dedicated "Pinned" section header.

## 5. Tray launches don't feed the MRU store

The tray sorts itself by `useActionMruStore` frecency, but only palette dispatches actually call `recordActionMru`. Tray launches never record, so the launcher's own usage doesn't feed its own ordering. There's already an inline comment in `AgentTrayButton` flagging this.

A `recordActionMru` call in the tray's launch handler closes the loop.

## Out of scope

- The inline shortcut-capture mechanism on launch rows. Recently shipped, leave alone.
- A hard cap on the Launch group size for when agent count grows past ~20. Not warranted at the current count.

gregpriday:
## Summary

- Replaces the outlined "NEW" pill with a status-info dot (matching the trigger button), adds `, new` aria-label suffix, and gives `SplitLaunchItem` SubTrigger parity dot + aria-label
- `markAgentsSeen` now fires per-launch on the launched agent only — opening the dropdown no longer bulk-clears all ready agents
- Adds a 14-day discovery-badge TTL backstop via `availabilityFirstSeen: Record<string, number>` on `OnboardingState`, a new `ONBOARDING_RECORD_AVAILABILITY_FIRST_SEEN` IPC channel, and an idempotent handler; a `useEffect` records first-seen when an agent becomes launchable, with finite-number sanitization in both renderer and main so a corrupt timestamp can't pin the dot indefinitely
- Dims the pinned `Pin` icon from `opacity-100` to `opacity-50` so it reads as a state marker, not an active control
- Tray launches now call `recordActionMru` so tray usage is reflected in the tray's own MRU sort

Resolves #8177

## Changes

- `AgentTrayButton.tsx` — dot indicator, per-launch `markAgentsSeen`, `recordActionMru` on launch, dimmed pin icon
- `useAgentDiscoveryOnboarding.ts` (new) — `availabilityFirstSeen` tracking hook with 14-day TTL and finite-number guard
- `electron/ipc/channels.ts`, `handlers/onboarding.ts`, `preload.cts`, `shared/types/ipc/` — `ONBOARDING_RECORD_AVAILABILITY_FIRST_SEEN` IPC channel wired end-to-end
- `electron/store.ts` — `availabilityFirstSeen` field persisted to `OnboardingState`

## Testing

69 tests passing across `AgentTrayButton.test.tsx` (updated) and the new `useAgentDiscoveryOnboarding.test.ts`. Full verify (typecheck across all 3 configs, format, lint, IPC drift check, build) all green.

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -594,6 +594,7 @@ export const CHANNELS = {
   ONBOARDING_MARK_NEWSLETTER_SEEN: "onboarding:mark-newsletter-seen",
   ONBOARDING_MARK_WAITING_NUDGE_SEEN: "onboarding:mark-waiting-nudge-seen",
   ONBOARDING_MARK_AGENTS_SEEN: "onboarding:mark-agents-seen",
+  ONBOARDING_RECORD_AVAILABILITY_FIRST_SEEN: "onboarding:record-availability-first-seen",
   ONBOARDING_DISMISS_WELCOME_CARD: "onboarding:dismiss-welcome-card",
   ONBOARDING_DISMISS_SETUP_BANNER: "onboarding:dismiss-setup-banner",
   ONBOARDING_CHECKLIST_GET: "onboarding:checklist-get",
--- a/electron/ipc/handlers/onboarding.ts
+++ b/electron/ipc/handlers/onboarding.ts
@@ -23,7 +23,7 @@ const SKIP_E2E = process.env.DAINTREE_E2E_SKIP_FIRST_RUN_DIALOGS === "1";
 function getOnboardingState(): OnboardingState {
   if (SKIP_E2E) {
     return {
-      schemaVersion: 1,
+      schemaVersion: 2,
       completed: true,
       currentStep: null,
       agentSetupIds: [],
@@ -31,6 +31,7 @@ function getOnboardingState(): OnboardingState {
       newsletterPromptSeen: true,
       waitingNudgeSeen: true,
       seenAgentIds: [],
+      availabilityFirstSeen: {},
       welcomeCardDismissed: true,
       setupBannerDismissed: true,
       checklist: {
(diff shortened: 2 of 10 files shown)


gregpriday:
Closing for now — CI red on `src/components/__tests__/agentPinSync.integration.test.tsx` (`TypeError: Cannot read properties of undefined (reading 'claude'/'gemini')`). The same failure appears on #8212, so this looks like a shared test fixture issue against current develop, not a PR-local regression. Once the test is repaired, a fresh PR (or reopening this one) rebased on develop will let the autonomous merge pipeline pick it up cleanly.

PR #8220, opened 2026-05-17
feat(toolbar): polish project pill — ARIA, truncation, tooltip, context menu

### LINKED ISSUE #8174: Polish toolbar project pill: ARIA, truncation, tooltip, context menu
Five small interaction-fidelity gaps on the center project pill button in the main toolbar, all on the same surface.

## Branch chip end-truncates and loses the meaningful tail

The branch chip in the pill renders the raw branch name with CSS `truncate`, so `feature/migrate-from-canvas-renderer-to-webgl-2026` becomes `feature/migrate-from…` — indistinguishable from any other branch sharing the `feature/migrate-` prefix. The codebase already has a `middleTruncate` utility and a `BranchLabel` component that solve this for branches in the worktree sidebar and elsewhere; the toolbar pill is the surface still using the bare-text pattern.

## No tooltip recovery when content truncates or hides

The pill has no hover or focus disclosure. Once the project name end-truncates, or the branch chip hides at the existing 700px container query, there is no path back to the full project + branch label. The shared `Tooltip` primitive is already imported in `Toolbar.tsx`.

## ARIA doesn't match the codebase's combobox convention

The pill is the trigger for `ProjectSwitcherPalette`, but it's a plain `<button>` with only a conditional `aria-label`. Five comparable list-of-values triggers in the codebase already use `role="combobox"` + `aria-haspopup="listbox"` + `aria-expanded`: `BaseBranchCombobox`, `ExistingBranchPicker`, `RecipePickerPopover`, `NewBranchInput`, and `SearchablePalette`. The pill is the odd one out, and `ProjectSwitcherPalette` itself is built on `SearchablePalette`, so the popup advertises listbox semantics the trigger doesn't reflect. When a project is set, the visible content (`{emoji} {name} {branch}`) is the natural accessible name; the static `"Open project"` label only fits the empty state.

## Responsive collapse stops a tier short

The container query at 700px hides the chip label but leaves the `GitBranch` icon as a vestigial bullet, and there's no further degradation below that. The chip wrapper keeps a reserved-but-empty slot even when there's nothing useful left in it.

## No right-click affordance on the identity surface

The pill represents the current project but exposes none of its actions. The same project-row context menu in `ProjectSwitcherPalette` (Open in new window, Pin/Unpin, Copy path, Stop all agents, Close project) is already wired to every row in the dropdown — including the active project — but not to the toolbar's identity representation of that project. Right-clicking the pill does nothing.

**Evidence:**

- [`src/components/Layout/Toolbar.tsx#L1040-L1075`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/components/Layout/Toolbar.tsx#L1040-L1075) — the pill `<button>` with emoji slot, name span, branch chip, and `ChevronsUpDown`
- [`src/styles/components/toolbar.css#L96-L129`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/styles/components/toolbar.css#L96-L129) — `.toolbar-project-pill` and the single container-query tier
- [`src/utils/textParsing.ts#L44-L55`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/utils/textParsing.ts#L44-L55) — existing `middleTruncate` utility
- [`src/components/Worktree/BranchLabel.tsx`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/components/Worktree/BranchLabel.tsx) — branches rendered with prefix typing, middle-truncation, and tooltip
- [`src/components/ui/SearchablePalette.tsx#L290-L298`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/components/ui/SearchablePalette.tsx#L290-L298) — combobox-role pattern used by `ProjectSwitcherPalette` and the four other list-of-values triggers
- [`src/components/Project/ProjectSwitcherPalette.tsx#L265-L324`](https://github.com/daintreehq/daintree/blob/decf4fa119925f2a1dfa694b1379c0352ab3d7c9/src/components/Project/ProjectSwitcherPalette.tsx#L265-L324) — per-row context menu with the active-project actions

gregpriday:
## Summary

Five interaction-fidelity gaps on the toolbar project pill, all on the same surface.

- Adds `role="combobox"` + `aria-haspopup="listbox"` + `aria-expanded` to match the combobox convention used by the five other list-of-values triggers in the codebase
- Middle-truncates branch names via the existing `middleTruncate` utility so long `feature/*` names stay distinguishable instead of collapsing to the same prefix
- Adds a `Tooltip` showing the full project + branch label whenever content truncates or the branch chip is hidden at narrow widths
- Adds a second container-query tier that removes the vestigial `GitBranch` icon when the chip label is already hidden, cleaning up the reserved-but-empty slot
- Wires the existing per-row project context menu (Open in new window, Pin/Unpin, Copy path, Stop all agents, Close project) to a right-click handler on the pill — same actions already available in the dropdown, now reachable from the identity surface

Also fixes a bug where `activeProject` was derived from the capped search results slice rather than the full project list, which caused the active indicator to disappear when the active project fell outside the cap.

Resolves #8174

## Changes

- `src/components/Layout/Toolbar.tsx` — combobox ARIA, `middleTruncate` for branch chip, `Tooltip` wrapper, second collapse tier, `onContextMenu` handler
- `src/styles/components/toolbar.css` — second container-query breakpoint for the icon-only state
- `src/hooks/useProjectSwitcherPalette.ts` — expose full project list alongside search results so callers can resolve the active project correctly
- `src/components/Layout/__tests__/Toolbar.projectPill.test.ts` — 130-line test covering all five gaps
- `src/hooks/__tests__/useProjectSwitcherPalette.test.tsx` — additional coverage for full-list exposure

## Testing

- Unit tests cover ARIA attributes in both empty and loaded states, middle-truncation output, tooltip content at narrow widths, second collapse tier class application, context menu invocation, and the active-project fix
- `npm run check` passes (typecheck + lint + format)

### DIFF
--- a/src/components/Layout/Toolbar.tsx
+++ b/src/components/Layout/Toolbar.tsx
@@ -309,7 +309,7 @@ export function Toolbar({
   );
 
   const handleCopyProjectPath = useCallback((path: string) => {
-    void navigator.clipboard.writeText(path).catch(console.error);
+    void navigator.clipboard.writeText(path).catch(console.warn);
   }, []);
 
   useEffect(() => {
--- a/src/components/Layout/Toolbar.tsx
+++ b/src/components/Layout/Toolbar.tsx
@@ -12,6 +12,11 @@ import {
   MonitorPlay,
   Ellipsis,
   GitBranch,
+  Pin,
+  PinOff,
+  Clipboard,
+  Square,
+  X,
 } from "lucide-react";
 import { Spinner } from "@/components/ui/Spinner";
 import { Folders, McpServerIcon } from "@/components/icons";
@@ -25,12 +30,20 @@ import { AgentTrayButton } from "./AgentTrayButton";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { ShortcutRevealChip } from "@/components/ui/ShortcutRevealChip";
 import {
+  ContextMenu,
+  ContextMenuContent,
+  ContextMenuItem,
+  ContextMenuSeparator,
+  ContextMenuTrigger,
+} from "@/components/ui/context-menu";
+import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
+import { middleTruncate } from "@/utils/textParsing";
 import { useToolbarOverflow } from "@/hooks/useToolbarOverflow";
 import { useWorktreeActions } from "@/hooks/useWorktreeActions";
 import { useAriaKeyshortcuts, useKeybindingDisplay, useShortcutHintHover } from "@/hooks";
@@ -186,6 +199,23 @@ export function Toolbar({
     activeWorktreeId ? state.worktrees.get(activeWorktreeId) : null
   );
   const branchName = activeWorktree?.branch;
+  // Branch label gets middle-truncated to preserve the meaningful tail
+  // (e.g. `feature/...-pill` instead of `feature/migrate-from-…`). The full
+  // value still appears in the tooltip on hover/focus.
+  const truncatedBranchName = branchName ? middleTruncate(branchName, 24) : undefined;
+  // Source of truth for runtime-only fields (processCount, isPinned) the
+  // context menu needs to gate its items. `currentProject` (Project type)
+  // doesn't carry those — they live on the SearchableProject derived in the
+  // palette hook.
+  const activeSearchableProject = useMemo(
+    () => projectSwitcher.results.find((project) => project.isActive),
+    [projectSwitcher.results]
+  );
+  const tooltipLabel = currentProject
+    ? branchName
+      ? `${currentProject.name} / ${branchName}`
+      : currentProject.name
+    : "Open project";
 
   useEffect(() => {
     loadProjects();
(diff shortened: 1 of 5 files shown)


gregpriday:
Closing — `src/components/Layout/__tests__/Toolbar.projectPill.test.ts` is a 130-line string-matching test that reads `Toolbar.tsx` / `toolbar.css` as raw strings and asserts on literal tokens (e.g. `aria-haspopup="listbox"`, `middleTruncate(branchName, 24)`, `@container toolbar (max-width: 560px)`). That's the "no tautological style tests" anti-pattern — passes for any string-equivalent JSX, breaks on benign refactors, and would miss real regressions like a Radix `asChild` prop-merge conflict that breaks tooltip-over-context-menu nesting.

The production code itself is clean: combobox ARIA matches convention, ContextMenuTrigger+Tooltip nesting has direct precedent in `AgentButton.tsx`, the pill context menu mirrors the palette row's existing menu exactly, and the `activeProject` hook fix correctly decouples pin/processCount gating from the 15-item `results` window. The other new test, `useProjectSwitcherPalette.test.tsx`, is fine — uses `renderHook` and asserts observable hook output.

Reopening with the Toolbar.projectPill test rewritten using `@testing-library/react` + role-based queries (or dropped in favor of the existing observable-behavior coverage) will let the pipeline merge cleanly.

PR #8260, opened 2026-05-18
refactor(notify): adopt supersedeKey and context.eventKind at five call sites

### LINKED ISSUE #8248: Adopt `supersedeKey` and `context.eventKind` at notify call sites
The `notify()` dispatcher supports two adoption-ready primitives that callers rarely pass. Both have exactly one production site each today. Wiring them at the named call sites is mechanical and delivers a directly user-visible payoff.

### `supersedeKey` — state-pair retirement

When a later `notify()` call carries the same `supersedeKey`, the prior non-archived inbox entry with that key is archived automatically. Only `src/store/listeners/panel/lifecycle.ts` uses it in production.

Two call sites adopt it cleanly.

**`src/hooks/useDiskSpaceWarnings.ts`** carries a bespoke `findLiveDiskSpaceToastId` helper plus a `dismissNotification` call on the `normal` status path. Replace the lookup and the dismiss with `supersedeKey: "disk-space"` on both warning and critical paths. On `normal`, either fire a `priority: "low"` resolution row carrying the same key (archives the warning automatically) or keep the explicit dismiss if no resolution row is desired. Either form deletes the manual lookup.

**`src/hooks/useGitHubTokenExpiryNotification.ts`** is a hysteresis latch. It fires a warning toast on the false-to-true transition and resets `firedRef` on recovery with no notify call. The user gets no signal that the new token actually validated. Two-part wiring:

- Add `supersedeKey: "github.token"` to the warning toast.
- On the true-to-false transition (the recovery branch at line 50), emit a `priority: "low"` "Token validated" inbox row with the same key. The warning row archives automatically, the user gets a quiet confirmation, and screen-reader / keyboard-only users get an explicit acknowledgement.

### `context.eventKind` — per-kind silence affordance

Setting `context.eventKind` (one of `completed`, `waiting`, `workingPulse`, `uiFeedback`) surfaces a "Silence completed notifications" kebab affordance on the toast and in the notification-center entry. The dispatcher and the kebab are already wired; callers just need to pass the value.

The only production site setting it today is `src/store/listeners/panel/identity.ts`. Three more sites map cleanly:

- `src/lib/watchNotification.ts` (`Agent task completed`) gets `context: { eventKind: "completed", panelId }`.
- `src/hooks/app/useAgentWaitingNudge.ts` gets `context: { eventKind: "waiting", panelId }`. Orthogonal to its existing `placement: "grid-bar"`.
- `src/hooks/useIdleTerminalNotifications.ts` gets `context: { eventKind: "completed", panelId }`.

After these land, the silence-on-the-toast affordance becomes discoverable on the agent-completion, agent-waiting, and idle-terminal toasts.

### Out of scope

A shared `<feature>SupersedeKey` helper. Worth introducing once `supersedeKey` adoption crosses around five production sites. With three after this issue it adds little leverage and risks premature abstraction.

gregpriday:
Closes #8248.

## Summary

Wires `supersedeKey` and `context.eventKind` at the five call sites called out in the issue. Both fields were fully implemented in the dispatcher but used by only one production site each, so the user-visible "silence completed notifications" kebab and the inbox state-pair retirement only worked in a fraction of the places they should.

## Changes

- **`useDiskSpaceWarnings`** — removed the bespoke `findLiveDiskSpaceToastId` helper and the manual dismiss path. Warning and critical notifies now carry `supersedeKey: "disk-space"`; the `normal` status emits a low-priority "Disk space restored" inbox row carrying the same key, which archives the prior warning entry automatically.
- **`useGitHubTokenExpiryNotification`** — added `supersedeKey: "github.token"` to the warning, and a low-priority "Token validated" recovery row on the `isTokenError` true→false transition. Recovery is gated on the caller's `isTokenError` actually clearing: a momentary `isUnhealthy` drop while `isTokenError` stays true silently re-arms the latch without emitting a premature confirmation (caught in review).
- **`watchNotification`** — agent-completed toast now carries `context: { eventKind: "completed", panelId }`. "Silence completed notifications" becomes discoverable from the toast kebab and the inbox row.
- **`useAgentWaitingNudge`** — captures the triggering `panelId` from both the hydrate-time and subscribe-time paths and passes `context: { eventKind: "waiting", panelId }`. The grid-bar bar itself doesn't render a kebab, but the inbox row does, so the silence affordance is reachable from the notification center.
- **`useIdleTerminalNotifications`** — `context: { eventKind: "completed" }`; `projectId` is included only in the single-project case. Multi-project payloads omit the scope so the "Mute project notifications" affordance can't fire against the wrong target.

## Notes

- The shared `<feature>SupersedeKey` helper called out as out-of-scope in #8248 is deliberately not introduced. With three `supersedeKey` sites after this PR, abstraction adds little.
- `supersedeKey` archives the inbox entry but does not dismiss an already-visible toast. For the sticky critical disk-space toast, the resolution row provides the positive inbox confirmation while the toast persists until user dismissal — explicitly accepted in the issue.
- `useIdleTerminalNotifications` uses `projectId` instead of the issue-templated `panelId` because the IPC payload (`IdleTerminalNotifyPayload`) is project-scoped and carries no per-panel identifiers.

## Test plan

- [x] `useGitHubTokenExpiryNotification.test.tsx` — asserts new `supersedeKey`, latch tests updated, added explicit recovery emission + recovery suppression scenarios (`isUnhealthy` drop while `isTokenError` stays true)
- [x] `useAgentWaitingNudge.test.tsx` — asserts `context: { eventKind: "waiting", panelId: "t1" }`
- [x] `npm run check` — typecheck, lint, format, IPC drift, lint ratchet all green
- [ ] Manual: trigger a disk-space warning then free space — confirm the "Disk space restored" inbox row archives the prior warning entry
- [ ] Manual: trigger a GitHub token error then re-auth — confirm the "Token validated" inbox row appears and the warning archives
- [ ] Manual: confirm the "Silence completed notifications" kebab is discoverable on agent-completed and idle-terminal toasts

### DIFF
--- a/src/hooks/__tests__/useGitHubTokenExpiryNotification.test.tsx
+++ b/src/hooks/__tests__/useGitHubTokenExpiryNotification.test.tsx
@@ -134,7 +134,7 @@ describe("useGitHubTokenExpiryNotification", () => {
     expect(notifyMock).toHaveBeenCalledTimes(1);
   });
 
-  it("latch resets when health recovers while error persists, re-fires on next unhealthy", () => {
+  it("latch resets silently when health recovers while error persists, re-fires on next unhealthy", () => {
     useGitHubTokenHealthStore.setState({ isUnhealthy: true });
     const { rerender } = renderHook(
       ({ isTokenError }) => useGitHubTokenExpiryNotification(isTokenError),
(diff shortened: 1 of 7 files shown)


gregpriday:
CI run #1 hit a known flake in `src/components/GitHub/__tests__/CommitList.test.tsx > Enter key handling > Enter on a commit with body toggles its expansion` (aria-expanded timing assertion). Test is unrelated to this PR's notify-callsite changes and passes locally twice in a row. Reran the failed jobs — watching for green.

PR #8315, opened 2026-05-18
feat(dev-preview): rebuild blocked-navigation banner with full URL, OAuth status, and cancel flow

### LINKED ISSUE #8263: Rebuild dev preview blocked-navigation banner UX
The dev preview pane's `BlockedNavBanner` blocks the right things (any non-localhost navigation, with an RFC 8252 loopback handoff for OAuth-looking URLs per #4892), but the UX layer around it leaves the user staring at a silent IDE during a five-minute flow. Several gaps to fix together.

## Show the full URL, not just the hostname

The banner reads `Navigation to external site blocked: {hostname}`. The full URL is in state but never rendered. For a third-party redirect the path matters: a user can't tell `accounts.google.com/oauth?...` from `accounts.google.com.attacker.tld/oauth?...` when both render identically as the same hostname. The 2026 convention (Chrome's address-bar elision) is to render the full URL with visual emphasis on the registrable domain (eTLD+1) and de-emphasis on subdomains and path/query. `new URL(blockedNav.url).hostname` already returns the ASCII / Punycode form for IDN domains in Electron 41, so the existing parser handles homograph defense as long as the renderer doesn't decode back to Unicode.

A `Copy URL` button and `font-mono` rendering help — substitution characters are easier to spot in a monospace font and easier to audit outside the IDE.

## Don't auto-dismiss

The banner clears itself after 10 seconds. That isn't long enough to read an 80-character third-party URL, and a security-relevant inline banner shouldn't time itself out — the `×` button plus the action buttons (`Sign in via browser` / `Open in external browser`) already give the user every recovery path they need.

## Show OAuth status while the flow is in flight

When the user clicks `Sign in via Browser`, `startOAuthLoopback` runs and the banner immediately closes (`onDismiss()` fires inline). The loopback runs for up to five minutes (RFC 8252 server, CDP `Fetch.requestPaused` token-exchange intercept, sessionStorage restore on callback), and the user sees nothing until the webview reloads, or doesn't. The banner should stay open while the flow runs and surface the phases the main process can actually emit: started (after `shell.openExternal` resolves), token exchange intercepted, completed, timed out, error.

## Allow Cancel

Closing the banner today does **not** abort the 127.0.0.1 listener. The server keeps running until callback or the five-minute timeout, even though the user has visibly given up. `electron/services/OAuthLoopbackService.ts` already exports `cancelOAuthLoopback(panelId)` (used internally by `startOAuthLoopback` itself for duplicate-flow defense), but nothing exposes it through preload. The OAuth banner needs an explicit `Cancel` button that calls a new IPC wrapping that export.

## Stop swallowing failures

When the loopback returns `null` (timeout or server-error), the renderer catch block at the `Sign in via Browser` click handler silently falls through and the user is left with nothing. The banner should swap to a failure variant with the cause (`Timed out` for the 5-min path, `Sign-in didn't complete` for the error-param path) and a `Try again` link that re-fires `startOAuthLoopback`. "Browser closed" isn't observable from a loopback server's perspective, so the failure copy is bounded to what main can actually emit.

## Don't stack banners on a redirect chain

Once auto-dismiss is gone, a buggy redirect loop firing three `will-navigate` blocks to different hosts in 200ms would stack visible banner state. The 150ms debounce on `onNavigationBlocked` already exists, but it doesn't coalesce — keying the debounce by URL hostname (or just always reflecting the most-recent block) handles this without escalating to a toast.

### What done looks like

- The banner shows the full URL with eTLD+1 emphasis, in `font-mono`, with a `Copy URL` button.
- The banner persists until the user dismisses it (no 10s timer).
- During OAuth loopback, the banner stays open and surfaces phase transitions plus a `Cancel` button.
- Cancel and Try again both call typed IPCs that resolve through `OAuthLoopbackService`.
- Failure paths render a typed cause inline instead of silently disappearing.
- A redirect chain doesn't stack banner state.

### Evidence

- [`src/components/DevPreview/DevPreviewPane.tsx#L89-L174`](https://github.com/daintreehq/daintree/blob/d6609aea5af75a5e1b0622299416d60d295a2cd2/src/components/DevPreview/DevPreviewPane.tsx#L89-L174) — current `BlockedNavBanner` (hostname-only display, OAuth handoff that closes the banner immediately)
- [`src/components/DevPreview/DevPreviewPane.tsx#L689-L694`](https://github.com/daintreehq/daintree/blob/d6609aea5af75a5e1b0622299416d60d295a2cd2/src/components/DevPreview/DevPreviewPane.tsx#L689-L694) — the 10s auto-dismiss `useEffect`
- [`src/components/DevPreview/DevPreviewPane.tsx#L653-L687`](https://github.com/daintreehq/daintree/blob/d6609aea5af75a5e1b0622299416d60d295a2cd2/src/components/DevPreview/DevPreviewPane.tsx#L653-L687) — `onNavigationBlocked` listener with the existing 150ms debounce
- [`electron/services/OAuthLoopbackService.ts#L160-L165`](https://github.com/daintreehq/daintree/blob/d6609aea5af75a5e1b0622299416d60d295a2cd2/electron/services/OAuthLoopbackService.ts#L160-L165) — `cancelOAuthLoopback(panelId)` exported but not exposed via IPC
- [`electron/ipc/handlers/webview.ts`](https://github.com/daintreehq/daintree/blob/d6609aea5af75a5e1b0622299416d60d295a2cd2/electron/ipc/handlers/webview.ts) — current `startOAuthLoopback` IPC handler; the natural place to add `cancelOAuthLoopback` + extend the return / event shape to distinguish success / timeout / error

### Notes for the implementer

- The notify() forced-choice routes this inline (Tier 3 banner), not a toast. Escalation to a global toast for high-frequency redirect-loop blocks is a separate call; no telemetry motivates it today.
- Sentence-case microcopy: titles and buttons start with a capital and end without a period.
- Destructive Action Tiers — the external-browser handoff is closer to D2 (the user must see what they're opening), which is exactly why the full URL belongs in the banner before they click `Sign in via browser`.

gregpriday:
## Summary

Rebuilt the dev preview blocked-navigation banner UX. The old banner showed only a hostname, auto-cleared after 10 seconds, closed immediately on OAuth handoff, and silently swallowed failures.

- Shows the full URL with eTLD+1 emphasis in monospace, plus a Copy URL button.
- Persists until the user dismisses it -- no more 10-second auto-clear.
- Surfaces OAuth loopback phase transitions (started, token-exchange-intercepted, completed, timed-out, error) with Cancel and Try again buttons.
- Coalesces redirect-chain blocks by hostname, so a redirect loop doesn't stack banner state.
- Surfaces failure causes inline rather than silently disappearing.

Resolves #8263.

## Changes

12 files changed. Extracted 2 new components from `DevPreviewPane.tsx`:

- `src/components/DevPreview/BlockedNavBanner.tsx` -- full banner component with URL display, OAuth status, and action buttons.
- `src/components/DevPreview/useOAuthLoopbackStatus.ts` -- hook for tracking OAuth phases emitted from the main process.
- `electron/ipc/handlers/webview.ts` -- added `cancelOAuthLoopback` IPC, extended return shape to distinguish success/timeout/error.
- `electron/services/OAuthLoopbackService.ts` -- aligned generation counter, exposed phase events.
- `shared/utils/urlUtils.ts` -- eTLD+1 extraction for URL display emphasis.
- `electron/ipc/channels.ts`, `electron/preload.cts`, `shared/types/ipc/api.ts`, `shared/types/ipc/maps.ts` -- IPC plumbing for cancel and phase events.

## Testing

Typecheck clean, lint clean, format clean, unit tests pass. Manual smoke: blocked external navigation shows full URL with domain emphasis, OAuth loopback banner phases render correctly, Cancel aborts the loopback listener, Try again re-fires the flow.

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -354,6 +354,8 @@ export const CHANNELS = {
   WEBVIEW_FIND_SHORTCUT: "webview:find-shortcut",
   WEBVIEW_NAVIGATION_BLOCKED: "webview:navigation-blocked",
   WEBVIEW_OAUTH_LOOPBACK: "webview:oauth-loopback",
+  WEBVIEW_OAUTH_LOOPBACK_CANCEL: "webview:oauth-loopback-cancel",
+  WEBVIEW_OAUTH_LOOPBACK_STATUS: "webview:oauth-loopback-status",
   WEBVIEW_START_CONSOLE_CAPTURE: "webview:start-console-capture",
   WEBVIEW_STOP_CONSOLE_CAPTURE: "webview:stop-console-capture",
   WEBVIEW_CLEAR_CONSOLE_CAPTURE: "webview:clear-console-capture",
--- a/electron/ipc/handlers/webview.ts
+++ b/electron/ipc/handlers/webview.ts
@@ -1053,7 +1053,6 @@ export function registerWebviewHandlers(_deps: HandlerDependencies): () => void 
     typedHandle(CHANNELS.WEBVIEW_STOP_CONSOLE_CAPTURE, handleStopConsoleCapture),
     typedHandle(CHANNELS.WEBVIEW_CLEAR_CONSOLE_CAPTURE, handleClearConsoleCapture),
     typedHandle(CHANNELS.WEBVIEW_GET_CONSOLE_PROPERTIES, handleGetConsoleProperties),
-    // @ts-expect-error: result type contains {success} | null — pending migration to throw AppError. See #6020.
     typedHandleWithContext(CHANNELS.WEBVIEW_OAUTH_LOOPBACK, handleOAuthLoopback as never),
     typedHandle(CHANNELS.WEBVIEW_OAUTH_LOOPBACK_CANCEL, handleCancelOAuthLoopback),
     typedHandle(CHANNELS.WEBVIEW_RELOAD_IGNORING_CACHE, handleReloadIgnoringCache),
--- a/electron/ipc/handlers/webview.ts
+++ b/electron/ipc/handlers/webview.ts
@@ -913,12 +913,16 @@ export function registerWebviewHandlers(_deps: HandlerDependencies): () => void 
               requestId: p.requestId,
               postData: Buffer.from(rewrittenBody).toString("base64"),
             })
+            .then(() => {
+              clearTimeout(timeout);
+              finishIntercept();
+            })
             .catch((err) => {
               logError("OAuth CDP token-exchange continueRequest failed", err, { panelId });
+              navigationError = err instanceof Error ? err : new Error(String(err));
+              clearTimeout(timeout);
+              finishIntercept();
             });
-
-          clearTimeout(timeout);
-          finishIntercept();
         };
 
         wc.debugger.on("message", interceptorListener);
--- a/electron/ipc/handlers/webview.ts
+++ b/electron/ipc/handlers/webview.ts
@@ -2,9 +2,16 @@ import { BrowserWindow, webContents } from "electron";
 import { getWindowForWebContents } from "../../window/webContentsRegistry.js";
 import { CHANNELS } from "../channels.js";
 import { getWebviewDialogService } from "../../services/WebviewDialogService.js";
-import { broadcastToRenderer, sendToRenderer, typedHandle } from "../utils.js";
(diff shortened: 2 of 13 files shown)


gregpriday:
Closing to redo.

PR #8711, opened 2026-05-20
refactor(motion): make anti-flicker threshold a single CSS/TS source of truth

### LINKED ISSUE #8657: Make the anti-flicker threshold a single CSS/TS source of truth
The 400ms anti-flicker threshold is written as a literal in two places in `src/index.css` and again as the `UI_DOHERTY_THRESHOLD` constant in `src/lib/animationUtils.ts`. The CSS and TypeScript copies are kept equal only by hand-written "keep in sync" comments, so they can silently drift.

In CSS it's the `animation-delay` on `.animate-pulse-delayed` and the `transition` on `.surface-stale` / `.palette-results-stale`:

```css
.animate-pulse-delayed {
  animation: pulse-delayed 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;
  animation-delay: 400ms;
}
```

Desired end state: move the threshold into the Tailwind v4 `@theme` block in `index.css` as a CSS custom property, and have the CSS rules reference it through `var()` instead of the hardcoded number. The `@theme` block already does exactly this for the `--duration-*` tokens, so the pattern is established.

Then add a test that parses the `@theme` values out of `index.css` and asserts they match the TypeScript constants, so any future drift fails the build instead of relying on a comment. The same comment-only coupling exists between the `--duration-*` CSS variables and the matching `DURATION_*` TypeScript constants, so the test should cover those too.

The 400ms value itself is not changing. This is purely about making it one source of truth.

[`src/index.css`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/src/index.css#L1124-L1151) · [`src/lib/animationUtils.ts`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/src/lib/animationUtils.ts#L47-L57)

gregpriday:
## Summary

- The 400ms Doherty anti-flicker threshold was duplicated across `src/index.css` (two rule sites: `.animate-pulse-delayed` and `.palette-results-stale` / `.surface-stale`) and `UI_DOHERTY_THRESHOLD` in `src/lib/animationUtils.ts`, kept in sync only by comments
- Adds `--anti-flicker-delay: 400ms` to the `:root` animation-vars cluster as the single CSS source of truth, and updates both CSS rule sites to use `var(--anti-flicker-delay)` instead of bare `400ms` literals
- Adds a drift-contract test that parses `index.css` and asserts: (1) the token is declared once in `:root`, (2) its value matches `UI_DOHERTY_THRESHOLD`, and (3) both CSS rule sites reference `var()` rather than a hardcoded literal

The `UI_DOHERTY_THRESHOLD` TS constant stays — it's still needed for JS consumers (`useDeferredLoading`, inline styles). The test enforces parity between the two representations so any future drift fails the build instead of relying on a comment.

Resolves #8657

## Why `:root` and not `@theme`?

Tailwind v4 auto-generates utility classes for tokens declared in `@theme`. A `--duration-*` or `--anti-flicker-*` token in `@theme` would produce a spurious `duration-*` utility class for what is purely a perceptual-floor constant, not a utility. `:root` keeps it accessible via `var()` without polluting the utility namespace.

## Changes

- `src/index.css` — adds `--anti-flicker-delay: 400ms` to `:root`, replaces both `400ms` animation-delay literals with `var(--anti-flicker-delay)`
- `src/lib/animationUtils.ts` — adds a `"keep in sync with --anti-flicker-delay"` note (TS constant unchanged)
- `src/lib/__tests__/animationUtils.test.ts` — 4 new contract assertions covering declaration, value parity, and both CSS rule sites

## Testing

- `npx vitest run src/lib/__tests__/animationUtils.test.ts` — 18 tests pass (4 new contract cases)
- `npm run check` — all 8 checks pass

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -661,6 +661,11 @@ code.hljs {
   --terminal-animation-duration: 50ms;
   --focus-transition-duration: 150ms;
   --focus-transition-easing: cubic-bezier(0.4, 0, 0.2, 1);
+  /* Anti-flicker (Doherty) gate — single source of truth for the 400ms
+   * perceptual floor below which loading affordances should stay hidden.
+   * Mirror of UI_DOHERTY_THRESHOLD in src/lib/animationUtils.ts; the contract
+   * test in src/lib/__tests__/animationUtils.test.ts enforces parity. */
+  --anti-flicker-delay: 400ms;
   --background: var(--theme-surface-canvas);
   --foreground: var(--theme-text-primary);
   --card: var(--theme-surface-panel);
@@ -1124,7 +1129,7 @@ body,
 .animate-pulse-delayed {
   opacity: 0;
   animation: pulse-delayed 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;
-  animation-delay: 400ms;
+  animation-delay: var(--anti-flicker-delay);
   will-change: opacity;
 }
 
@@ -1143,11 +1148,13 @@ body,
  * (typical for ~258 actions on fast hardware), the dim never starts. On a
  * genuinely overloaded render the surface fades to half opacity as a "still
  * working" cue.
- * Keep in sync with UI_DOHERTY_THRESHOLD in src/lib/animationUtils.ts. */
+ * The 400ms gate lives in the --anti-flicker-delay token in :root above; the
+ * UI_DOHERTY_THRESHOLD constant in src/lib/animationUtils.ts mirrors it for
+ * JS consumers. */
 .palette-results-stale,
 .surface-stale {
   opacity: 0.5;
-  transition: opacity 150ms ease-out 400ms;
+  transition: opacity 150ms ease-out var(--anti-flicker-delay);
 }
 
 /* Opt-in shimmer sweep for skeleton bones. Layered on top of the opacity pulse;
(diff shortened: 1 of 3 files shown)


gregpriday:
CI is red on `check (ubuntu-latest)` — 5 vitest unit failures in `electron/services/__tests__/ProcessMemoryMonitor.test.ts` (1) and `electron/window/__tests__/ProjectViewManager.eviction.test.ts` (4).

These failures are **pre-existing on `develop`** and unrelated to this refactor:

- Confirmed the same 5 tests fail on the latest `develop` commit (`d4e863d84`) — see run 26160601323.
- All three most recent CI runs on `develop` are failing on the same suites.
- This PR touches only `src/index.css`, `src/lib/animationUtils.ts`, and `src/lib/__tests__/animationUtils.test.ts` — none of the eviction or memory-monitor code paths.

Local verification of the changed code:
- `npx vitest run src/lib/__tests__/animationUtils.test.ts` → 18 pass (4 new contract cases)
- `npm run check` (typecheck, lint:ratchet, format:check, channels, ipc, help) → all 8 checks pass

The `ProcessMemoryMonitor` / `ProjectViewManager` test breakage should be diagnosed in its own issue; merging this PR is independent of that work.

PR #8715, opened 2026-05-20
feat(fleet): show armed panel count on preset menu items

### LINKED ISSUE #8688: Show armed-count on fleet preset menu items
The state-preset dropdown in `FleetArmingRibbon.tsx` shows text-only labels for each preset (`All waiting — this worktree`, `All in this worktree`, and so on) with no indication of how many panes the preset would actually arm. Users have to commit before seeing the scope. ([preset menu items today](https://github.com/daintreehq/daintree/blob/e1df597dc604a7395339df3f9f71d37f03d8564b/src/components/Fleet/FleetArmingRibbon.tsx#L277-L309).)

The dry-run compute is already running. Hovering or focusing a preset item fires `computePreviewByState` (backed by `computeArmByStateIds` and `collectEligibleIds` in `src/store/fleetArmingStore.ts`) and stores the result in `previewArmedIds` so the matching panes can be tinted amber. The count is sitting in `previewArmedIds.size`; it just never makes it back to the menu item the user is reading.

Surfacing scope at the selection step gives the user a chance to confirm intent before the arming commit, and downstream before the broadcast step where the destructive-tier preview rule kicks in.

gregpriday:
## Summary
- Shows the number of panels each preset would arm directly in the dropdown menu, so users can confirm scope before committing.
- Subscribes the ribbon to the panel store so counts stay reactive as panels change.
- Uses the same `computeArmByStateIds` and `collectEligibleIds` functions already powering the hover preview — the count was already computed, just not surfaced.

Resolves #8688

## Changes
- Added panel store subscriptions in `FleetArmingRibbon` for reactive preset counts
- Computed `presetCounts` object with armed-count for each of the five presets
- Rendered count as a right-aligned muted label next to each preset label in the dropdown

## Testing
- Verified counts match the amber-tinted preview when hovering each preset
- Counts react correctly when panels are added/removed or change state
- "All in this worktree" count factors in idleness threshold correctly

### DIFF
--- a/src/components/Fleet/FleetArmingRibbon.tsx
+++ b/src/components/Fleet/FleetArmingRibbon.tsx
@@ -272,6 +272,8 @@ export function FleetArmingRibbon(): ReactElement | null {
   );
 
   const activeWorktreeId = useWorktreeSelectionStore((s) => s.activeWorktreeId) ?? null;
+  usePanelStore((s) => s.panelIds);
+  usePanelStore((s) => s.panelsById);
 
   const presetCounts = {
     waitingCurrent: computeArmByStateIds("waiting", "current", activeWorktreeId).length,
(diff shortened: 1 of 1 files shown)


gregpriday:
Re-running CI — the 5 failures are in electron/ main-process tests (ProcessMemoryMonitor + ProjectViewManager eviction), unrelated to this renderer-only fleet preset change.

PR #8727, opened 2026-05-20
fix(crash-guard): close slow-flap blind spot in decay window

### LINKED ISSUE #8683: Fix slow-flap blind spot in crash-loop guard decay
A crash cadence just outside the 5-minute rapid window (6+ minutes between unclean launches) never trips the safe-mode counter despite being chronic instability. The current `STABILITY_TIMEOUT_MS` (5 minutes of clean uptime) zeros the entire `launches[]` array on a single tick — so the slow-flap pattern the guard is supposed to detect is the one it silently bypasses.

```typescript
// CrashLoopGuardService.startStabilityTimer (current)
this.stabilityTimer = setTimeout(() => {
  this.stabilityTimer = null;
  try {
    const state = freshState();   // <-- full reset, every entry dropped
    this.writeState(state);
    this.safeMode = false;
    this.relaunchAllowed = true;
    ...
```

The fix needs chronic-but-slow crash cadences to accumulate strikes over wall-clock time. One shape worth weighing: lazy-evaluated decay — compute the drain arithmetic at boot and at crash-record-time from a persisted `lastEvaluatedAt` timestamp, no `setTimeout` involved. For a disk-persisted guard this avoids the timer-vs-disk race and naturally handles long-closed-app gaps that an active timer cannot.

Three guards share the policy semantics and must remain in lockstep:

- [`CrashLoopGuardService.startStabilityTimer`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/CrashLoopGuardService.ts#L144-L158) — disk-persisted, app-level
- [`PtyHostLifecycle.startStabilityTimer`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/pty/PtyHostLifecycle.ts#L564-L573) — in-memory, utility-process
- [`WorkspaceHostProcess.startStabilityTimer`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/WorkspaceHostProcess.ts#L726-L735) — in-memory, utility-process

The constants are intentionally duplicated rather than shared (see the source comments at the top of `PtyHostLifecycle.ts` and `WorkspaceHostProcess.ts` — the layered-independence design is deliberate). **Don't extract a shared module.** The right outcome is that the next person tuning one file's constants is forced to consider the other two — via documentation, a test that fails on misalignment, or both. Implementation's call which mechanism.

A fourth guard, [`GpuCrashMonitorService`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/GpuCrashMonitorService.ts#L82-L96), uses a different intentional shape — inline window decay, no separate stability timer, soft-then-nuclear escalation. It lives outside the alignment set and stays there.

### Constraints

- Three guards in lockstep after the fix, **no shared-module consolidation** of constants.
- `GpuCrashMonitorService` is intentionally excluded from the alignment set.
- Schema for the disk-persisted state likely needs a new field (or equivalent); existing state files must read cleanly via a backward-compat shim.
- Adversarial test coverage should include a reliably-paced slow-flap scenario (e.g., a 6-minute cadence eventually accumulates strikes).

gregpriday:
## Summary

The three crash-loop guards (`CrashLoopGuardService`, `PtyHostLifecycle`, `WorkspaceHostProcess`) had a 5-minute stability timer that fully reset the crash counter on a single tick. A crash cadence just outside that window — 6+ minutes between unclean launches — accumulated zero strikes despite being chronic instability. Closes #8683.

Two coordinated changes across all three guards:

- **Delete the proactive stability-timer reset.** The per-crash filter (`crashTimestamps.filter(t => now - t < CRASH_WINDOW_MS)`) already implements lazy decay at crash-record-time. The `setTimeout`-driven wholesale reset was both redundant and the cause of the blind spot.
- **Widen the decay window from 5 minutes to 30 minutes** (`CRASH_WINDOW_MS = 30 * 60 * 1000`). With the prior 5-minute window, removing the timer alone wouldn't have helped — a 6-minute cadence still expires each entry before the next crash lands. The 30-minute window keeps three 6-minute-spaced strikes inside the window and trips the cap at strike 3.

The constants stay duplicated across the three files per the existing layered-independence design. A new alignment test (`crashGuardAlignment.test.ts`) imports `CRASH_WINDOW_MS` from each guard and asserts the three values match — the test-on-misalignment mechanism the issue called for so future tuning forces consideration of all three.

`GpuCrashMonitorService` is excluded from the alignment set (its inline-decay shape is intentional and predates this fix).

## Test plan

- [x] `npx vitest run electron/services/__tests__/CrashLoopGuardService.test.ts electron/services/__tests__/CrashLoopGuardService.adversarial.test.ts electron/services/__tests__/crashGuardAlignment.test.ts electron/services/__tests__/WorkspaceHostProcess.test.ts electron/services/pty/__tests__/PtyHostLifecycle.test.ts` — 111 tests pass
- [x] New slow-flap regression coverage:
  - `CrashLoopGuardService` — four 6-minute-cadence boots accumulate to safe mode (was: zero strikes under the prior window)
  - `PtyHostLifecycle` — three 6-minute-cadence crashes emit `onMaxRestartsReached`
  - `WorkspaceHostProcess` — three 6-minute-cadence crashes emit `host-crash`
- [x] `npm run check` clean (typecheck + lint + format)
- [x] Production build succeeds

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -413,7 +413,6 @@ if (!gotTheLock) {
         }
       }
       await createWindow(undefined, lastActiveProjectId ?? undefined);
-      getCrashLoopGuard().startStabilityTimer();
     } catch (error) {
       console.error("[MAIN] Startup failed:", error);
       // Startup crashes hard-exit without running before-quit, which means
--- a/electron/services/CrashLoopGuardService.ts
+++ b/electron/services/CrashLoopGuardService.ts
@@ -6,8 +6,13 @@ import { resilientAtomicWriteFileSync } from "../utils/fs.js";
 const STATE_FILENAME = "crash-loop-state.json";
 const CRASH_THRESHOLD = 3;
 const HARD_STOP_THRESHOLD = 5;
-const STABILITY_TIMEOUT_MS = 5 * 60 * 1000;
-const RAPID_CRASH_WINDOW_MS = 300_000;
+// Sliding-window decay for the crash counter. Lazy-evaluated at boot and at
+// crash-record-time — no setTimeout, no proactive reset. A wider window
+// (compared to the prior 5-minute rapid-crash setting) closes the slow-flap
+// blind spot where a crash every 6+ minutes accumulated no strikes because
+// each entry expired before the next crash. Exported so the alignment test
+// (`crashGuardAlignment.test.ts`) can assert the three guards share one value.
+export const CRASH_WINDOW_MS = 30 * 60 * 1000;
 
 // Grace window before a leftover `.tmp` from `resilientAtomicWriteFileSync` is
 // swept. Matches the precedent in `terminalSessionPersistence.ts` — long enough
@@ -56,7 +61,6 @@ export class CrashLoopGuardService {
   private statePath: string;
   private safeMode = false;
   private relaunchAllowed = true;
-  private stabilityTimer: ReturnType<typeof setTimeout> | null = null;
   private initialized = false;
   private crashCount = 0;
   private lastCrashAt: number | undefined;
@@ -82,7 +86,7 @@ export class CrashLoopGuardService {
     const now = Date.now();
 
     if (!state.cleanExit) {
-      const recentLaunches = state.launches.filter((ts) => now - ts < RAPID_CRASH_WINDOW_MS);
+      const recentLaunches = state.launches.filter((ts) => now - ts < CRASH_WINDOW_MS);
       state.crashes = recentLaunches.length;
     } else {
       state.crashes = 0;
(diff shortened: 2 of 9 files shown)


gregpriday:
CI is failing on `check (ubuntu-latest)` with 5 pre-existing test failures in `electron/window/__tests__/ProjectViewManager.eviction.test.ts` (4) and `electron/services/__tests__/ProcessMemoryMonitor.test.ts` (1). Verified these reproduce on develop HEAD (`d4e863d84`) with my changes reverted — they are not caused by this PR. Worth tracking as a separate issue; the merge gate is currently broken regardless of incoming work.

Tests touched by this PR (5 crash-guard suites, 111 tests) all pass locally and would pass in CI if the gate weren't already red.

gregpriday:
Rebased onto current develop (which includes the fixes for the pre-existing `ProjectViewManager` / `ProcessMemoryMonitor` failures). CI now green on `1414a3b93`: build, check (ubuntu-latest), and ci-ok all pass.

PR #8731, opened 2026-05-20
feat(panels): prioritize last-active panel per worktree on restore

### LINKED ISSUE #8703: Prioritize last-active panel per worktree on restore
Saved panels restore in three phases: non-PTY concurrent, priority PTY sequential, background PTY staggered. The priority predicate in [`panelRestorePhase.ts`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/src/utils/stateHydration/panelRestorePhase.ts#L107-L109) is currently `activeWorktreeId === savedWorktreeId` equality, so only panels in the worktree the user last had focused get the fast tier. The panel they were last looking at in every *other* worktree falls into the slow staggered tier.

Those are the panels they're most likely to switch to next. They should be in the priority set too.

Chrome's `session_restore.cc` loads the focused tab of the active window immediately. Firefox's `restore_on_demand` (the default) eagerly loads the previously-active tab in **each** restored window, not just the most-recently-focused one. The same generalisation maps onto the worktree model: the priority set becomes the saved-active-worktree panels plus the highest-`lastActiveAt` panel from every other worktree.

Data-model piece: [`TerminalState`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/shared/types/ipc/terminal.ts#L57) has `createdAt` but no `lastActiveAt`, and nothing in the panel store writes a per-panel last-active timestamp today. [`AppStateTerminalEntrySchema`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/schemas/ipc.ts#L45-L100) is `.passthrough()`d, so an optional `lastActiveAt: number` field is forward/backward compatible: old snapshots read into a newer build don't fail, and a snapshot written by a newer build still parses fine in an older one.

gregpriday:
## Summary

- Adds an optional `lastActiveAt` timestamp to `TerminalState`, `SavedPanel`, and `SavedPanelState` types (forward/backward compatible via `.passthrough()`)
- `terminalFocusSlice` writes the stamp on focus; `GridTabGroup` reads it from the store and stamps serialized snapshots
- `panelRestorePhase` priority predicate now promotes the highest-`lastActiveAt` panel from every non-active worktree into the priority tier, so each worktree's most-recently-viewed panel gets an eager restore instead of falling into the slow staggered tier

Resolves #8703

## Changes

- `shared/types/ipc/terminal.ts`, `shared/types/panel.ts`, `shared/types/project.ts` — optional `lastActiveAt: number` field on the three relevant types
- `src/store/slices/terminalFocusSlice.ts` + `src/store/slices/panelRegistry/core.ts` / `types.ts` — write timestamp on focus, surface it through the registry
- `src/components/Terminal/GridTabGroup.tsx` — read `lastActiveAt` from the store and include it in serialized panel snapshots
- `src/panels/terminal/serializer.ts` — round-trip the field through serialization
- `src/utils/stateHydration/panelRestorePhase.ts` — extend the priority predicate to include the best-`lastActiveAt` panel per non-active worktree
- New unit tests for the focus slice, serializer, and restore phase

## Testing

Unit tests added for all three layers (focus slice, serializer, restore phase) and all pass. The restore-phase tests cover the cross-worktree promotion logic directly, including the tie-breaking between multiple candidates and the fallback when no `lastActiveAt` is present.

### DIFF
--- a/shared/types/ipc/terminal.ts
+++ b/shared/types/ipc/terminal.ts
@@ -122,6 +122,12 @@ export interface TerminalState {
    * when its registration is gone.
    */
   pluginId?: string;
+  /**
+   * Timestamp (ms) of the last user-initiated focus on this panel. Used by
+   * panel restore to promote the most-recently-active panel per worktree to
+   * the priority restore tier.
+   */
+  lastActiveAt?: number;
 }
 
 /** Terminal data payload for IPC */
--- a/shared/types/panel.ts
+++ b/shared/types/panel.ts
@@ -600,6 +600,12 @@ export interface TerminalInstance {
    * when its registration is gone.
    */
   pluginId?: string;
+  /**
+   * Timestamp (ms) of the last user-initiated focus on this panel. Used by
+   * panel restore to promote the most-recently-active panel per worktree to
+   * the priority restore tier.
+   */
+  lastActiveAt?: number;
   // Note: Tab membership is now stored in TabGroup objects, not on panels
 }
 
--- a/shared/types/project.ts
+++ b/shared/types/project.ts
@@ -133,6 +133,12 @@ export interface PanelSnapshot {
    * when its registration is gone.
    */
   pluginId?: string;
+  /**
+   * Timestamp (ms) of the last user-initiated focus on this panel. Used by
+   * panel restore to promote the most-recently-active panel per worktree to
+   * the priority restore tier.
+   */
+  lastActiveAt?: number;
   // Note: Tab membership is now stored in ProjectState.tabGroups, not on terminals
 }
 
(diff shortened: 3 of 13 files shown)


gregpriday:
CI is failing in `electron/window/__tests__/ProjectViewManager.eviction.test.ts` (4 cases) and `electron/services/__tests__/ProcessMemoryMonitor.test.ts` (1 case) — these failures **also reproduce on `develop` HEAD (0abb93e01)** locally with the exact same errors, so they are pre-existing and not caused by this PR. The failures are isolated to recently-changed memory/projectview test files (touched in 86a9e7b90, bd91fbf37, 766777d50, 1e4258640, ac5f035b6) and have no overlap with the renderer panel restore code touched here.

The unit tests for the actual PR scope (panelRestorePhase, terminalFocusSlice, serializer — 123 cases) all pass locally, as does `npx tsc --noEmit` and the full local `npm run check` + `npm run build` pipeline (clean).

PR #8742, opened 2026-05-20
fix(crash-loop): quarantine corrupt state and sweep stale temp files

### LINKED ISSUE #8684: Quarantine corrupt crash-loop state and sweep stale temp files
Two failure modes around `crash-loop-state.json`:

**1. Silent fallback on corruption.** When the file is unreadable or fails schema validation, [`CrashLoopGuardService.readState`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/CrashLoopGuardService.ts#L167-L190) returns `freshState()` with a single `console.warn`. A corruption that itself caused the crash loop (disk full mid-write, partial write from SIGKILL, antivirus interference) becomes invisible the moment it matters most.

**2. Temp-file leak.** The atomic write uses `crash-loop-state.json.{timestamp-random}.tmp` (see [`generateTempPath`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/utils/fs.ts#L45-L48)) then renames over the target. If the process is killed between the write and the rename (SIGKILL, power loss), the temp file leaks — the `try/catch unlinkSync(tempPath)` cleanup only runs on JS-level throws.

Both shapes already have battle-tested patterns elsewhere in the codebase:

- [`quarantineCorruptConfig` in `electron/store.ts`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/store.ts#L504-L517) — renames the corrupt file to `.corrupted.${Date.now()}`, tightens permissions, returns the path so the caller can surface it.
- [`pendingSettingsRecovery` in `electron/store.ts`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/store.ts#L645-L655) and [`ProjectStateManager.quarantinedPath`](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/ProjectStateManager.ts#L18-L25) — established renderer-facing surface for quarantine events.
- [`terminalSessionPersistence` stale-tmp sweep](https://github.com/daintreehq/daintree/blob/36ba84173ba0269d028d9161d73895a2732a8787/electron/services/pty/terminalSessionPersistence.ts#L322-L364) — TTL-gated boot-time cleanup of atomic-write residue.

The work is applying those patterns to the crash-loop state file. The user-visible outcome of quarantining (rather than silently resetting) is that a corruption coinciding with safe-mode entry can be diagnosed instead of hidden.

### Constraints

- The atomic-write primitive (`resilientAtomicWriteFileSync` in `electron/utils/fs.ts`) is correct as-is. Don't touch it.
- Don't quarantine on missing-file — ENOENT is the legitimate first-run case.
- Any renderer notification must respect CLAUDE.md's `notify()` four-question gate. A reasonable gate is `safeMode === true && quarantined === true` — silent corruption with no safe-mode trip is a log-only event.
- Quarantine prune (keep N most-recent `.corrupted.*` siblings, delete older) runs at boot, not on read — debugging value depends on retention.

gregpriday:
## Summary

Fixes #8684.

- `CrashLoopGuardService.readState()` now **quarantines** a corrupt or schema-invalid `crash-loop-state.json` to `.corrupted.<ts>` instead of silently resetting it via `freshState()`. Mirrors the existing `quarantineCorruptConfig` pattern in `electron/store.ts` (rename + best-effort `chmod 0o600` on non-Windows).
- `initialize()` now **sweeps stale `.tmp` residue** (5-minute TTL) left behind when `resilientAtomicWriteFileSync` is SIGKILL'd between write and rename, and **prunes `.corrupted.*` siblings** beyond `N=3` so a chronic crash loop can't fill the userData directory.
- New `CrashLoopStateRecovery` field on `HydrateResult`, surfaced **only when `safeMode === true` AND a quarantine occurred** — silent corruption that didn't trip safe mode stays log-only per the `notify()` four-question gate.
- Renderer surface: `dispatchRecoveryNotifications` gets a third branch matching the existing project-state pattern (high-priority warning toast, `duration: 0`).
- Hardened number validation: `Number.isFinite()` guards on `crashes` and `lastReset` so JSON-overflow `Infinity` (`1e309 → Infinity`) is caught immediately instead of corrupting the file one boot later when `JSON.stringify` serializes it as `null`.

Plan ordering: sweep + prune run **before** `readState()`, so the current boot's quarantine file is never pruned in the same run. `isSafeModeActive()` (the pre-`app.whenReady` standalone) intentionally stays read-only — letting it also quarantine would race with the service's `readState()`.

## Test plan

- [x] Unit: 10 new cases in `CrashLoopGuardService.test.ts`
- [x] Adversarial: 5 new cases in `CrashLoopGuardService.adversarial.test.ts`
- [x] IPC: 3 new cases in `app.state.test.ts` covering the `safeMode && quarantined` gate
- [x] Renderer: 3 new cases in `recoveryNotifications.test.ts`
- [x] `npm run check` — typecheck, lint, format, ipc checks all clean
- [ ] Manual: corrupt the on-disk `crash-loop-state.json`, restart, confirm `.corrupted.<ts>` sibling appears and app boots normally

### DIFF
--- a/electron/ipc/__tests__/app.state.test.ts
+++ b/electron/ipc/__tests__/app.state.test.ts
@@ -71,6 +71,7 @@ const crashGuard = {
   isSafeMode: vi.fn(() => false),
   getCrashCount: vi.fn(() => 0),
   getLastCrashTimestamp: vi.fn(() => null),
+  getQuarantinedStatePath: vi.fn(() => null),
   resetForNormalBoot: vi.fn(),
 };
 
--- a/electron/ipc/__tests__/app.state.test.ts
+++ b/electron/ipc/__tests__/app.state.test.ts
@@ -320,7 +320,9 @@ describe("app:boot handler", () => {
 
   it("surfaces crashLoopStateRecovery only when in safe mode AND quarantine path is set", async () => {
     crashGuard.isSafeMode.mockReturnValue(true);
-    crashGuard.getQuarantinedStatePath.mockReturnValue("/tmp/crash-loop-state.json.corrupted.42");
+    (crashGuard.getQuarantinedStatePath as any).mockReturnValue(
+      "/tmp/crash-loop-state.json.corrupted.42"
+    );
 
     const result = (await invokeBoot()) as { crashLoopStateRecovery: unknown };
     expect(result.crashLoopStateRecovery).toEqual({
@@ -330,7 +332,9 @@ describe("app:boot handler", () => {
 
   it("omits crashLoopStateRecovery when not in safe mode (silent corruption is log-only)", async () => {
     crashGuard.isSafeMode.mockReturnValue(false);
-    crashGuard.getQuarantinedStatePath.mockReturnValue("/tmp/crash-loop-state.json.corrupted.42");
+    (crashGuard.getQuarantinedStatePath as any).mockReturnValue(
+      "/tmp/crash-loop-state.json.corrupted.42"
+    );
 
     const result = (await invokeBoot()) as { crashLoopStateRecovery: unknown };
     expect(result.crashLoopStateRecovery).toBeNull();
--- a/electron/ipc/__tests__/app.state.test.ts
+++ b/electron/ipc/__tests__/app.state.test.ts
@@ -318,6 +318,32 @@ describe("app:boot handler", () => {
     expect(result.crashPending).toBeNull();
   });
 
+  it("surfaces crashLoopStateRecovery only when in safe mode AND quarantine path is set", async () => {
+    crashGuard.isSafeMode.mockReturnValue(true);
+    crashGuard.getQuarantinedStatePath.mockReturnValue("/tmp/crash-loop-state.json.corrupted.42");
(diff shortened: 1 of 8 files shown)


gregpriday:
ci-ok is red, but the 5 failing tests are **pre-existing on develop** (`d4e863d84`) in code this PR does not touch — `ProcessMemoryMonitor.test.ts` (aggregate working-set pressure escalation) and `ProjectViewManager.eviction.test.ts` (LRU stamp + telemetry cold-switch + memory sampling). Verified via diff: this PR only changes 8 files, all crash-loop-related. Targeted vitest on the changed files (CrashLoopGuardService + IPC handler + renderer notifications) is **99/99 passing**.

This branch's typecheck, lint, format, ipc checks, and build are all green.

PR #8866, opened 2026-05-23
fix(terminal): defer post-WebGL-attach refresh past xterm's paused renderer

### LINKED ISSUE #8864: Bulk-opening agent terminals leaves second pane blank
Opening 4 agent terminals in quick succession leaves the second one blank. The first, third, and fourth render correctly.

Commit bae8344c1 added a synchronous `terminal.refresh(0, rows - 1)` call inside the addon-webgl attach path in `src/services/terminal/TerminalWebGLManager.ts` ([lines 734-740](https://github.com/daintreehq/daintree/blob/ec758ba262efd9ac0be1b77ec50db56fa5fce487/src/services/terminal/TerminalWebGLManager.ts#L734-L740)), gated on `managed.isOpened`, `managed.isVisible`, and `rows > 0`. The bug still reproduces with that fix in place.

The root cause is xterm 6.0's own intersection-based pause. `RenderService` registers an `IntersectionObserver` on `.xterm-screen` with `threshold: 0` (`node_modules/@xterm/xterm/src/browser/services/RenderService.ts` lines 119 and 127). When the screen element's intersection ratio hits 0, line 134 flips `_isPaused = true`. While paused:

```ts
public refreshRows(start: number, end: number, isRedrawOnly: boolean = false): void {
  if (this._isPaused) {
    this._needsFullRefresh = true;
    return;
  }
  ...
}
```

The renderer swap that addon-webgl triggers via `setRenderer` (line 242) routes through `_fullRefresh` at line 250 and hits the same gate. Both the swap-time refresh and the explicit refresh after `loadAddon` are absorbed silently.

The app's own IntersectionObserver in [`src/components/Terminal/TerminalPane.tsx:568`](https://github.com/daintreehq/daintree/blob/ec758ba262efd9ac0be1b77ec50db56fa5fce487/src/components/Terminal/TerminalPane.tsx#L568) can flip `managed.isVisible` to false during the same grid layout settle window, arming the 500ms `webGLHideTimer` in `TerminalInstanceService`. When that timer fires it releases the WebGL addon while xterm still has `_needsFullRefresh` queued. The deferred refresh then runs against a transitioning renderer and short-circuits at RenderService line 172 (`if (!this._renderer.value) return;`).

T1 lands before the layout perturbation. T3 and T4 land after `.xterm-screen` settles into intersection. T2 is the only pane whose xterm-pause / WebGL-release / resume window overlaps.

The recent fix guards on `managed.isVisible`, but that is the service-level visibility flag, not xterm's screen-element intersection. The two gates can disagree, which is what is happening here.

The unit test at [`src/services/terminal/__tests__/TerminalInstanceService.webglVisibility.test.ts:239-253`](https://github.com/daintreehq/daintree/blob/ec758ba262efd9ac0be1b77ec50db56fa5fce487/src/services/terminal/__tests__/TerminalInstanceService.webglVisibility.test.ts#L239-L253) exercises the debounced `setVisible(true)` restore path. The cold-mount attach path that bulk-open uses is uncovered.

gregpriday:
## Summary

- The synchronous `terminal.refresh()` call fired after the WebGL addon attaches was being silently absorbed by xterm's `RenderService` — when `_isPaused` is true (set by xterm's own `IntersectionObserver` on `.xterm-screen`), `refreshRows` sets `_needsFullRefresh` and returns immediately
- This defers the refresh by two `requestAnimationFrame` ticks so it lands after xterm's `IntersectionObserver` has cleared the paused flag and the renderer is ready to accept the call
- The second pane in a bulk-open sequence is the one whose xterm pause/resume window overlaps with the WebGL attach timing, which is why it was the one rendering blank

Resolves #8864

## Changes

- `src/services/terminal/TerminalWebGLManager.ts` — replace synchronous post-attach `terminal.refresh()` with a double-rAF deferred call; cancel token tracked on `managed` so it is cleaned up if the terminal is torn down before the frames fire
- `src/services/terminal/__tests__/TerminalWebGLManager.test.ts` — 11 new tests covering the deferred refresh timing window, cancellation on teardown, and all guard paths (not opened, not visible, zero rows)

## Testing

Unit tests added directly covering the bug path. The 11 new cases in `TerminalWebGLManager.test.ts` exercise the double-rAF deferral, guard conditions, and cancellation behaviour on early teardown.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -179,7 +179,15 @@ jobs:
           }
 
       - name: Check out repository
-        uses: actions/checkout@v6
+        shell: bash
+        run: |
+          set -euo pipefail
+          git init "$GITHUB_WORKSPACE"
+          git -C "$GITHUB_WORKSPACE" remote add origin "https://github.com/${GITHUB_REPOSITORY}.git"
+          git -C "$GITHUB_WORKSPACE" fetch --no-tags --prune --no-recurse-submodules --depth=1 origin "$GITHUB_SHA"
+          git -C "$GITHUB_WORKSPACE" checkout --force FETCH_HEAD
+          git -C "$GITHUB_WORKSPACE" clean -ffdx
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
       - name: Cache Playwright browsers
         id: cache-playwright
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -33,7 +33,16 @@ jobs:
     runs-on: ubuntu-22.04
     timeout-minutes: 15
     steps:
-      - uses: actions/checkout@v6
+      - name: Check out repository
+        shell: bash
+        run: |
+          set -euo pipefail
+          git init "$GITHUB_WORKSPACE"
+          git -C "$GITHUB_WORKSPACE" remote add origin "https://github.com/${GITHUB_REPOSITORY}.git"
+          git -C "$GITHUB_WORKSPACE" fetch --no-tags --prune --no-recurse-submodules --depth=1 origin "$GITHUB_SHA"
+          git -C "$GITHUB_WORKSPACE" checkout --force FETCH_HEAD
+          git -C "$GITHUB_WORKSPACE" clean -ffdx
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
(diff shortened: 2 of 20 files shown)


gregpriday:
Superseded by #8884. The fix changed from a WebGL repaint retry to terminal startup backpressure, so this PR's history/title no longer match the review surface.

PR #8875, opened 2026-05-23
fix: PowerShell call operator for CLI path invocation + agent-preset-icon-color E2E Windows fixes

### LINKED ISSUE #8873: E2E agent-preset-icon-color fails on Windows — PowerShell cannot invoke double-quoted CLI path, bash VAR=val syntax
**Platform:** Windows 11 (win32), PowerShell terminal
**Test:** \`e2e/core/core-agent-preset-icon-color.spec.ts\`

## Error (before fix)

\`\`\`
"C:\...\daintree-e2e-agent-preset-icon-color-i9QyMf\.e2e-bin\claude.cmd"
C:\...\claude.cmd
PS C:\...>
\`\`\`

The agent process exits immediately with no FAKE_CLAUDE_* output. Test times out waiting for \`FAKE_CLAUDE_MANUAL=\`.

## Root causes

**1. \`resolveAgentLaunchBaseCommand\` produces a cmd.exe-quoted path, not a PowerShell invocation.**

On Windows, \`escapeShellArgOptional(resolvedPath)\` wraps the absolute CLI path in double quotes: \`"C:\path\claude.cmd"\`. In PowerShell a bare double-quoted string is a string expression, not a command. The path was echoed to the terminal and the process exited. The fix is to prefix with the PowerShell call operator: \`& "C:\path\claude.cmd"\`.

**2. \`probePrependedCliPath\` resolved the raw shebang script before \`claude.cmd\`.**

On Windows, \`fs.access(path, X_OK)\` succeeds for any readable file regardless of the execute bit. The no-extension probe for \`claude\` found the plain Node script before \`claude.cmd\`, so \`resolvedPath\` pointed to a file PowerShell could not execute. Fixed by writing the implementation as \`claude.js\` on Windows so the no-extension probe misses it; the \`.cmd\` wrapper wins.

**3. Inline \`VAR=val command\` and bash single-quote syntax used in \`shellInvoke\` call.**

\`DAINTREE_E2E_MANUAL_RUN=1 'C:\...\claude.cmd'\` is bash-only syntax. In PowerShell, env vars are set with \`$env:VAR='val'\` and executables are invoked with \`& 'path'\`.

## Fix

**\`src/hooks/useAgentLauncher.ts\` — \`resolveAgentLaunchBaseCommand\`:**
When on Windows and the resolved path is not safe-unescaped, prefix the double-quoted path with \`& \` so PowerShell treats it as a command invocation.

**\`e2e/core/core-agent-preset-icon-color.spec.ts\`:**
- \`writeFakeAgent\`: use \`claude.js\` extension on Windows (avoids no-extension probe collision with \`claude.cmd\`)
- Added \`shellInvoke(agentId, env)\` helper that emits \`$env:K='V'; & 'path.cmd'\` on Windows and \`K=val 'path'\` on POSIX
- \`fakeAgentCommand\` call site replaced with \`shellInvoke\`

dev-en-m:
Fixes #8873

## Summary

- **`src/hooks/useAgentLauncher.ts`** — `resolveAgentLaunchBaseCommand`: on Windows, prefix double-quoted resolved CLI paths with `& ` so PowerShell treats them as invocations (`& "C:\path\claude.cmd"`) rather than string expressions. Without this, `agent.launch` wrote the path to the PTY but PowerShell just echoed it and returned to the prompt.

- **`e2e/core/core-agent-preset-icon-color.spec.ts`**:
  - `writeFakeAgent`: write implementation as `claude.js` on Windows so the no-extension probe in `probePrependedCliPath` misses it and `claude.cmd` wins (on Windows `fs.access(X_OK)` succeeds for any readable file, so the no-extension probe was resolving the raw shebang script before the `.cmd` wrapper)
  - Added `shellInvoke(agentId, env)` cross-platform helper: emits `$env:K='V'; & 'path.cmd'` on Windows and `K=val 'path'` on POSIX
  - Replaced the old `DAINTREE_E2E_MANUAL_RUN=1 fakeAgentCommand(...)` call (bash inline-env + single-quote syntax) with `shellInvoke`

- **`src/hooks/__tests__/useAgentLauncher.commandResolution.test.ts`**: updated the "quotes resolved paths" test case to be platform-aware (Windows expects `& "..."`, POSIX expects `'...'`)

## Test plan

- [ ] `npx vitest run src/hooks/__tests__/useAgentLauncher.commandResolution.test.ts` passes on Windows and Linux
- [ ] `npx playwright test e2e/core/core-agent-preset-icon-color.spec.ts` passes on Windows
- [ ] Same E2E test passes on Linux/macOS (no regression)

### DIFF
--- a/e2e/core/core-agent-preset-icon-color.spec.ts
+++ b/e2e/core/core-agent-preset-icon-color.spec.ts
@@ -55,7 +55,12 @@ async function dispatchAction<T = unknown>(
 }
 
 function writeFakeAgent(agentId: "claude" | "codex"): void {
-  const scriptPath = path.join(fakeBinDir, agentId);
+  // On Windows the raw (no-extension) script file would be picked up by
+  // probePrependedCliPath before claude.cmd because access() succeeds for
+  // any readable file regardless of execute bit. Use a .js extension so the
+  // no-extension probe misses it, then the .cmd wrapper wins.
+  const implName = process.platform === "win32" ? `${agentId}.js` : agentId;
+  const scriptPath = path.join(fakeBinDir, implName);
   const label = agentId.toUpperCase();
   writeFileSync(
     scriptPath,
@@ -109,7 +114,7 @@ function writeFakeAgent(agentId: "claude" | "codex"): void {
   if (process.platform === "win32") {
     writeFileSync(
       path.join(fakeBinDir, `${agentId}.cmd`),
-      [`@echo off`, `node "%~dp0\\${agentId}" %*`, ""].join("\r\n")
+      [`@echo off`, `node "%~dp0\\${agentId}.js" %*`, ""].join("\r\n")
     );
   }
 }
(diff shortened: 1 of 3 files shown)


gregpriday:
Sorry about this. I've had a long-running automated process working on Windows E2E stabilization since earlier today, and everything's now merged to `develop` in [8f4bb40](https://github.com/daintreehq/daintree/commit/8f4bb40075d5c7796bf33f0f0caee64860ca3b31). All Windows E2E tests are passing now.

Your work overlapped with that automation, which I feel bad about. The careful Windows triage was genuinely useful and fed into the overall effort.

Thanks again. Closing.

PR #8885, opened 2026-05-23
fix(e2e): poll buffer API directly in waitForTerminalReady, skip DOM fallback after WebGL swap

### LINKED ISSUE #8881: waitForTerminalReady races WebGL reader registration — DOM fallback returns empty after renderer swap
## Summary

\`waitForTerminalReady\` fails intermittently across 4 tests because the text-present poll falls back to an empty DOM when the WebGL renderer is active but \`__daintreeReadTerminalBuffer\` is not yet registered on the window.

**Affected tests:**
- \`e2e/full/resilience/core-concurrent-operations.spec.ts:53\`
- \`e2e/full/platform/core-multi-window.spec.ts:114\`
- \`e2e/core/core-process-badge.spec.ts:32\`
- \`e2e/full/worktree/core-worktree-cross-boundary.spec.ts:329\`

## Error

\`\`\`
Error: expect(received).toBe(expected) // Object.is equality
Expected: true
Received: false
Timeout 10000ms exceeded while waiting on the predicate
\`\`\`

All four failures originate at \`e2e/helpers/terminal.ts:116\` — the second phase of \`waitForTerminalReady\`.

## Root Cause

\`waitForTerminalReady\` runs two sequential phases:

1. \`waitForTerminalPty\` — polls until \`hasPty === true\` → **passes**
2. Text-present poll via \`getTerminalText\` → **times out**

\`getTerminalText\` tries \`__daintreeReadTerminalBuffer\` first. When it is not a function (not yet registered on the window), it returns \`null\` and falls through to \`panelLocator.locator(SEL.terminal.xtermRows).innerText()\`.

After the DOM→WebGL renderer swap introduced in commits \`7b0d1e312\` + \`bae8344c1\`, the DOM rows element still exists in the tree but xterm.js no longer paints text into it — content is rendered to a WebGL canvas. So \`innerText()\` returns \`""\` on every poll for the full 10 s budget, and the check never resolves \`true\`.

The race: the PTY process can be attached (\`hasPty === true\`) before \`__daintreeReadTerminalBuffer\` is registered on the window. In that gap, every \`getTerminalText\` call hits the empty DOM fallback.

## Fix

In \`waitForTerminalReady\`, replace the \`getTerminalText\` text-present check with a poll that uses **only** the buffer API and treats \`null\` (reader not registered yet) as a continue signal:

\`\`\`ts
// e2e/helpers/terminal.ts — waitForTerminalReady
export async function waitForTerminalReady(
  page: Page,
  panelLocator: Locator,
  timeout = 10_000
): Promise<void> {
  await waitForTerminalPty(page, panelLocator, timeout);

  const panelId = await getPanelId(panelLocator);
  await expect
    .poll(
      () =>
        page.evaluate((id) => {
          const reader = (window as unknown as Record<string, unknown>)
            .__daintreeReadTerminalBuffer;
          if (typeof reader !== "function") return null; // reader not ready yet
          return reader(id) as string | null;
        }, panelId),
      { timeout, intervals: [100, 250, 500] }
    )
    .toSatisfy(
      (text: string | null) => typeof text === "string" && text.trim().length > 0
    );
}
\`\`\`

\`getTerminalText\`'s DOM fallback is left intact for non-WebGL environments. Only \`waitForTerminalReady\` stops using it.

## Related

Renderer swap commits: \`7b0d1e312\`, \`bae8344c1\`

dev-en-m:
Fixes #8881

## Summary

`waitForTerminalReady` was calling `getTerminalText`, which falls back to
`xtermRows.innerText()` when `__daintreeReadTerminalBuffer` is not yet registered.
After the DOM→WebGL renderer swap (`7b0d1e312`, `bae8344c1`), the DOM rows element
exists but xterm.js no longer writes text into it — so the fallback always returns `""`
and the text-present poll times out.

The root race: `hasPty === true` (PTY attached) can be true before
`__daintreeReadTerminalBuffer` is registered on the window. In that window every
call to `getTerminalText` silently returns empty DOM text.

## Change

In `waitForTerminalReady`, replaced the `getTerminalText`-based text-present check
with a direct inline buffer-API poll:

- Returns `null` when the reader is not yet registered → `toSatisfy` keeps polling
- Returns the buffer string once registered → poll resolves when non-empty
- `getTerminalText`'s DOM fallback is left intact for non-WebGL environments

**File:** `e2e/helpers/terminal.ts`

## Test plan

- [ ] `npx playwright test e2e/core/core-process-badge.spec.ts` passes
- [ ] `npx playwright test e2e/full/resilience/core-concurrent-operations.spec.ts` passes
- [ ] `npx playwright test e2e/full/platform/core-multi-window.spec.ts` passes
- [ ] `npx playwright test e2e/full/worktree/core-worktree-cross-boundary.spec.ts` passes

### DIFF
--- a/e2e/helpers/terminal.ts
+++ b/e2e/helpers/terminal.ts
@@ -105,15 +105,22 @@ export async function waitForTerminalReady(
 ): Promise<void> {
   await waitForTerminalPty(page, panelLocator, timeout);
 
+  // Bypass getTerminalText's DOM fallback: after DOM→WebGL swap the rows element
+  // is empty, so null (reader not yet registered) must mean "keep polling", not
+  // "fall through to DOM text".
+  const panelId = await getPanelId(panelLocator);
   await expect
     .poll(
-      async () => {
-        const text = await getTerminalText(panelLocator);
-        return text.trim().length > 0;
-      },
+      () =>
+        page.evaluate((id) => {
+          const reader = (window as unknown as Record<string, unknown>)
+            .__daintreeReadTerminalBuffer;
+          if (typeof reader !== "function") return null;
+          return reader(id) as string | null;
+        }, panelId),
       { timeout, intervals: [100, 250, 500] }
     )
-    .toBe(true);
+    .toSatisfy((text: string | null) => typeof text === "string" && text.trim().length > 0);
 }
 
 async function activateTerminal(page: Page, panelId: string): Promise<void> {


gregpriday:
Hey @dev-en-m, thanks for digging into this so carefully in #8881.

The diagnosis was slightly off in one place. `__daintreeReadTerminalBuffer` is registered as a module-load side effect in `src/services/terminal/TerminalInstanceService.ts`, so it's always on `window` before any PTY attaches. The registration race the PR posits can't actually fire.

The real bug was on the production side: xterm wasn't repainting the buffer after the DOM→WebGL renderer swap, so the buffer reader returned empty. I landed `bae8344c1` separately, which calls `terminal.refresh(0, rows - 1)` immediately after the WebGL addon activates. With that in develop, the four affected tests pass without this helper change.

Closing for that reason. If you'd like to push for the helper change as defensive hardening (treating reader-not-ready as keep-polling rather than falling through to DOM), happy to pick it up in a follow-up issue.

PR #9049, opened 2026-05-24
fix(palette): split typed-input stale budget from Doherty gate

### LINKED ISSUE #8993: Palette stale-state runs on Doherty's discrete-action budget
The palette family runs the typed-input stale-dim AND the skeleton-on-fetch anti-flicker floor off the same 400ms gate. Same token, two genuinely different perceptual budgets, and the typed-input side is taking the wrong number.

`--anti-flicker-delay: 400ms` (mirrored as `UI_DOHERTY_THRESHOLD = 400`) gates three different things:

- `.palette-results-stale` 0.5 opacity dim while the deferred filter pass catches up
- `palette-loading-bar` fade pinned to the bottom of `AppPaletteDialog`'s header
- `.animate-pulse-delayed` skeleton consumers (the legitimate Doherty use case)

400ms is Doherty & Thadani 1982. Their System Response Time was measured Enter-to-first-character on IBM mainframes, i.e. discrete request/response. Not the budget for between-keystroke feedback during continuous typing, where the published numbers run an order of magnitude tighter:

- Card/Moran/Newell + Nielsen: 100ms is the "instantaneous" threshold.
- Google RAIL: 100ms for input response.
- Core Web Vitals INP: under 200ms is "good".

A survey of 11 modern keyboard-first palettes — VS Code, Raycast, Linear, Slack, cmdk, Vercel, Warp, Spotlight, Arc, Fig, Telescope — is unanimous: none of them dim result rows during a filter pass. All of them surface busy state in or near the input. Raycast's right-edge spinner, VS Code's `QuickInput.busy` progress bar above the listbox, Linear's 1px bottom-of-input sweep for slow async queries. The closest precedent for our current pattern is no precedent at all.

## Desired behaviour

Split the palette's typed-input budget from the global Doherty constant. Skeleton consumers (`.animate-pulse-delayed`) want the 400ms floor — sub-400ms fetch work should still show nothing. The palette stale-dim and loading-bar fade want something in the 150–200ms band the surveyed palettes inhabit; 180ms is a clean midpoint with a frame of safety either side. Whatever number lands, the two budgets should be separate tokens, not one mutated value.

## Pair: the 0.5 opacity

While this surface is open: `.palette-results-stale` dims to `opacity: 0.5`, which is harsher than anything in the surveyed palettes. Linear and Vercel's loading/stale dim sits in the 0.7–0.85 range; most of the rest don't dim results at all. Applying 0.5 to already-muted body text probably collapses contrast below WCAG 1.4.3 AA in the still-active interpretation — SC 1.4.3 exempts truly-inactive components but not elements that remain selectable. Two defensible landings: dim at ~0.85, or remove the dim entirely and rely on the loading-bar alone (the cmdk default).

## Constraints

- `.animate-pulse-delayed` keeps the 400ms floor. That is the Doherty use case as currently written.
- The `.surface-stale` rule (Review + Commit modal background-refresh) currently shares the `.palette-results-stale` selector. Splitting cleanly means the palette consumer adopts the shorter budget while `.surface-stale` stays on whatever the Review modal needs — likely 400ms unchanged, but worth confirming, not assuming.
- The drift-contract block in `src/lib/__tests__/animationUtils.test.ts` enforces parity between the JS constant and the CSS variable plus the exact `.palette-results-stale` transition signature. Any new sibling token wants the same contract.
- CLAUDE.md's "Loading Indicators" section currently mandates the 400ms gate for all anti-flicker. Needs a clarifying note that typed-input feedback and discrete-action response are separate budgets, so the next person doesn't re-collapse them.

## Evidence

- [`src/index.css#L664-L668`](https://github.com/daintreehq/daintree/blob/f79883b9b9e618c1887aa83c03dee6bca950048c/src/index.css#L664-L668) — the `--anti-flicker-delay` declaration and its comment.
- [`src/index.css#L1171-L1175`](https://github.com/daintreehq/daintree/blob/f79883b9b9e618c1887aa83c03dee6bca950048c/src/index.css#L1171-L1175) — the `.palette-results-stale, .surface-stale` rule.
- [`src/lib/animationUtils.ts#L47-L52`](https://github.com/daintreehq/daintree/blob/f79883b9b9e618c1887aa83c03dee6bca950048c/src/lib/animationUtils.ts#L47-L52) — `UI_DOHERTY_THRESHOLD` with the existing JSDoc.
- [`src/components/ui/AppPaletteDialog.tsx#L257-L272`](https://github.com/daintreehq/daintree/blob/f79883b9b9e618c1887aa83c03dee6bca950048c/src/components/ui/AppPaletteDialog.tsx#L257-L272) — loading-bar with the 400ms `transitionDelay`.
- [`src/lib/__tests__/animationUtils.test.ts#L146-L186`](https://github.com/daintreehq/daintree/blob/f79883b9b9e618c1887aa83c03dee6bca950048c/src/lib/__tests__/animationUtils.test.ts#L146-L186) — the drift-contract block enforcing JS/CSS parity and the exact transition signature.

Related: #8929 and #8948 both name the 400ms Doherty gate as a constraint to preserve. That stands for the skeleton case; this issue is the carve-out for the typed-input case.

gregpriday:
## Summary

- The command palette's stale-state feedback was running on the 400ms Doherty discrete-action gate, which is the wrong budget for continuous typed-input — a typist who pauses expects feedback well within that window, not at the same threshold as a discrete button press.
- Added a dedicated `--palette-stale-delay` CSS token (150ms) and `UI_PALETTE_STALE_DELAY` JS constant, aligned with the Tier 2-fast palette timing. Split the combined `.palette-results-stale`/`.surface-stale` rule: palette rows now dim to `opacity: 0.85` (the previous `0.5` failed WCAG 1.4.3 on keyboard-navigable rows); background-refresh `.surface-stale` surfaces keep the 400ms/0.5 values.
- Migrated `AppPaletteDialog`'s loading-bar `transitionDelay` to the new constant, restructured the JS/CSS drift-contract test into per-token assertions with negative cross-token guards, and added a clarifying note to the CLAUDE.md Loading Indicators rule.

Resolves #8993

## Changes

- `src/index.css` -- add `--palette-stale-delay: 150ms` token; split stale rule into `.palette-results-stale` (150ms, 0.85 opacity) and `.surface-stale` (400ms, 0.5 opacity)
- `src/lib/animationUtils.ts` -- export `UI_PALETTE_STALE_DELAY = 150` constant
- `src/components/ui/AppPaletteDialog.tsx` -- use `UI_PALETTE_STALE_DELAY` for loading-bar `transitionDelay`
- `src/components/ui/SearchablePalette.tsx` -- minor cleanup
- `src/lib/__tests__/animationUtils.test.ts` -- per-token assertions, negative cross-token guards
- `src/components/ui/__tests__/AppPaletteDialogHeader.test.tsx` -- updated for new delay value
- `CLAUDE.md` -- Loading Indicators clarifying note on typed-input vs discrete-action budgets

## Testing

40 unit tests pass. `npm run check` fully green (lint warnings down by 1 from the split rule).


gregpriday:
Closing as superseded. While this was in flight, `c9c8e94c2` landed on `develop` and already closes #8993 with an equivalent split — `UI_PALETTE_STALE_DELAY` + a dedicated palette token (`--anti-flicker-delay-palette`), `.palette-results-stale` moved off the 400ms Doherty gate, `.surface-stale` kept at 400ms/0.5, the `AppPaletteDialog` loading-bar migrated, and the drift-contract test restructured with a negative-recombination assertion. No point maintaining a parallel PR.

The two implementations differ only on judgment calls that aren't worth re-litigating now that develop's is merged:

- **Delay:** develop uses 200ms (Algolia `stalledSearchDelay` / `useDeferredValue` convention); this PR used 150ms (RAIL/INP band, aligned with `UI_PALETTE_ENTER_DURATION`). Both fine.
- **Stale opacity:** develop dims palette rows to `0.6`; this PR used `0.85`. One thing worth a second look: palette rows stay keyboard-navigable during a filter pass, so WCAG 1.4.3 applies to them. At `0.6`, already-muted secondary text can drop below the 4.5:1 AA floor. If that proves true on the darker themes, nudging `.palette-results-stale` toward `0.7–0.85` would be the fix — but that's a follow-up against the merged code, not this PR.

Issue #9229, opened 2026-05-27
Build the inline renderer host for plugin views


gregpriday:
The `contributes.views` schema is accepted by [`ViewContributionSchema`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/schemas/plugin.ts#L55) but logged-and-skipped at plugin load. The renderer-side mount path doesn't exist yet.

The no-PTY branch of `usePluginPanelKinds` at [`src/hooks/usePluginPanelKinds.ts:81-91`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/hooks/usePluginPanelKinds.ts#L81-L91) actively calls `unregisterPanelKindDefinition(config.id)` for every plugin view kind, so every external plugin view falls through to [`PluginMissingPanel`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Panel/PluginMissingPanel.tsx) regardless of whether the owning plugin is loaded.

The host needs to resolve `componentPath`, `React.lazy` the plugin module, and call `registerPanelKindDefinition` with a real component.

### `PluginViewHost` wrapper

Compose the existing `ErrorBoundary` (`variant: "component"`) with `<Suspense>` and discriminate chunk-load failures (`TypeError: Failed to fetch dynamically imported module`) from render errors in the fallback copy. [`BrowserPaneWrapper` at `src/panels/registry.tsx:73-83`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/panels/registry.tsx#L73-L83) is the proven in-repo template for this composition.

### `PanelViewProps` type

Actually export `PanelViewProps` from `shared/types/plugin.ts`:

```ts
export interface PanelViewProps {
  panelId: string;
  disposeSignal: AbortSignal;
  // ...host-provided fields per PanelComponentProps
}
```

[...]

PR #9374, opened 2026-05-27
feat(plugins): register plugin:// protocol handler for static assets

### LINKED ISSUE #9283: Register a `plugin://` protocol handler for plugin static assets
#9229 covers loading the React component bundle for a plugin view, but a plugin view ships more than one file: a code-split JS chunk, a sourcemap, a CSS file, an icon SVG, an image used inside the component. The renderer host needs a stable origin scheme to load all of that from, and it can't be `file://` (path traversal, mixed-content rules in `WebContentsView`, no per-plugin sandboxing) or `app://` (already taken). This issue is the `plugin://` scheme — the prerequisite #9229 loads against.

Register a privileged custom scheme via `protocol.registerSchemesAsPrivileged` before `app.whenReady()` (Electron requires this — late registration silently downgrades the scheme) and implement the handler with `protocol.handle("plugin", …)` once ready resolves. URL shape: `plugin://{pluginId}/{relative/path/inside/the/plugin/dir}`. Resolution: look up the plugin's installed root from `PluginService`, `path.resolve(pluginRoot, requestPath)`, then `fs.realpath` and verify the real path is still under `pluginRoot` — refuse anything that escapes (`../../etc/passwd`, symlink outside the plugin dir). Containment is the entire point of the custom scheme; without it we've reinvented `file://` with a worse name.

Privileged flags: `standard: true` (relative imports inside a bundle resolve), `secure: true` (renderer treats it as a secure context — required for fetch from a `plugin://` origin and ESM imports), `supportFetchAPI: true`, `corsEnabled: true` (each plugin is its own origin via the `{pluginId}` host segment, which is what we want for per-plugin isolation). `bypassCSP: false` — plugins stay subject to CSP.

Content-type from extension: `.js/.mjs` → `application/javascript`, `.css` → `text/css`, `.json/.map` → `application/json`, `.svg` → `image/svg+xml`, common images matched normally, fallback to `application/octet-stream`. No directory listings, no auto-`index.html`.

## Evidence

- [`docs/plugins/contribution-points.md:127-158`](https://github.com/daintreehq/daintree/blob/develop/docs/plugins/contribution-points.md#L127-L158) — views ship pre-built ESM modules via `componentPath`, which #9229 loads against some scheme
- [`docs/plugins/dev-loop.md:128-132`](https://github.com/daintreehq/daintree/blob/develop/docs/plugins/dev-loop.md#L128-L132) — DevTools section names `daintree-plugin:{pluginId}/...` as the source-path scheme; this issue picks the real shipping name
- Electron 41 `protocol.handle` docs — replaced `registerFileProtocol`; the modern handler returns a `Response`/`ReadableStream`

## Constraints

- Privileged registration must happen synchronously before `app.whenReady()` — late registration is a silent footgun.
- `fs.realpath` containment is load-bearing. A `path.relative`-only check trusts the filesystem and fails on symlinks pointing out.
- Handler works for built-ins (`plugins/builtin/{name}/`), user installs (`~/.daintree/plugins/{pluginId}/`), and dev symlinks — the `pluginId → root` lookup goes through `PluginService`.
- Lands before #9229 — the renderer host loads chunked bundles via `plugin://`, so it can't merge against `file://` and then migrate.

## Depends on

- #9274 — `.dntr` archive layout and installed plugin directory structure (resolver needs to know where `{pluginId}` lives on disk)

Under epic #5651.

gregpriday:
## Summary

- Registers `plugin://` as a privileged Electron custom scheme (`standard`, `secure`, `supportFetchAPI`, `corsEnabled`, `bypassCSP: false`) before `app.whenReady()` — late registration silently downgrades the scheme so this has to be synchronous at startup.
- Handler resolves `plugin://{pluginId}/{path}` to disk via `PluginService.getPluginDir`, with `fs.realpath` + `O_NOFOLLOW` containment to block path traversal and symlink escapes out of the plugin directory.
- Registered on both the default session and per-project `persist:daintree` sessions; CORP + trusted-origin ACAO headers set on responses; `plugin:` added to `script-src`, `connect-src`, `img-src`, and `style-src` CSP directives in both prod and dev configs; `.mjs` and `.map` MIME entries included.

Resolves #9283. Prereq for #9229.

## Changes

- `electron/setup/protocols.ts` — scheme registration + handler implementation
- `electron/main.ts` — wires scheme registration before `app.whenReady()`
- `electron/services/PluginService.ts` — exposes `getPluginDir` for the handler
- `shared/config/csp.ts` — adds `plugin:` to relevant directives
- `electron/utils/appProtocol.ts` — `.mjs`/`.map` MIME entries
- Test coverage: `electron/setup/__tests__/protocols.test.ts`, `electron/services/__tests__/PluginService.test.ts`, `shared/config/__tests__/csp.test.ts`

## Testing

All 265 unit tests pass. Full `npm run check` clean (typecheck + lint + format). Traversal and symlink containment covered by adversarial test cases in `protocols.test.ts`.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -21,6 +21,7 @@ import { enforceIpcSenderValidation, setupPermissionLockdown } from "./setup/sec
 import {
   registerAppProtocol,
   registerDaintreeFileProtocol,
+  registerPluginProtocol,
   setupWebviewCSP,
 } from "./setup/protocols.js";
 import {
@@ -114,6 +115,19 @@ protocol.registerSchemesAsPrivileged([
     privileges: {
       secure: true,
       supportFetchAPI: true,
+    },
+  },
+  {
+    // Per-plugin static asset scheme: plugin://{pluginId}/{path}. standard+secure
+    // so relative ESM imports resolve and the renderer treats it as a secure
+    // context; the {pluginId} host segment is a distinct origin per plugin
+    // (corsEnabled). bypassCSP defaults to false — plugins stay subject to CSP.
+    scheme: "plugin",
+    privileges: {
+      standard: true,
+      secure: true,
+      supportFetchAPI: true,
+      corsEnabled: true,
     },
   },
 ]);
@@ -420,6 +434,7 @@ if (!gotTheLock) {
       setupPermissionLockdown();
       registerAppProtocol(distPath);
       registerDaintreeFileProtocol();
+      registerPluginProtocol();
       setupWebviewCSP();
       // Prime the hydrate prefetch cache for the last-active project so the
       // renderer's first `app:boot` invoke resolves as a cache hit instead of
--- a/electron/services/PluginService.ts
+++ b/electron/services/PluginService.ts
@@ -918,6 +918,16 @@ export class PluginService {
     return this.plugins.has(pluginId);
   }
 
+  /**
+   * Resolve a plugin's installed on-disk root directory by id, or null if no
+   * plugin with that id is loaded. Used by the `plugin://` protocol handler to
+   * map `plugin://{pluginId}/...` URLs to files on disk. Looked up at request
+   * time so reinstall/unload is reflected immediately.
+   */
+  getPluginDir(pluginId: string): string | null {
+    return this.plugins.get(pluginId)?.dir ?? null;
+  }
+
   registerHandler(pluginId: string, channel: string, handler: PluginIpcHandler): void {
     if (!this.plugins.has(pluginId)) {
       throw new Error(`Unknown plugin: ${pluginId}`);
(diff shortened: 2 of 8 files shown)


gregpriday:
CI note: the two red unit tests (`electron/__tests__/clipboard.test.ts` — "No \"app\" export" and `src/store/__tests__/persistenceBoundaryHardening.test.ts:614`) are pre-existing failures on `develop` itself (same two failures in develop run 26520148330), not introduced by this PR. Neither file is in this PR's import graph. All `plugin://` protocol tests and the rest of the suite pass.

gregpriday:
Closing — out of Wave 1 scope. Will revisit.

PR #9377, opened 2026-05-27
feat(plugins): implement host.showToast

### LINKED ISSUE #9278: Implement `host.showToast`
The docs advertise `host.showToast({ message, intent })` as the standard surface for plugin-issued toasts (`docs/plugins/host-api.md:232-243`). The `PluginHostApi` definition in `shared/types/plugin.ts:191-262` has no `showToast` member, and `createHost` in `electron/services/PluginService.ts:486-738` does not attach one. Plugins have no supported way to surface a notification today.

## Evidence

- `docs/plugins/host-api.md:53` lists `showToast(options: ToastOptions): Promise<void>` on `PluginHostApi`.
- `docs/plugins/host-api.md:234-241` documents the shape (`title`, `description`, `type`, `durationMs`).
- `shared/types/plugin.ts:191-262` — no `showToast`.
- `electron/services/PluginService.ts:486-738` — `createHost` builds nine methods, none of them `showToast`.
- `src/lib/notify.ts:749-1021` — `notify()` is the only sanctioned entry point for any toast in the app, with the routing, quiet-hours, rate-limit, and inbox-fallback semantics already wired.

## Constraints

- Implementation routes through `notify()`. Do not call the toast store directly — the rate-limit, quiet-hours, suppression-grace, and inbox-write paths all live inside `notify()` and a parallel surface would silently bypass them.
- Plugin toasts namespace as `{pluginId}: {message}` for provenance. Read `pluginId` from the closure on `createHost` — plugins cannot spoof another plugin's id since the host instance is bound at activation.
- Enforce CLAUDE.md's `notify()` rules at the bridge:
  - `priority: "low"` with `type: "error"` is rejected (the lint rule at `eslint.config.js:413` documents why — low-priority errors silently drop the toast). Reject server-side too; plugins don't see the lint rule.
  - `message` accepts string only across the IPC boundary. Plugin code cannot pass a `ReactNode`, so the `inboxMessage`-required rule (`notify.ts:769-776`) collapses — every plugin toast writes a history entry from the string itself.
- The `intent`/`type` field maps directly to `NotificationType` (`error | warning | success | info`); reject unknown values at the schema layer.
- Bridge takes a Zod-validated payload through IPC since the call originates in renderer (when SDK React hooks invoke it) or in main (when the plugin's `activate` code runs there). Both paths must converge on the same validator.
- Audit toast hygiene per CLAUDE.md's four-question checklist applies to plugin authors, not the host — the SDK docs should restate it. The host's job is to deliver, not to second-guess.

## Depends on

- #9269 (SDK export boundary)

## Blocks

- F10b (host contract test harness — toast assertions are the canonical visible side effect)
- F19 (Settings tab uses plugin-issued toasts in its install/enable flow)
- F31 (SDK publish — plugin authors expect `showToast` as a baseline)

Under epic #5651.

gregpriday:
## Summary

- `host.showToast` was documented in the plugin host API but never wired up — this implements it end-to-end
- Adds `PluginToastOptions` type with strict Zod validation (`PluginToastOptionsSchema`); toasts are namespaced as `{pluginId}: {message}` so provenance can't be spoofed
- Rate-limit bucket is scoped per plugin + type; `durationMs` threads through `MainProcessToastPayload` to `notify()`

Resolves #9278

## Changes

- `shared/types/plugin.ts` — `PluginToastOptions` type and `PluginToastOptionsSchema`
- `shared/types/ipc/maps.ts` — `MainProcessToastPayload` extended with `duration`
- `electron/schemas/plugin.ts` — Zod schema for incoming toast options
- `electron/services/PluginService.ts` — `createHost()` implementation: liveness-guarded (mirrors `invalidateFileDecorations`), namespaced message, per-plugin+type rate-limit bucket
- `electron/preload.cts` — `PluginToastOptions` type exposed on preload
- `src/hooks/useMainProcessToastListener.ts` — threads `duration` through to `notify()`
- `electron/services/__tests__/PluginService.test.ts` — 8 new tests covering happy path, validation errors, revoked plugin, rate-limit, duration clamping, message namespacing
- `src/hooks/__tests__/useMainProcessToastListener.test.tsx` — boundary test for `duration` passthrough
- `docs/plugins/host-api.md` + `docs/plugins/getting-started.md` — reconciled with actual implementation

## Testing

Unit tests cover the full contract: valid calls, invalid option shapes, revoked plugins, rate-limit exhaustion, `durationMs` boundary clamping (positive int, max 60s), and message namespace format. Existing `useMainProcessToastListener` tests pass with the extended payload.

### DIFF
--- a/docs/plugins/getting-started.md
+++ b/docs/plugins/getting-started.md
@@ -60,7 +60,7 @@ The `commands[].name` maps to `src/say-hello.ts` by filesystem convention. Its d
 import { showToast } from "@daintreehq/plugin-sdk";
 
 export default async function sayHello() {
-  await showToast({ title: "Hello from my plugin" });
+  await showToast({ message: "Hello from my plugin" });
 }
 ```
 
--- a/docs/plugins/host-api.md
+++ b/docs/plugins/host-api.md
@@ -50,7 +50,7 @@ interface PluginHostApi {
   settings: SettingsApi;
 
   // UI helpers
-  showToast(options: ToastOptions): Promise<void>;
+  showToast(options: PluginToastOptions): Promise<void>;
 }
 ```
 
@@ -233,14 +233,15 @@ Secret-type settings are stored in the OS keychain via `keytar`. They're returne
 
 ```ts
 await host.showToast({
-  title: "Synced",
-  description: "Fetched 12 issues from Linear",
-  type: "success", // "info" | "success" | "warning" | "error"
-  durationMs: 4000,
+  message: "Fetched 12 issues from Linear",
+  type: "success", // "info" | "success" | "warning" | "error" — defaults to "info"
+  durationMs: 4000, // optional; defaults to the app's per-type duration
 });
 ```
 
-Toasts render in Daintree's standard toast container. There's no "sticky" or "action required" toast type — for persistent UI, register a panel view instead.
+The host prefixes `message` with your plugin id (`{pluginId}: {message}`) so users can tell which plugin raised the toast — you don't add the prefix yourself. `message` is a string only; `priority` and action buttons aren't exposed to plugins. An empty message or an unknown `type` rejects.
+
+Toasts route through Daintree's standard `notify()` path, so rate-limit, quiet-hours, and inbox-history semantics apply. Audit your toasts against the four-question checklist (timely, helpful, not already visible, ignorable) — the host delivers what you ask for, it doesn't second-guess. There's no "sticky" or "action required" toast type — for persistent UI, register a panel view instead.
 
 ## React hooks — `@daintreehq/plugin-sdk/react`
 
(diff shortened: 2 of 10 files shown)


gregpriday:
CI note: the red `check (ubuntu-latest)` is **not** caused by this PR. Two tests fail, both pre-existing on `develop` and unrelated to `host.showToast`:

- `electron/__tests__/clipboard.test.ts` — `[vitest] No "app" export is defined on the "electron" mock` (via `ProjectStore.ts:78` ← `clipboard.ts:10`). Deterministic; last touched in `91cb5f01c`, not in this diff.
- `src/store/__tests__/persistenceBoundaryHardening.test.ts` — `expected to be called 1 times, but got 2/3 times`. Flaky under parallel load (count varies run-to-run; passes in isolation).

`develop`'s own latest CI run (`26520148330`) fails on the same two tests, so this branch inherits a red base. This PR's own changes are green: `build` passes and `PluginService.test.ts` (212 tests incl. the new `showToast` cases) passes. Needs a base-branch fix to go fully green.

PR #9378, opened 2026-05-27
fix(ci): rank native macOS ZIPs ahead of universal in update metadata

### LINKED ISSUE #9259: Intel Macs download universal ZIP instead of native x64 on auto-update
Every Intel Mac in our installed base downloads the larger `universal-mac.zip` on auto-update instead of the smaller native `-x64-mac.zip`. Apple Silicon is unaffected.

## Root cause

The `macZipPriority` function in [`scripts/ci/generate-update-metadata.mjs`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/scripts/ci/generate-update-metadata.mjs#L50-L55) ranks `universal-mac.zip` at priority 0, `-x64-mac.zip` at 1, and `-arm64-mac.zip` at 2. The generated `latest-mac.yml` lists `universal-mac.zip` first in the `files` array, and the top-level `path:` points at it too.

In electron-updater 6.x, `MacUpdater.ts` filters mac artifacts by whether the URL contains the substring `arm64`. On Intel Macs both `universal` and `x64` pass that filter, and `findFile` then picks the first surviving entry via `Array.prototype.find`. The universal ZIP wins every time.

Apple Silicon picks the arm64 ZIP correctly because the substring filter excludes anything without `arm64` in the URL, so the choice is unambiguous on that side.

## Locked-in test

The current ordering is asserted by a test at [`scripts/ci/generate-update-metadata.test.mjs`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/scripts/ci/generate-update-metadata.test.mjs#L59-L83) titled *"uses the universal macOS ZIP as the top-level updater path"*. It locks in the universal-first ordering as desired behaviour, so the test needs flipping in lockstep with the fix and an explanatory comment about why.

## Desired behaviour

`latest-mac.yml` should rank `-x64-mac.zip` first so Intel Macs select it. Apple Silicon's arm64 selection must keep working. With both natives shipping, the universal ZIP should never be picked by any client.

## Upstream context

This is the known electron-updater limitation tracked in:

- electron-userland/electron-builder#8554 ("Universal Auto-Update Upgrade?")
- electron-userland/electron-builder#6694 ("macOS Intel installs ARM on auto-update")
- The substring filter behaviour was introduced in electron-userland/electron-builder#5524

Upstream does not prioritise native over universal; the ordering is on us.

## Stack

`electron-updater@6.3.0`, `electron-builder@26.8.1`, `electron@41.1.0`.

gregpriday:
## Summary
- Intel Macs were downloading the larger `universal-mac.zip` on auto-update instead of the native `-x64-mac.zip`. The `macZipPriority` function ranked universal first, and `electron-updater` picks the first surviving entry after its arch filter.
- The fix reorders priorities so native arch ZIPs come first, then universal last. Apple Silicon is unaffected -- the arm64 substring filter picks `-arm64-mac.zip` regardless of ordering.
- Tests are updated to assert the new ordering with explicit x64-first and arm64-filtering coverage.

Resolves #9259

## Changes
- `scripts/ci/generate-update-metadata.mjs` -- reorder `macZipPriority` return values so `-x64-mac.zip` (1) and `-arm64-mac.zip` (2) rank ahead of `universal-mac.zip` (0 -> 3)
- `scripts/ci/generate-update-metadata.test.mjs` -- flip the locked-in test assertion from "universal first" to "native arch first", add explicit x64-first and arm64-only ordering tests

## Testing
- Updated unit tests run and pass locally
- Manual inspection confirms `-x64-mac.zip` surfaces at `path` and first in `files[]` in the generated YAML

### DIFF
--- a/scripts/ci/generate-update-metadata.mjs
+++ b/scripts/ci/generate-update-metadata.mjs
@@ -47,10 +47,18 @@ async function artifactEntry(releaseDir, fileName) {
   return entry;
 }
 
+// electron-updater's MacUpdater.ts filters by "arm64" substring: on Apple Silicon
+// only files containing "arm64" are kept; on Intel Macs "arm64" files are excluded.
+// It then picks the first surviving entry. Rank native builds before universal so
+// each architecture selects its own native ZIP and universal is never picked when
+// a native build exists.
 function macZipPriority(fileName) {
-  if (fileName.includes("universal-mac.zip")) return 0;
-  if (fileName.includes("-arm64-mac.zip")) return 2;
-  if (fileName.includes("-x64-mac.zip") || fileName.endsWith("-mac.zip")) return 1;
+  if (fileName.includes("-x64-mac.zip")) return 0;
+  if (fileName.includes("-arm64-mac.zip")) return 1;
+  if (fileName.includes("universal-mac.zip")) return 2;
+  // Bare -mac.zip (no arch suffix) is the x64 build from electron-builder.
+  // Check after arm64 so arm64-mac.zip doesn't match this branch.
+  if (fileName.endsWith("-mac.zip")) return 0;
   return 3;
 }
 
--- a/scripts/ci/generate-update-metadata.test.mjs
+++ b/scripts/ci/generate-update-metadata.test.mjs
@@ -56,7 +56,11 @@ describe("generate-update-metadata", () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


gregpriday:

gregpriday:

Issue #9239, opened 2026-05-27
Audit forge-provider impl method calls


gregpriday:
The MCP side has a rich audit pipeline: `McpAuditRecord` at [`shared/types/ipc/mcpServer.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/types/ipc/mcpServer.ts#L105-L145), `AuditService` at [`electron/services/mcp-server/auditLog.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/services/mcp-server/auditLog.ts#L62-L594), anomaly signals across latency-drift / first-seen-combination / failure-cluster / p95-z-score, and assistant-turn correlation via `turnId`. None of it covers forge-provider calls.

`getForgeProviderImpl` at [`electron/services/forgeProviderRegistry.ts:138-141`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/services/forgeProviderRegistry.ts#L138-L141) returns the registered impl raw. Every downstream method call (`provider.createPR(...)`, `provider.listIssues(...)`, anything on `ForgeProviderImpl`) runs untimed and unaudited. A slow provider, a credential failure, or a malformed response leaves no trail.

Forge calls match MCP's semantic weight: discrete user-visible operations at low frequency. That makes them the right surface to wire first, before the higher-frequency surfaces in the follow-up issues.

## Scope

In: a new audit-record shape for non-MCP host invocations under `shared/types/ipc/`, a parallel audit service mirroring `AuditService` with anomaly scoping that handles heterogeneous source latencies, and forge-provider impl method calls audited as the first source. Successful and failed outcomes both audit, per the `#8442` audit-even-when-silenced doctrine.

Out: changes to the existing `McpAuditRecord` on-disk shape (no migration), the higher-frequency surfaces (`PLUGIN_INVOKE`, file-decoration timeouts), and the settings-tab viewer. Each lives in a follow-up issue.

## References

- `#8442`: audit-even-when-silenced doctrine
- `#9151`: the parallel MCP-side observability gap, which uses `appendGrantRecord`'s discriminated-union extension as the precedent for sibling record shapes

[...]

PR #9383, opened 2026-05-27
feat(plugin): audit plugin invokes and file-decoration failures

### LINKED ISSUE #9240: Audit plugin invokes and file-decoration timeouts
Two gaps in the plugin IPC layer leave failed calls invisible to the audit pipeline.

The `PLUGIN_INVOKE` handler at [`electron/ipc/handlers/plugin.ts:244-260`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/ipc/handlers/plugin.ts#L244-L260) writes no audit record on any outcome. Success, throw, and the `isTrustedRendererUrl` rejection at line 248 all run silently.

The file-decoration provider path at [`electron/ipc/handlers/plugin.ts:140-218`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/ipc/handlers/plugin.ts#L140-L218) drops both per-impl timeouts (`DECORATION_PROVIDER_TIMEOUT_MS`, 3000ms budget) and rejected provider promises to `console.warn` at lines 170-181. Neither reaches an audit row.

Both violate the `#8442` audit-even-when-silenced doctrine. A misbehaving plugin or a stalled decoration provider should be diagnosable from the audit log, not lost to a `console.warn` that the user never sees.

File decorations fire at render-tree-change frequency, so successful settles should not audit, only failures. Same shape as `#8442` writing the silenced-denial row without recording every nominal call.

The existing scrub pipeline (`summarizeMcpArgs`, `scrubSecrets`, `sanitizePath` at [`shared/utils/mcpArgsSummary.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/utils/mcpArgsSummary.ts)) covers privacy at the call site and is reusable as-is.

Depends on #9239 for the audit-record shape and service.

gregpriday:
## Summary

- Adds audit records for two previously-silent plugin IPC failure paths, extending the audit-even-when-silenced doctrine to the plugin host
- `plugin:invoke` now audits all outcomes: success, `UNTRUSTED_SENDER` rejection, and `DISPATCH_THREW` — `plugin:file-decorations-get` audits only per-provider failures (`TIMEOUT`, `PROVIDER_REJECTED`), since successful settles run at render-tree-change frequency and would flood the log
- New `PluginAuditService` mirrors the MCP `AuditService` (config-backed ring buffer, lazy hydration, debounced flush) and reuses the existing scrub pipeline so raw args and paths are never persisted

Resolves #9240

## Changes

- `shared/types/ipc/plugin.ts` — new `PluginAuditRecord` type with a `source: 'plugin'` discriminator; forward-compatible with the host-audit union planned in #9239
- `electron/services/pluginAuditService.ts` — new service; backed by audit fields added to the existing `plugins` store key; flushed on app shutdown via `electron/lifecycle/shutdown.ts`
- `electron/ipc/handlers/plugin.ts` — `plugin:invoke` wired for all three outcomes; `plugin:file-decorations-get` wired for per-provider failures; `provideDecorations` wrapped so synchronous throws become auditable rejections instead of escaping `Promise.allSettled`

## Testing

- `electron/services/__tests__/pluginAuditService.test.ts` — 13 new tests covering ring-buffer eviction, scrubbing, flush debounce, and hydration
- `electron/ipc/handlers/__tests__/plugin.handlers.test.ts` — additions covering all three `plugin:invoke` outcomes and both `plugin:file-decorations-get` failure paths
- 59 tests pass

### DIFF
--- a/eager-import-baseline.json
+++ b/eager-import-baseline.json
@@ -1,6 +1,6 @@
 {
-  "count": 1258,
-  "moduleCount": 1258,
+  "count": 1260,
+  "moduleCount": 1260,
   "allowlist": [
     "electron/ipc/errorHandlers.ts",
     "electron/ipc/handlers/agentCapabilities.ts",
@@ -66,6 +66,7 @@
     "electron/services/mcp-server/httpLifecycle.ts",
     "electron/services/persistence/db.ts",
     "electron/services/persistence/readLastProjectId.ts",
+    "electron/services/pluginAuditService.ts",
     "electron/services/pty/TerminalRegistry.ts",
     "electron/services/pty/agentSessionHistory.ts",
     "electron/services/pty/terminalSessionPersistence.ts",
@@ -1045,6 +1046,16 @@
       "pattern": "sync-sqlite"
     },
     {
+      "file": "electron/services/pluginAuditService.ts",
+      "line": 169,
+      "pattern": "sync-store-get"
+    },
+    {
+      "file": "electron/services/pluginAuditService.ts",
+      "line": 176,
+      "pattern": "sync-store-get"
+    },
+    {
       "file": "electron/services/PluginService.ts",
       "line": 325,
       "pattern": "sync-store-get"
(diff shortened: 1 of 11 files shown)


gregpriday:

PR #9385, opened 2026-05-27
fix(notifications): tighten GridNotificationBar live-region a11y and motion parity

### LINKED ISSUE #9190: Tighten GridNotificationBar live-region a11y and motion parity
The `role="status"` wrapper in `GridNotificationBar.tsx` has four overlapping a11y problems. They all live in the same render path and want one PR.

### Action buttons inside the live region

The wrapper around both the text content and the action buttons is `role="status"`. That's a WAI-ARIA APG anti-pattern. Live-region content is announced as flat text, and screen readers don't reliably convey nested-button semantics during the announcement. Action buttons should render as siblings outside the live region.

See [`src/components/Terminal/GridNotificationBar.tsx#L140-L216`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Terminal/GridNotificationBar.tsx#L140-L216).

### Missing explicit `aria-atomic="true"`

The wrapper is `role="status" aria-live="polite"` but doesn't set `aria-atomic` explicitly. The role implies it, but the W3C spec recommends explicit override. Without it, some browsers announce only the diffed nodes instead of the full region. The sibling `InlineStatusBanner.tsx` already sets it explicitly.

### Synchronous content swap drops VoiceOver announcements

When a notification swaps mid-display, `setDisplayedNotification(notification)` runs synchronously. Replacing the DOM text in one tick can wipe the active speech buffer. VoiceOver on macOS Sequoia and iOS 18 has been observed to drop the announcement entirely in this case. The APG mitigation is clear-then-add-on-next-tick: clear the region's text, wait roughly 150ms, then render the new content. The delay sits well below the 400ms Doherty threshold so it's invisible to sighted users.

See [`src/components/Terminal/GridNotificationBar.tsx#L73-L109`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Terminal/GridNotificationBar.tsx#L73-L109).

### No reduced-motion / performance-mode short-circuit

The component applies `BANNER_ENTER_DURATION` and `BANNER_EXIT_DURATION` directly as inline style, without checking `prefers-reduced-motion`, `data-reduce-animations`, or `data-performance-mode`. The sibling `InlineStatusBanner.tsx` reads all three at [`src/components/Terminal/InlineStatusBanner.tsx#L115-L122`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Terminal/InlineStatusBanner.tsx#L115-L122) and short-circuits its animation. The grid bar should match.

Note: the clear-then-add delay should not be gated on reduced-motion. It's a screen-reader buffer, not motion.

The existing tests in `src/components/Terminal/__tests__/GridNotificationBar.test.tsx` will need timing assertions and a structural-split update.

gregpriday:
## Summary

- Fixes four `GridNotificationBar` accessibility issues: action buttons split out of `role="status"` so screen readers only announce notification text; `aria-atomic="true"` set explicitly; a 150ms buffer-flush delay on notification swap so VoiceOver doesn't drop the announcement; and `prefers-reduced-motion` / `data-reduce-animations` / `data-performance-mode` honored, matching the `InlineStatusBanner` pattern.
- Always-mount the live region — removed the early-return that dropped `role="status"` from the DOM when empty. The outer wrapper collapses to `h-0` when there's nothing to show, but AT registers the region on page load and won't miss the first announcement.
- A→B→C retarget logic collapses rapid notification replacements into a single announcement of the latest.

Resolves #9190

## Changes

- `src/components/Terminal/GridNotificationBar.tsx` — structural split of live region and action buttons, always-mount, swap delay, reduced-motion guard, retarget collapse
- `src/lib/animationUtils.ts` — `LIVE_REGION_SWAP_DELAY` constant (150ms; not gated on reduced-motion — it's a screen-reader timing concern, not a motion concern)
- `src/components/Terminal/__tests__/GridNotificationBar.test.tsx` — 19 vitest cases covering structural split, swap delay, retarget, reduced motion, and existing animation behavior

## Testing

`npm run check` and build both clean. Test suite updated with 19 cases; all pass.

### DIFF
--- a/src/components/Terminal/GridNotificationBar.tsx
+++ b/src/components/Terminal/GridNotificationBar.tsx
@@ -2,7 +2,11 @@ import React, { useEffect, useRef, useState } from "react";
 import { AlertTriangle, CheckCircle2, Info, XCircle } from "lucide-react";
 import { useShallow } from "zustand/react/shallow";
 import { cn } from "@/lib/utils";
-import { BANNER_ENTER_DURATION, BANNER_EXIT_DURATION } from "@/lib/animationUtils";
+import {
+  BANNER_ENTER_DURATION,
+  BANNER_EXIT_DURATION,
+  LIVE_REGION_SWAP_DELAY,
+} from "@/lib/animationUtils";
 import { useNotificationStore, type Notification } from "@/store/notificationStore";
 
 const STATUS_CONFIG = {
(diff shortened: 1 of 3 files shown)


gregpriday:

gregpriday:

PR #9389, opened 2026-05-27
feat(worktree): restore focus and add file-stepping in change viewers

### LINKED ISSUE #9217: Restore focus and add file-stepping in worktree change viewers
Two related navigation gaps in the worktree change viewers leave users stranded deep in a diff with no clear way back to the dashboard.

### Focus is not restored on close

When a user closes a diff, the file viewer, or the Review Hub opened from a worktree card, keyboard focus is not returned to the `FileChangeList` row (or card control) that opened it. Instead focus drops to the top of the document, so keyboard users lose their place in the change list.

`AppDialog` already solves this through its `restoreFocusTo` prop, and `ConfirmDialog` passes it. The change viewers don't:

- [`FileViewerModal`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/FileViewer/FileViewerModal.tsx) renders inside `AppDialog` but never passes `restoreFocusTo`.
- [`ReviewHub`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Worktree/ReviewHub/ReviewHub.tsx) is a hand-rolled `role="dialog"` that doesn't use `AppDialog` at all. It focuses itself on open but restores nothing on close.

The `focus()` calls already present in these surfaces are focus-into-on-open, never focus-return-on-close. Closing a viewer should land focus back on the row that opened it.

### No way to step between changed files

[`FileDiffModal`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Worktree/FileDiffModal.tsx) opens a single file with no next/previous affordance. Reviewing several files means closing the viewer and reopening the next file from the list every time.

[`FileChangeList`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Worktree/FileChangeList.tsx) already computes a deterministic sorted order of the changes, but it passes only the file path and status to the viewer. The viewer has no knowledge of its neighbours. A user reviewing a worktree's changes should be able to move to the next or previous changed file from inside the viewer, keyboard-driven, without returning to the dashboard between files.

gregpriday:
## Summary

- Closing a file diff returns keyboard focus to the row that opened it. `FileChangeList` rows are now real buttons (`role=button`, `tabIndex`, Enter/Space, `aria-label`, focus ring). `ReviewHub` (hand-rolled dialog) restores focus to its opener on close; diffs opened from inside `ReviewHub` return focus to the inspected row.
- `FileDiffModal` and `FileViewerModal` gain previous/next file controls -- ChevronLeft/ChevronRight header buttons, a `{n} of {total}` position indicator, and `[`/`]` keyboard shortcuts. Steps across the full sorted change set, bounds-clamped at both ends. `[` and `]` were chosen because `n`/`p` are taken by hunk navigation and arrow keys conflict with split-diff horizontal scroll.
- The open file's index is derived from its identity each render, so a background worktree refresh that reorders or shrinks the list keeps the indicator and stepping cursor correct without stale index state.

Resolves #9217

## Changes

- `FileChangeList.tsx` -- rows promoted to interactive buttons with full keyboard and ARIA support
- `FileDiffModal.tsx` -- previous/next controls wired to `onNavigate` prop; `[`/`]` shortcuts
- `FileViewerModal.tsx` -- same stepping controls + shortcuts; index derived from identity
- `ReviewHub.tsx` / `ReviewHubContent.tsx` -- focus-restore on close; diffs return focus to their source row
- Tests: 256 passing across `FileChangeList`, `FileDiffModal`, `FileViewerModal`, and `ReviewHub` suites

Out of scope (follow-up): file-stepping within ReviewHub's staged/unstaged sections -- ordering across multi-select/filtered sections is ambiguous and warrants a separate issue.

## Testing

`npm run check` clean (typecheck, lint ratchet, format, IPC checks). 256 unit tests passing across all four affected suites.

### DIFF
--- a/src/components/FileViewer/FileViewerModal.tsx
+++ b/src/components/FileViewer/FileViewerModal.tsx
@@ -1,11 +1,19 @@
 import { useEffect, useEffectEvent, useCallback, useState, useRef, useMemo } from "react";
 import { AppDialog } from "@/components/ui/AppDialog";
+import type { RestoreFocusTarget } from "@/components/ui/AppDialog";
 import { DiffViewer } from "@/components/Worktree/DiffViewer";
 import { CodeViewer } from "./CodeViewer";
 import type { CodeViewerHandle } from "./CodeViewer";
 import { filesClient } from "@/clients/filesClient";
 import { actionService } from "@/services/ActionService";
-import { ExternalLink, Copy, Check, Image as ImageIcon } from "lucide-react";
+import {
+  ExternalLink,
+  Copy,
+  Check,
+  Image as ImageIcon,
+  ChevronLeft,
+  ChevronRight,
+} from "lucide-react";
 import { Skeleton, SkeletonBone, SkeletonText } from "@/components/ui/Skeleton";
 import { cn } from "@/lib/utils";
 import { formatBytes } from "@/lib/formatBytes";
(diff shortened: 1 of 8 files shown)


gregpriday:

gregpriday:

PR #9392, opened 2026-05-27
fix(recipes): guard in-repo recipe writes against newer disk content

### LINKED ISSUE #9186: In-repo recipe edits silently overwrite newer disk content
In-repo recipes live in git-tracked `.daintree/recipes/*.json` files. The app reads them once via `loadRecipes` (project switch, new-worktree dialog, startup) and never re-reads them otherwise — there's no file watcher and no window-focus reload.

The write path persists with no precondition check. So if a `git pull`, branch switch, or `git stash pop` changes a recipe file underneath a running app, the next edit to that recipe writes the in-memory copy straight over the newer content on disk. The change that came in from git is gone, with no warning.

`writeInRepoRecipe` in [`electron/services/ProjectIdentityFiles.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/services/ProjectIdentityFiles.ts) writes unconditionally; the [`updateInRepoRecipe` handler](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/ipc/handlers/projectRecipes.ts#L225-L255) does no staleness check before calling it. The only reload is [`loadRecipes`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/store/recipeStore.ts#L193-L236).

This is a data-loss path for team-shared recipes, and it's exactly the "no silent fallback defaults" case `CLAUDE.md` calls out — newer disk content should never be clobbered silently.

Desired behavior:

- The in-repo write path rejects the write when the file changed on disk since it was loaded, rolls back the optimistic state, and surfaces a conflict prompt (reload / overwrite). A content hash is preferable to raw mtime if false-positive conflicts show up.
- A cheap reload on window focus, so most external edits reconcile without a persistent watcher.

A persistent file watcher on `.daintree/recipes/` is explicitly out of scope here — it carries an FD/lifecycle cost and should wait until there's evidence background-pull collisions are frequent enough to justify it.

gregpriday:
## Summary

- In-repo recipes (`.daintree/recipes/*.json`) were loaded once and never re-read, with no staleness check on the write path. A `git pull` or branch switch could change a file on disk, and the next edit would silently overwrite the newer content.
- Adds a content-hash guard to `writeInRepoRecipeChecked` in `ProjectStore.ts`. If the on-disk hash differs from the cached hash, the write throws `RECIPE_STALE_CONFLICT`. The renderer catches it in `useRecipeDialogState.ts` and routes to `RecipeConflictDialog.tsx` where the user can reload or overwrite.
- Adds a `useRecipeFocusReload` hook that re-reads in-repo recipes on window focus, so most git-pull collisions reconcile automatically before the user even opens a recipe.

Resolves #9186

## Changes

- `electron/services/ProjectStore.ts` — `writeInRepoRecipeChecked` with hash guard; `readInRepoRecipes` populates the hash cache; `deleteInRepoRecipe` clears stale entries
- `electron/services/ProjectIdentityFiles.ts` — `getInRepoRecipeFileHash` and `readInRepoRecipesWithHashes` to support cache population
- `electron/ipc/handlers/projectRecipes.ts` — passes `force` flag through to `writeInRepoRecipeChecked`
- `shared/types/appError.ts` / `shared/types/ipc/*` — `RECIPE_STALE_CONFLICT` error code and `updateInRepoRecipe` force flag
- `src/store/recipeConflictStore.ts` — Zustand store driving the conflict dialog
- `src/store/recipeEditorActivityStore.ts` — tracks open editor sessions so focus-reload skips recipes currently being edited
- `src/store/recipeStore.ts` — handles `RECIPE_STALE_CONFLICT`, rolls back optimistic state, opens conflict dialog
- `src/components/TerminalRecipe/RecipeConflictDialog.tsx` — reload / overwrite conflict resolution dialog
- `src/hooks/app/useRecipeFocusReload.ts` — focus-triggered reload hook, wired in `App.tsx`
- `src/components/Sidebar/useRecipeDialogState.ts` — catches the stale conflict and routes to the conflict store

## Testing

- Unit tests in `electron/services/__tests__/writeInRepoFiles.test.ts` and `electron/services/__tests__/ProjectStore.recipes.test.ts` cover the hash guard, force-write path, rename staleness check, and cache invalidation on delete
- `src/store/__tests__/recipeStore.test.ts` covers optimistic rollback and conflict store wiring on `RECIPE_STALE_CONFLICT`
- `src/store/__tests__/recipeEditorActivityStore.test.ts` covers editor-open tracking
- `src/hooks/app/__tests__/useRecipeFocusReload.test.ts` covers the focus-reload hook

### DIFF
--- a/electron/ipc/handlers/projectRecipes.ts
+++ b/electron/ipc/handlers/projectRecipes.ts
@@ -257,15 +257,19 @@ export function registerProjectRecipesHandlers(_deps: HandlerDependencies): () =
       throw new Error(`Project not found: ${projectId}`);
     }
     assertNoSecretEnvValues(recipe.terminals);
-    await projectStore.writeInRepoRecipeChecked(project.path, recipe, { force: force === true });
+    await projectStore.writeInRepoRecipeChecked(project.path, recipe, {
+      force: force === true,
+      previousName: typeof previousName === "string" ? previousName : undefined,
+    });
     if (
       previousName &&
       typeof previousName === "string" &&
       safeRecipeFilename(previousName) !== safeRecipeFilename(recipe.name)
     ) {
-      // Delete the old-name file; the stale cache entry for the old recipe id
-      // self-heals on the next loadRecipes (which clears the per-project hash
-      // map and repopulates from disk).
+      // Delete the old-name file. The staleness of the old file is checked
+      // inside writeInRepoRecipeChecked before the new write runs, so by the
+      // time we reach here the rename has been authorized. The cache entry
+      // for the old recipe id self-heals on the next loadRecipes.
       await projectStore.deleteInRepoRecipe(project.path, previousName);
     }
   };
--- a/electron/ipc/handlers/projectRecipes.ts
+++ b/electron/ipc/handlers/projectRecipes.ts
@@ -240,11 +240,12 @@ export function registerProjectRecipesHandlers(_deps: HandlerDependencies): () =
     projectId: string;
     recipe: TerminalRecipe;
     previousName?: string;
+    force?: boolean;
   }): Promise<void> => {
     if (!payload || typeof payload !== "object") {
       throw new Error("Invalid payload");
     }
-    const { projectId, recipe, previousName } = payload;
+    const { projectId, recipe, previousName, force } = payload;
     if (typeof projectId !== "string" || !projectId) {
       throw new Error("Invalid project ID");
     }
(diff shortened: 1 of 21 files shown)


gregpriday:

gregpriday:

Issue #8442, opened 2026-05-19
Replace sticky tier elevation with time-bounded per-tool grants


gregpriday:
`setSessionTier` in `McpServerService` writes a single in-memory tier per help session, sticky until the `MCP_SSE_IDLE_TIMEOUT_MS` idle reaper (30 minutes) clears the session. Once a help session elevates to `action` or `system`, every tool at that tier remains callable for the rest of the session: no per-tool scope, no TTL, no automatic decay, no explicit revoke path.

That's the Azure PIM request-fatigue anti-pattern. Users elevate for the maximum duration to dodge re-prompting and JIT collapses back to standing privilege. A confused-deputy LLM can ride the same window.

Evidence: [`setSessionTier`](https://github.com/daintreehq/daintree/blob/6a8f2e99a97f019c99d2186db7b089098237d324/electron/services/McpServerService.ts#L377-L383), [`MCP_SSE_IDLE_TIMEOUT_MS`](https://github.com/daintreehq/daintree/blob/6a8f2e99a97f019c99d2186db7b089098237d324/electron/services/mcp-server/shared.ts#L68), [`isTierPermitted`](https://github.com/daintreehq/daintree/blob/6a8f2e99a97f019c99d2186db7b089098237d324/electron/services/mcp-server/tierAuth.ts#L97-L106).

## Desired behaviour

Replace the sticky session tier with a per-`(sessionId, toolId)` grant cache, modelled on `sudo`'s `timestamp_timeout`: a 15-minute sliding window that refreshes on each successful dispatch, capped well below the 30-minute session idle to avoid races against the reaper.

The static `TIER_ALLOWLISTS` floor stays. The grant cache layers on top, consulted at `isTierPermitted` before falling back to set membership.

## Auditing

The grant lifecycle becomes its own auditable surface. Add an `McpGrantRecord` variant alongside the existing `McpAuditRecord` in the ring buffer, carrying:

- `type`: `grant.issued`, `grant.expired`, or `grant.revoked`
- `sessionId`
- `toolId`
- `decision`
- `ttlMs`
- `webContentsId`
- `projectId`

`McpServerService` emits renderer-facing grant lifecycle events through the same broadcast pattern already used for `onRuntimeStateChange`.

## Revoke path

A session-level revoke drops every grant for that session from the cache and writes matching `grant.revoked` records. The session returns to its underlying tier floor.

## Denial tracking

[...]

PR #9395, opened 2026-05-27
feat(forge): audit forge-provider impl method calls

### LINKED ISSUE #9239: Audit forge-provider impl method calls

gregpriday:
## Summary

- Adds audit coverage for forge-provider impl method calls, mirroring the existing MCP audit pipeline. Forge calls match MCP's semantic weight — discrete user-visible operations at low frequency — and were the right surface to wire first.
- Per-call-site instrumentation via `auditForgeCall` covers eight methods across both success and failure paths, following the #8442 doctrine (audit even when silenced). Redaction strips tokens, usernames, and raw URLs before anything touches disk. `null` returns map to not-found; rejected credentials classify as error.
- Shutdown flushes the audit log so in-flight records aren't lost on quit.

Resolves #9239

## Changes

- `shared/types/ipc/forge.ts` — new `ForgeAuditRecord` discriminated-union type
- `electron/store.ts` — `forgeAudit` store key
- `electron/services/forge/auditLog.ts` — pure `ForgeAuditService` with ring buffer, anomaly signals (low-frequency thresholds: floor 10, failure-cluster 2/5, p95 ≥ 3 methods), and redaction
- `electron/services/forge/forgeAuditService.ts` — singleton wrapper + `auditForgeCall` helper
- `electron/ipc/handlers/forgeData.ts`, `forge.ts`, `forgeSettings.ts` — per-call-site instrumentation of `listIssues`, `listPRs`, `getIssue`, `getPR`, `getRepoMetadata`, `assignIssue`, `unassignIssue`, `validateToken`
- `electron/lifecycle/shutdown.ts` — flush on `before-quit`

## Testing

81 tests pass across `auditLog.test.ts`, `forgeAuditService.test.ts`, `forgeData.handlers.test.ts`, and `forgeSettings.handlers.test.ts`. Typecheck, lint, and format clean.


gregpriday:
**CI note — the one red check is a pre-existing, unrelated flake, not the forge work.**

All forge-specific tests pass and the build/typecheck/lint/format/import-budget checks are green. The single failing unit test is `src/store/__tests__/persistenceBoundaryHardening.test.ts > "notifies exactly once when a permanent structural fallback occurs"`, which is unrelated to this PR's scope (renderer `safeStorage`, not the `electron/` forge audit code).

Root cause: that suite's sibling fallback tests each fire a notification via an **unawaited** dynamic `import("@/lib/notify")`. Those in-flight dispatches can resolve while the "notifies exactly once" test has its spy installed, inflating the count (seen as 2–6 depending on CI file-ordering). The product guard (`safeStorage`'s synchronous `hasNotifiedPermanentFallback`) is correct — this is purely a test-isolation race. It passes reliably in isolation locally; adding this PR's new forge test files shifted vitest's file-execution order and surfaced the latent flake.

I attempted an isolation fix here (drain pending dispatches + `mockClear` the spy right before the test triggers its own fallback), which reduced the over-count but couldn't fully eliminate the timing-dependent straggler. A complete fix likely belongs in a focused follow-up (e.g. awaiting the dispatch via a test seam in `safeStorage`, or isolating that test into its own file). Flagging rather than expanding scope further.

PR #9397, opened 2026-05-27
fix(plugin): sync plugin lifecycle to LRU-evicted views on revival

### LINKED ISSUE #9285: Sync plugin lifecycle across project views and survive LRU eviction
Each project gets its own `WebContentsView` with an independent V8 context (`electron/window/ProjectViewManager.ts:8-12, 282-349`). `ProjectViewManager` caps the live view count and evicts the LRU view when over the limit (`evictStaleViews("lru")` at `:533-534`). Plugin lifecycle today is push-on-change from `PluginService` — `usePluginActions`, `usePluginToolbarButtons`, and `usePluginPanelKinds` each pull on mount and listen for `plugin:*-changed` broadcasts. That works for the active view. It is not tested against the three-view matrix, and the eviction-and-restore path is where it most plausibly breaks.

Failure modes I expect tests to find:

- **Unload doesn't propagate to inactive views.** Plugin uninstalled in view A. View B is active and updates. View C is idle, hasn't pumped the IPC event yet. C's `ActionService` keeps a stale registration; a later dispatch hits a synthetic definition whose `run()` calls into main for a plugin that no longer exists.
- **Eviction discards state without a replay path.** View C is evicted, later restored cold (`acquireView` at `:282-349`). The renderer remounts and pull-on-mount fires. The `pushReceived`-vs-pull race is only audited for first-mount (`usePluginActions.ts:65-82`), never for an LRU restore. The pull may also be missing a "ready" gate — calling `getActions()` before `PluginService.init` finishes returns an empty array and the hook quietly registers nothing.
- **Toolbar stale-prune race.** `usePluginToolbarButtons` only sweeps on a `complete: true` push (`:42-44`). A restored view receives the next partial push first, sweeps nothing, persists a stale `plugin.` entry until uninstall finally fires.

Two-step issue. Tests first against the three-view setup (a Playwright `full-resilience` spec covering install/uninstall while two views are idle, then again after evicting one). Then patch: gate pulls on a "PluginService ready" promise; emit a synthetic full push on view-revival via `plugin:hydrate(pluginId)` IPC; verify every hook's teardown drops `pluginConfirmStore` resolvers (done in `usePluginActions.ts:88-93`, confirm parity).

## Evidence

- [`electron/window/ProjectViewManager.ts:282-349`](https://github.com/daintreehq/daintree/blob/develop/electron/window/ProjectViewManager.ts#L282-L349) — `acquireView` and the eviction-timestamp telemetry around revival
- [`electron/window/ProjectViewManager.ts:533-534`](https://github.com/daintreehq/daintree/blob/develop/electron/window/ProjectViewManager.ts#L533-L534) — `evictStaleViews("lru")`
- [`src/hooks/usePluginActions.ts:65-82`](https://github.com/daintreehq/daintree/blob/develop/src/hooks/usePluginActions.ts#L65-L82) — pull-vs-push race; audited only for first-mount
- [`src/hooks/usePluginToolbarButtons.ts:42-44`](https://github.com/daintreehq/daintree/blob/develop/src/hooks/usePluginToolbarButtons.ts#L42-L44) — `complete: true` gating on the stale-prune sweep

## Constraints

- Tests come first. The patch shape depends on what actually breaks, and the patch can't ship without a regression net. `full-resilience` is the right bucket.
- The hydrate IPC has to be cheap — fires on every cold view mount. Reuse the same snapshot the pull returns.
- Audit `usePluginPanelKinds` and `usePluginToolbarButtons` teardowns for the `pluginConfirmStore`-drop pattern parity.

## Depends on

- #9271 — provenance, so a restored view renders the right "installed/disabled/load-error" state
- #9277 — `host.registerAction` is the broadcast source whose sync this tests

Under epic #5651.

gregpriday:
## Summary

- Plugin lifecycle events (installs, uninstalls, toolbar changes, panel kinds) broadcast to live/cached `WebContentsView`s but not to evicted ones. A cold-revived view's mount-time pull ran before `PluginService` finished scanning, seeding an empty registry; uninstalls that happened while the view was evicted were never swept.
- Gates the three pull handlers (`plugin:actions-get`, `plugin:toolbar-buttons`, `plugin:panel-kinds-get`) on a new `waitForInitialized()` so a revived view's pull returns the full registry rather than whatever partial state happens to be loaded at mount time.
- Fixes a secondary race: `handleToolbarButtons` now returns `{ buttons, complete: isInitialized }`. The renderer sweeps stale `plugin.*` pinned-button prefs only on a complete snapshot, preventing a partial load-time push from suppressing the later authoritative pull.

Resolves #9285

## Changes

- `PluginService`: `initializePromise` coalescing (concurrent callers share one scan), `isInitialized` getter, `waitForInitialized()` (swallows scan rejection so callers always unblock).
- `electron/ipc/handlers/plugin.ts`: all three pull handlers await `waitForInitialized()` before responding.
- `src/hooks/usePluginToolbarButtons.ts`: reads the new `complete` flag; skips stale-pin sweep when `complete` is false.
- `shared/types/ipc/generated.ts`: regenerated for the new toolbar-buttons result shape.
- `electron/main.ts` + `electron/window/globalServicesInit.ts`: E2E escape hatches (`__daintreeE2ELoadPlugin`, `__daintreeE2EUnloadPlugin`, `__daintreeHasAction`) — all gated behind `DAINTREE_E2E_MODE`.

## Testing

- 257 unit tests pass (`plugin.handlers`, `usePluginToolbarButtons`, `usePluginActions`, `PluginService`).
- New `full-resilience` E2E spec (`plugin-lifecycle-lru-eviction.spec.ts`) covers a fixture plugin (action + toolbar button contributions) across a three-view LRU eviction matrix: install before eviction, uninstall while evicted, revive and verify clean state.
- Typecheck, lint, format, and build all pass.

### DIFF
--- a/e2e/full/resilience/plugin-lifecycle-lru-eviction.spec.ts
+++ b/e2e/full/resilience/plugin-lifecycle-lru-eviction.spec.ts
@@ -1,0 +1,256 @@
+import { test, expect } from "@playwright/test";
+import type { Page } from "@playwright/test";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { launchApp, closeApp, type AppContext } from "../../helpers/launch";
+import { createFixtureRepos } from "../../helpers/fixtures";
+import { openAndOnboardProject } from "../../helpers/project";
+import { addAndSwitchToProject, selectExistingProjectAndRefresh } from "../../helpers/workflows";
+
+// Plugin lifecycle sync must survive LRU project-view eviction (#9285): a view
+// evicted (destroyed) while a plugin is installed/uninstalled must reflect the
+// current plugin contribution set once it cold-starts — its mount-time pull is
+// the only reconciliation path, since the broadcast fired into a dead view.
+
+const PROJECT_A = "project-A";
+const PROJECT_B = "project-B";
+const PROJECT_C = "project-C";
+
+// cache=2 over three projects evicts exactly one LRU view per switch — the same
+// determinism the core eviction spec relies on.
+const CACHE_LIMIT = 2;
+
+const PLUGIN_ID = "e2e.lifecycle";
+const ACTION_ID = "e2e.lifecycle.ping";
+const ACTION_CONTRIBUTION = {
+  id: ACTION_ID,
+  title: "Lifecycle ping",
+  description: "E2E lifecycle-sync probe action",
+  category: "plugin",
+  kind: "command",
+  danger: "safe",
+} as const;
+
+let ctx: AppContext;
+let fixtureCleanups: Array<() => void> = [];
+let pluginRoot = "";
+let pluginDir = "";
+let projectIdA = "";
+let projectIdC = "";
+
+async function configurePvm(app: AppContext["app"], limit: number): Promise<void> {
+  await app.evaluate((_electron, n) => {
+    const g = globalThis as Record<string, unknown>;
+    const getPvm = g.__daintreeGetPvm as (() => unknown) | undefined;
+    const pvm = getPvm?.() as
+      | {
+          setCachedViewLimit: (n: number) => void;
+          setLowMemoryFreeThresholdMb?: (mb: number | null) => void;
+        }
+      | null
+      | undefined;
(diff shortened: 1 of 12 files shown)


gregpriday:

Issue #9230, opened 2026-05-27
Schema-driven IPC types at the plugin host API boundary


gregpriday:
The widest unclosed type-safety gap in the plugin host API is the IPC handler shape. Plugin authors writing against the current surface end up with `as` casts at every callsite, with no runtime validation at the boundary either.

The handler type lives in [`shared/types/plugin.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/types/plugin.ts#L130-L133):

```ts
export type PluginIpcHandler = (
  ctx: PluginIpcContext,
  ...args: unknown[]
) => unknown | Promise<unknown>;
```

`registerHandler` and `dispatchHandler` in [`electron/services/PluginService.ts`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/services/PluginService.ts#L851-L877) take that raw `unknown[]` straight off the wire and spread it into the handler with no validation. The symmetric surface for host-to-renderer pushes is `broadcastToRenderer(channel: string, payload: unknown)` — same loose shape.

## The pattern we already have nearby

The Anthropic MCP SDK (`@modelcontextprotocol/sdk`, already a dependency for Daintree's host MCP server) pairs an `inputSchema: z.object({...})` with `z.infer<typeof schema>` so the SDK consumer gets a fully-typed, schema-validated payload in their handler. Zod 4 is already heavy in `electron/schemas/`.

The same shape works here:

```ts
registerHandler<T extends z.ZodType>(
  channel: string,
  schema: T,
  handler: (ctx: PluginIpcContext, payload: z.infer<T>) => unknown | Promise<unknown>
);
```

`schema.parse()` at the boundary inside `dispatchHandler`; parse errors surface back through the dispatch reply. Mirror the same signature on `broadcastToRenderer`. Once `@daintreehq/plugin-sdk` exists, re-export `z` from there so plugin authors don't need a direct Zod install.

## Why now

This is the single largest type-safety improvement available without introducing codegen (vs Zed's WIT, Tauri's `tauri-specta`) and Zod is already the runtime-validation tool of choice in the codebase. Pre-1.0 with the github builtin not using `registerHandler` means the variadic-vs-single-payload signature can be settled cleanly without compat shims. The longer this stays loose, the more `as` casts accumulate in the first wave of external plugins.

[...]

PR #9399, opened 2026-05-27
feat(plugins): typed plugin.invoke channel registry with capability gate

### LINKED ISSUE #9231: Typed plugin.invoke channel registry with capability gate
The plugin IPC perimeter has no payload validation. [`PluginIpcHandler` at `shared/types/plugin.ts:130-133`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/types/plugin.ts#L130-L133) is literally:

```ts
export type PluginIpcHandler = (
  ctx: PluginIpcContext,
  ...args: unknown[]
) => unknown | Promise<unknown>;
```

PR #5277 already authenticates the sender. Payload validation is the missing layer, and the same dispatch site is the natural attachment point for a host-classified capability gate.

### Typed channel registration

Extend `registerHandler` on `PluginHostApi` to take a per-channel Zod schema for args and result, plus an optional `requires` array of declared permissions:

```ts
registerHandler<TArgs, TResult>(
  channel: string,
  schema: {
    args: ZodSchema<TArgs>;
    result: ZodSchema<TResult>;
    requires?: BuiltInPluginPermission[];
  },
  handler: (ctx: PluginIpcContext, args: TArgs) => Promise<TResult>,
): void;
```

`PluginService` dispatch at [`electron/services/PluginService.ts:880-886`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/electron/services/PluginService.ts#L880-L886) validates args via the registered Zod schema before invoking the handler and rejects on schema fail with a structured error returned to the renderer.

### Host-classified capability gate

Before dispatch, the host checks the channel's `requires[]` against the plugin's declared permissions ([`BUILT_IN_PLUGIN_PERMISSIONS` at `shared/types/plugin.ts:33-46`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/types/plugin.ts#L33-L46)) and fails closed with `PERMISSION_REQUIRED` if the manifest didn't declare what the channel needs. Plugin views today can `window.electron.plugin.invoke()` any channel a handler registered without the host re-checking declared permissions, which violates the rubric.

This is the direct extension of PR #8331's host-derived `danger` pattern to the IPC channel surface. CLAUDE.md commits to "classification not just wiring."

### Renderer-side typed hook

Mirror the typed contract on the renderer side so plugin authors get end-to-end types:

```ts
const { invoke, loading, error } = useHostChannel<TArgs, TResult>(channel);
```

Lands in `@daintreehq/plugin-sdk/react` (the package itself is a separate follow-up; this issue ships the host-side registration API and the renderer-side hook signature).

### Out of scope

Runtime permission prompt UI or sandboxing. Disclosure remains the model per [`docs/plugins/architecture.md:180`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/docs/plugins/architecture.md#L180).

gregpriday:
Closes #9231.

## What

Adds a typed contract and a host-side capability gate to the plugin `plugin:invoke` perimeter, plus a renderer hook for plugin view authors. Disclosure model only — runtime permission-prompt UI and sandboxing are explicitly out of scope.

### 1. Typed `registerHandler` overload (`shared/types/plugin.ts`)
Plugins can now register a channel with a Zod schema instead of a bare function:

```ts
host.registerHandler("greet", { args: z.object({ name: z.string() }), requires: ["network:fetch"] }, (ctx, payload) => `hi ${payload.name}`);
```

`schema` carries `args` (single payload — the variadic-vs-single-payload question from #9230 is settled in favor of single payload), an optional `result` schema, and an optional `requires` permission array. The untyped `(channel, handler)` overload is unchanged.

### 2. Host capability gate (`PluginService.dispatchHandler`)
For typed channels, before dispatch the host:
- **fails closed** with `PERMISSION_REQUIRED` (listing the missing permissions) when the plugin manifest doesn't grant a declared `requires` permission;
- validates the payload against `args`, returning `VALIDATION_FAILED` with a serializable issue projection;
- validates the handler's return against `result` (if given);
- wraps success in `{ ok: true, data }`.

A handler that throws *after* passing the gate returns a structured `HANDLER_ERROR` rather than rejecting. Untyped registrations keep the legacy raw passthrough and the throw-on-missing-handler contract, so existing plugin **action dispatch** (`usePluginActions`) is unchanged — minimal blast radius.

### 3. `useHostChannel` renderer hook (`src/hooks/useHostChannel.ts`)
`useHostChannel<TArgs, TResult>(pluginId, channel)` returns `{ invoke, isPending, data, error }`. Uses `useTransition` + a mounted-ref guard + a last-write-wins call-id guard (so a slow earlier invoke can't clobber a newer result), routes the envelope to `data`/`error`, clears stale data on failure, and synthesizes `HANDLER_NOT_FOUND` on a rejected invoke. Foundation for the future `@daintreehq/plugin-sdk/react` package; ships in-tree for now.

### Envelope
Structured `PluginInvokeResult<T>` / `PluginInvokeError` + serializable `PluginInvokeIssue` projection so validation/permission failures survive the contextBridge (raw `ZodError` instances lose all but `.message` across Structured Clone).

## Tests
- `PluginService`: capability gate (full grant / partial grant / no-permissions-field fail-closed / gate-before-validation), arg + result validation, `HANDLER_ERROR`, missing-handler still throws, bad result-schema rejected at registration.
- `useHostChannel`: success / validation / permission / `HANDLER_NOT_FOUND` / stale-data-cleared / concurrent out-of-order / legacy-raw / unmount guard.
- Typecheck, lint-ratchet (no new warnings), format, and IPC/channel gates all pass.

### DIFF
--- a/electron/services/PluginService.ts
+++ b/electron/services/PluginService.ts
@@ -5,6 +5,7 @@ import os from "os";
 import { pathToFileURL } from "url";
 import { app } from "electron";
 import * as semver from "semver";
+import type { z } from "zod";
 import { PluginManifestSchema } from "../schemas/plugin.js";
 import type {
   PluginManifest,
@@ -14,6 +15,8 @@ import type {
   PluginActivate,
   PluginActionContribution,
   PluginActionDescriptor,
+  PluginChannelSchema,
+  PluginInvokeResult,
   BuiltInPluginPermission,
 } from "../../shared/types/plugin.js";
 import type { WorktreeSnapshot } from "../../shared/types/workspace-host.js";
@@ -100,6 +103,27 @@ interface PluginLoadErrorRecord {
   at: number;
 }
 
+/**
+ * A registered channel handler plus its optional typed contract. Untyped
+ * registrations leave `argsSchema` undefined and dispatch raw (the legacy
+ * action-dispatch path); typed registrations carry the Zod schemas and the
+ * declared `requires` permissions enforced by the capability gate.
+ */
+interface HandlerEntry {
+  handler: PluginIpcHandler;
+  argsSchema?: z.ZodType;
+  resultSchema?: z.ZodType;
+  requires: BuiltInPluginPermission[];
+}
+
+function isChannelSchema(value: unknown): value is PluginChannelSchema {
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    typeof (value as { args?: { safeParse?: unknown } }).args?.safeParse === "function"
+  );
+}
+
 const ACTIVATE_TIMEOUT_MS = 5000;
 
 /**
@@ -147,7 +171,7 @@ type WorkspaceWorktreeEvent = "worktree-update" | "worktree-activated" | "worktr
 
 export class PluginService {
   private plugins = new Map<string, LoadedPlugin>();
-  private handlerMap = new Map<string, PluginIpcHandler>();
+  private handlerMap = new Map<string, HandlerEntry>();
   private cleanupMap = new Map<string, () => void>();
   private pluginActions = new Map<string, PluginActionDescriptor>();
   private pluginActionOwners = new Map<string, Set<string>>();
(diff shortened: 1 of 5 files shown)


gregpriday:

Issue #9277, opened 2026-05-27
Implement main-side `host.registerAction(descriptor, handler)`


gregpriday:
The docs ship `host.registerAction(descriptor, handler)` as the flagship runtime API a plugin's `activate(host)` calls to bind an action handler that runs in main. The host object exposed to plugins has no `registerAction` member. The only path to a runtime-contributed action is the renderer-side IPC flow on `plugin:actions-register`, which carries metadata but cannot carry a handler closure across the bridge — main has no callable when the action dispatches.

Every example in `docs/plugins/host-api.md:59-97` is non-executable, and downstream contribution surfaces (#9231, #9232, #9249, F12, F27, F28, F30) all assume a working main-side handler registry.

## Evidence

- `docs/plugins/host-api.md:35` declares `registerAction(descriptor, handler)` on `PluginHostApi`.
- `docs/plugins/host-api.md:59-97` walks through the call shape, the `{pluginId}.{descriptor.id}` namespacing rule, and the `"restricted"`-danger rejection.
- `shared/types/plugin.ts:191-262` defines `PluginHostApi`. No `registerAction` member.
- `electron/services/PluginService.ts:486-738` builds the host object via `createHost`. No `registerAction` closure is added.
- `electron/services/PluginService.ts:967-1042` is the only existing registration path — `registerPluginAction` — invoked over IPC from the renderer at `electron/ipc/handlers/plugin.ts:94`. It carries the descriptor but no handler reference.
- `shared/types/plugin.ts:276-301` defines `PluginActionContribution` and `PluginActionDescriptor` including `effectiveDanger`.

## Constraints

- Reuse the renderer-IPC path's validation: id regex, namespace-ownership check, `"restricted"` rejection, `effectiveDanger` promotion via `CONFIRM_TRIGGERING_PERMISSIONS`. Do not fork the validator. Extract a shared helper from `registerPluginAction` if needed.
- Wire into the existing `revoke` guard pattern used by the other host methods (`PluginService.ts:493-499`). Calling `registerAction` after `activate` resolves throws — handlers are imperative-only during activation.
- Dispatch must route through `ActionService.dispatch` so observability, MRU exclusion, and `effectiveDanger`-driven confirm gating apply. `ActionSource` for these invocations is the same as #9231's `"plugin"` extension (Wave 3 dependency).
- Dedup against the renderer IPC path — re-registering the same id from main after a renderer-side register replaces the prior entry, matching `host-api.md:95`.
[...]

Issue #9272, opened 2026-05-27
Lock the `daintree.*` plugin namespace to first-party plugins


gregpriday:
Nothing today prevents a sideloaded plugin from claiming `daintree.foo`. The duplicate-name check at [`PluginService.ts:343-354`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L343-L354) reserves disabled-built-in names and rejects exact-name collisions, but the regex at [`schemas/plugin.ts:20`](https://github.com/daintreehq/daintree/blob/develop/electron/schemas/plugin.ts#L20) accepts `daintree.anything` as a valid scoped name. A user who drops a plugin directory `~/.daintree/plugins/daintree.github-evil/` will load cleanly if the real `daintree.github` builtin isn't present — and once it is loaded, every renderer code path that checks `pluginId.startsWith("daintree.")` to gate first-party trust is wrong.

The first-party namespace is a trust signal. It needs to be locked at the contract layer, not by hoping the duplicate-name check covers it.

## Deliverables

- Manifest validation rejects any plugin whose `name` matches `/^daintree\./` unless the load originates from the bundled-builtin root (the path passed to `PluginService` as the builtin scan root). Non-builtin loads that try to use the prefix get a Zod refinement failure that surfaces in `console.error` and the toast emitted at [`PluginService.ts:362-367`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L362-L367).
- The install dialog from F21 performs the same check before the extract step so users see "this plugin claims a reserved namespace" before any filesystem mutation.
- Error copy follows the Title-Message-Action rule. Title `Plugin uses a reserved namespace`. Message names the offending `name` and explains that `daintree.*` is first-party only. No action button — the user has nothing to retry.

## Evidence

- [`electron/schemas/plugin.ts:20`](https://github.com/daintreehq/daintree/blob/develop/electron/schemas/plugin.ts#L20) — `SCOPED_PLUGIN_NAME_PATTERN` is publisher-agnostic.
- [`electron/services/PluginService.ts:343-354`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L343-L354) — reserved-name set only covers disabled builtins, not the prefix.
- [`plugins/builtin/github/plugin.json:2`](https://github.com/daintreehq/daintree/blob/develop/plugins/builtin/github/plugin.json#L2) — the canonical first-party plugin uses `daintree.github`.

[...]

Issue #9271, opened 2026-05-27
Persist plugin install provenance


gregpriday:
`LoadedPluginInfo` (`shared/types/plugin.ts:110-121`) carries `manifest`, `dir`, `loadedAt`, and `isBuiltin`. That is enough to load and disable a plugin, but not enough to answer any of the questions the Settings tab, diagnostics export, install rollback, or update flow will need to ask. `isBuiltin` is a single boolean — it can't distinguish between a sideloaded `.dntr`, a URL install, and a future catalog install. There is no record of when a plugin was installed (only when it was loaded), no archive hash to verify integrity, no original install URL, no slot for activation errors, no per-plugin disable state for non-builtins, no flag for dev-mode loads, and no slot for "update available".

The persistence layer is also missing. `electron/store.ts:285-300` only tracks `disabledBuiltins`. Non-builtins have nowhere to record their provenance, so even if `LoadedPluginInfo` grew the fields, they would be lost on next launch.

Concrete deliverables:

- Extend `LoadedPluginInfo` (`shared/types/plugin.ts:110-121`) with: `source: "builtin" | "sideload" | "url" | "catalog"`, `installedAt: number`, `archiveHash: string | null` (sha256 of the `.dntr` archive at install; `null` for builtins and dev-mode dir loads), `originalUrl: string | null`, `loadError: { message: string; stack?: string; at: number } | null`, `disabled: boolean` (distinct from the existing `disabledBuiltins` array — non-builtins need their own enable/disable state), `updateAvailable: { version: string; channel: "manual" } | null`, and `devMode: boolean` (loaded from a directory outside the managed plugins dir, e.g. via `daintree-plugin` CLI dev command).
- Add a `plugins.installed: Record<string, InstalledPluginRecord>` slot to the electron-store schema in `electron/store.ts`. Keyed by `manifest.name`. The record persists `source`, `installedAt`, `archiveHash`, `originalUrl`, `disabled`, `updateAvailable`, `devMode`, and `loadError`. `manifest`, `dir`, `loadedAt`, and `isBuiltin` are reconstructed at load time and not persisted.
- Migrate existing installs: anything currently in the user plugins dir gets a one-time record written on first launch under the new schema with `source: "sideload"`, `installedAt: Date.now()`, `archiveHash: null`, `originalUrl: null`, `disabled: false`. Built-ins do not get records — they're identified by load path.
[...]

PR #9401, opened 2026-05-27
feat(plugins): wire manifest commands to lazy handlers and standardize {pluginId}.{actionId} namespace

### LINKED ISSUE #9281: Wire manifest commands to lazy `src/<name>.ts` handlers and standardize the `{pluginId}.{actionId}` namespace
Today a plugin can declare `contributes.commands` in its manifest and `PluginService` will register the panel kinds, toolbar buttons, menu items, and forge providers — but the command entries themselves are never bound to a handler. The only way to get a runnable command is the imperative `host.registerAction` path inside `activate`. The Raycast-style filesystem convention documented at `docs/plugins/contribution-points.md:46-60` — "a command named `plan-from-issue` looks for `src/plan-from-issue.ts`, its default export is the handler" — has no implementation. Every plugin author who reads the docs and writes the file gets silent nothing.

Two things land here. First, the filesystem-convention binder: at load time, walk `manifest.contributes.commands`, resolve `src/{name}.{ts,tsx,js,mjs}` relative to the plugin root, register a synthetic descriptor whose handler `import()`s the module lazily on first dispatch (defers heavy module evaluation past the 5s activation budget). A command declared in the manifest with no handler file and no matching `registerAction` call shows the documented `Command "{pluginId}.{name}" has no handler` toast on dispatch — not at load time.

Second, the namespace reconciliation. `PluginService.ts:415` registers toolbar buttons as `plugin.{pluginId}.{id}` — divergent from the `{pluginId}.{actionId}` convention used everywhere else (see `registerPluginAction` validation at `PluginService.ts:975-984`). Pick one. The docs and `registerPluginAction` win; the toolbar registration is the outlier. Reconcile to the doc form, migrate the renderer-side `id.startsWith("plugin.")` check in `usePluginToolbarButtons.ts:77`, and add a single namespacing helper in `shared/types/plugin.ts` so future contribution surfaces can't drift again.

Collisions: reject any command whose resolved id matches a built-in action id at load time, surfaced via the #9271 provenance record's load-error field.

## Evidence

- [`docs/plugins/contribution-points.md:46-60`](https://github.com/daintreehq/daintree/blob/develop/docs/plugins/contribution-points.md#L46-L60) — documented filesystem convention with no implementation
- [`docs/plugins/dev-loop.md:144-150`](https://github.com/daintreehq/daintree/blob/develop/docs/plugins/dev-loop.md#L144-L150) — "Common issues → Command runs but handler doesn't execute" implies the binder exists
- [`electron/services/PluginService.ts:414-424`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L414-L424) — toolbar buttons use the divergent `plugin.{pluginId}.{id}` form
- [`electron/services/PluginService.ts:975-984`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L975-L984) — `registerPluginAction` already validates the canonical form
- [`src/hooks/usePluginToolbarButtons.ts:77`](https://github.com/daintreehq/daintree/blob/develop/src/hooks/usePluginToolbarButtons.ts#L77) — renderer-side dependency on the `plugin.` prefix

## Constraints

- Lazy import only — never resolve handler modules at load time. Activation has a 5s budget; twenty commands cannot pay twenty `import()` costs up front.
- Extension precedence `.ts`, `.tsx`, `.js`, `.mjs`. Single `default` export, no named `handler` fallback — keep it one shape.
- Namespace migration ships in one commit so toolbar buttons and the renderer prefix check don't get out of sync mid-PR.

## Depends on

- #9277 — `host.registerAction(descriptor, handler)` provides the descriptor shape this binds against
- #9272 — namespace-ownership reservation (so collision rejection has something to check against)

Under epic #5651.

gregpriday:
Closes #9281.

Implements two coordinated pieces of plugin infrastructure.

## 1. Manifest command binder (lazy handlers)

Adds `contributes.commands` to the plugin manifest schema (`CommandContribution`: `name`/`title`/`description`/`category` required; `kind` defaults to `"command"`, `danger` to `"safe"`). At load, each command:

- registers a synthetic `{pluginId}.{name}` action through the existing `registerPluginAction` path, and
- probes `src/{name}.{ts,tsx,js,mjs}` (first match wins) for a handler file.

The handler module is imported **lazily — only on first dispatch** — and its default export is memoized, keeping plugin activation within the 5s budget. Commands with no handler file still appear in the palette but throw `Command "{id}" has no handler` when dispatched (never at load). A command id that collides with a built-in action id is rejected at load and skipped.

## 2. Namespace standardization

Toolbar buttons were registered under a divergent `plugin.{pluginId}.{id}` prefix while everything else used the canonical `{pluginId}.{actionId}`. Reconciled in one atomic change:

- new `pluginActionId(pluginId, name)` helper in `shared/types/plugin.ts` (single source of truth for the namespace);
- `PluginToolbarButtonId` relaxed from `` `plugin.${string}` `` to `` `${string}.${string}` ``;
- registry guard (`isRegisteredPluginButton`) and renderer prefix check (`usePluginToolbarButtons`) simplified to registry/membership lookups;
- persisted `pinnedButtons` sweep switched to a dot discriminator (built-in toolbar/agent ids are all dot-free);
- v9 toolbar-prefs migration strips stale `plugin.*` pinned keys.

## Notes

- Dependency issues #9277 (imperative `host.registerAction`), #9271 (provenance `loadError`), #9272 (namespace reservation) are still open; this PR reuses existing infrastructure (`registerPluginAction`, `BUILT_IN_ACTION_IDS`, `console.error` for collisions) and does not block on them.
- The imperative `host.registerAction` escape hatch documented in `contribution-points.md` is #9277 scope; the filesystem-convention binder shipped here is the primary path.

## Tests

- **Manifest commands**: descriptor registration, kind/danger defaults, lazy fs dispatch + caching, extension precedence, no-handler / no-default-export / handler-throws errors, built-in collision skip, duplicate-name survival, unload cleanup.
- **Schema**: `commands` default `[]`, kind/danger defaults, name-pattern rejection.
- **Store**: v9 migration preserves new-format pins while stripping `plugin.*`.

### DIFF
--- a/docs/plugins/contribution-points.md
+++ b/docs/plugins/contribution-points.md
@@ -196,7 +196,7 @@ Toolbar buttons dispatch an existing action from the main toolbar.
 
 | Field | Required | Notes |
 | --- | --- | --- |
-| `id` | yes | Namespaced at runtime as `plugin.{pluginId}.{id}`. |
+| `id` | yes | Namespaced at runtime as `{pluginId}.{id}`. |
 | `label` | yes | Hover tooltip. |
 | `iconId` | yes | Registered icon ID. |
 | `actionId` | yes | Fully-qualified action ID, including plugin namespace. Built-in actions (e.g. `terminal.new`) also work. |
--- a/electron/schemas/plugin.ts
+++ b/electron/schemas/plugin.ts
@@ -4,6 +4,7 @@ import { BUILT_IN_PLUGIN_PERMISSIONS } from "../../shared/types/plugin.js";
 import type {
   PluginManifest,
   PanelContribution,
+  CommandContribution,
   ToolbarButtonContribution,
   MenuItemContribution,
   ViewContribution,
@@ -28,6 +29,20 @@ export const PanelContributionSchema = z.object({
   canRestart: z.boolean().default(false),
   canConvert: z.boolean().default(false),
   showInPalette: z.boolean().default(true),
+});
+
+const COMMAND_NAME_PATTERN = /^[a-z0-9][a-z0-9-]*$/;
+
+export const CommandContributionSchema = z.object({
+  name: z.string().min(1).max(64).regex(COMMAND_NAME_PATTERN, {
+    error: "Command name must match /^[a-z0-9][a-z0-9-]*$/",
+  }),
+  title: z.string().min(1),
+  description: z.string().min(1),
+  category: z.string().min(1),
+  kind: z.enum(["command", "query"]).default("command"),
+  danger: z.enum(["safe", "confirm"]).default("safe"),
+  keywords: z.array(z.string().min(1)).optional(),
 });
 
 export const ToolbarButtonContributionSchema = z.object({
@@ -139,6 +154,7 @@ export const PluginManifestSchema = z
     contributes: z
       .object({
         panels: z.array(PanelContributionSchema).default([]),
+        commands: z.array(CommandContributionSchema).default([]),
         toolbarButtons: z.array(ToolbarButtonContributionSchema).default([]),
         menuItems: z.array(MenuItemContributionSchema).default([]),
         views: z.array(ViewContributionSchema).default([]),
@@ -148,6 +164,7 @@ export const PluginManifestSchema = z
       })
       .default({
         panels: [],
+        commands: [],
         toolbarButtons: [],
         menuItems: [],
         views: [],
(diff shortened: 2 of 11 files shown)


gregpriday:
Closing — out of Wave 1 scope. Will revisit.

Issue #9232, opened 2026-05-27
Structured audit log for plugin-action dispatch


gregpriday:
The audit-log primitive is already shipped on the MCP side and there's no equivalent for plugin actions. [`ActionService.dispatch`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/services/ActionService.ts) emits only console and Sentry on error for plugin-contributed actions.

The MCP precedent is concrete:

- [`McpAuditRecord` / `McpLogRecord` / `McpGrantRecord` at `shared/types/ipc/mcpServer.ts:105`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/shared/types/ipc/mcpServer.ts#L105) define a structured audit format with `turnId` correlation.
- [`McpServerSettingsTab` at `src/components/Settings/McpServerSettingsTab.tsx:76-172`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Settings/McpServerSettingsTab.tsx#L76-L172) surfaces records, stats, copy and export, and an `auditEnabled` toggle.

The scope here is extending that pattern to plugin-action dispatch, not building a new audit system.

### `PluginActionAuditRecord` shape

Mirror `McpAuditRecord`:

```ts
export interface PluginActionAuditRecord {
  recordId: string;
  pluginId: string;
  actionId: string;
  source: ActionSource;
  danger: ActionDanger;
  args_hash: string;
  ts: number;
  turnId?: string;
}
```

Hash args by default (SHA-256). A developer-only setting can opt into plaintext for debugging; the privacy default is hashed so logs don't leak file paths, prompts, or tokens.

### Wiring

`ActionService.dispatch` emits a record when the action's `pluginId` is non-null. The existing audit-log storage backend persists it. Add a Plugin actions sub-tab or section under the existing MCP audit tab in `McpServerSettingsTab` with the same copy and export affordances. Emit one record per restricted-capability use as well, paired with the capability gate in the typed channel registry — record both grants and denials.

### Related work

Open issue #9157 ("Show requesting-bearer identity in MCP confirm dialog") is the parallel work on the MCP side. This is the plugin-side complement.

### Dependency

Best landed after #9231 so the capability-gate decisions are recordable. The audit-record shape and emission can ship first; the restricted-capability grant/deny records depend on the gate existing.

PR #9402, opened 2026-05-27
feat(plugins): plugin diagnostics export — host.logger + report-issue payload (#9291)

### LINKED ISSUE #9291: Plugin diagnostics export — host.logger plus report-issue payload
Two scopes, one issue because they share a buffer. (1) `host.logger` — a per-plugin tagged log channel a plugin's `activate(host)` writes to during normal operation. (2) Diagnostics export — the plugin slice that gets bundled into the report-issue payload built by `src/components/ErrorBoundary/buildReportIssueUrl.ts:130`. Without either, plugin bugs are invisible: a misbehaving plugin can't surface its own errors to the user, and a Daintree bug report from a user running plugins arrives with no plugin context for the maintainer.

### Scope 1 — `host.logger`

Add `logger` to the `PluginHostApi` surface at `shared/types/plugin.ts:191-262` with three methods: `info(message: string, fields?: Record<string, unknown>)`, `warn(...)`, `error(...)`. Each method writes to a per-plugin ring buffer (~500 lines, evicted FIFO) keyed by `manifest.name`. The buffer lives in main next to the existing plugin state, owned by `PluginService`. Forwarding to the host console (`console.info` / `warn` / `error`) is on by default with the message prefixed `[plugin:${pluginId}]` so terminal-attached devs see it; suppressing the console forward is a follow-up if it gets noisy.

The ring buffer surfaces in #9290's per-plugin detail view — expand a plugin row, see the last 500 log lines in reverse-chronological order. The detail view also displays whatever load-error was caught at `PluginService.ts:478-` (the `try/catch` around `import(pathToFileURL(...))`).

### Scope 2 — Diagnostics export

The existing `buildReportIssueUrl` builds the report payload; today it doesn't include plugin state. Extend its input to accept a `pluginDiagnostics` slice with: list of loaded plugins (manifest name, version, source, installedAt, devMode flag), recent action audit tail from #9232 (last ~50 plugin-action dispatches per plugin with timestamps and `effectiveDanger`), and the last 500 log lines per plugin from Scope 1's buffer. The slice is gathered by a new IPC handler that snapshots all three sources atomically and returns them to the renderer for inclusion in the report body.

When no plugins are loaded, the section is omitted entirely — the report stays compact for users not running plugins. When plugins are loaded, the section appears between the existing system info and the user-supplied description, marked clearly so a maintainer can scan past it.

## Evidence

- [`shared/types/plugin.ts:191-262`](https://github.com/daintreehq/daintree/blob/develop/shared/types/plugin.ts#L191-L262) — `PluginHostApi` surface (no `logger` member today)
- [`electron/services/PluginService.ts:478`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L478) — load-error catch site (the error the detail view shows)
- [`src/components/ErrorBoundary/buildReportIssueUrl.ts:130`](https://github.com/daintreehq/daintree/blob/develop/src/components/ErrorBoundary/buildReportIssueUrl.ts#L130) — `buildReportIssueUrl` entry point (extension target for Scope 2)
- [`src/components/ErrorBoundary/ErrorBoundary.tsx:166`](https://github.com/daintreehq/daintree/blob/develop/src/components/ErrorBoundary/ErrorBoundary.tsx#L166) — call site that builds the URL on the "Report issue" path
- #9232 — plugin-action audit log (this issue consumes its tail)

## Constraints

- Ring buffer size is fixed at ~500 lines per plugin. Per-line cap (~2KB after structured stringify) prevents a plugin from `logger.error`-bombing the host into memory pressure. Drop lines past the cap with a single truncation marker, not silent loss.
- `logger` writes are synchronous from the plugin's POV (no Promise) — a plugin author calling `host.logger.info(...)` inside a tight loop must not be backpressured. The actual write is enqueued and the call returns immediately.
- The diagnostics export must not include secrets stored via #9279's `host.settings` — even if a plugin's log line accidentally captures one. Easiest defense: log lines pass through a regex scrub for common secret patterns (`bearer`, `sk-`, `gh[pousr]_`, etc.) before being added to the report payload. Per-buffer storage is the raw write; only the export pass scrubs.
- Per CLAUDE.md `notify()` rule: plugin errors caught at load-time (Scope 1's load-error detail view) do *not* surface as toasts. The detail view in #9290 is the discovery surface; an inert plugin doesn't need a notification because the only action is to open Settings. This is Tier 0 / Tier 1 territory, not Tier 2.

## Depends on

- #9271 — provenance fields (the diagnostics slice reads them per plugin)
- #9232 — plugin-action audit log (the export consumes its recent-actions tail)

## Blocks

- #9290's per-plugin detail view consumes Scope 1's ring buffer — #9290 can ship with an empty detail panel and this issue fills it in.

Under epic #5651.

gregpriday:
## Summary

- Adds `host.logger` (`info`/`warn`/`error`) to the plugin host API, backed by a per-plugin 500-entry FIFO ring buffer in `PluginService`. Not revoke-guarded so log writes during teardown aren't silently dropped. Message and `fields` each capped at ~2KB UTF-8 with a truncation marker; all writes forward to the host console with a `[plugin:id]` prefix.
- Adds `getDiagnosticsSnapshot()` and a new `plugin:get-diagnostics` IPC channel. Secrets are scrubbed on export only (via `shared/utils/pluginDiagnosticsScrub.ts`), not on write, so the in-memory buffer holds the raw lines.
- Extends `buildReportIssueUrl` with an optional collapsible, budget-gated plugin section; `ErrorBoundary` fetches the snapshot best-effort with a 2s timeout so a slow IPC call never blocks the error UI.

`#9271` (provenance) and `#9232` (audit log) are still open — those fields use graceful fallbacks for now.

Resolves #9291

## Changes

- `shared/types/plugin.ts` — `PluginLogLine`, `PluginDiagnosticsEntry`, `PluginDiagnosticsSnapshot` types; `logger` added to `PluginHostApi`
- `shared/utils/pluginDiagnosticsScrub.ts` — shared scrub helper (bearer tokens, `sk-*`, `gh[pousr]_*`, etc.)
- `electron/services/PluginService.ts` — `PluginLogBuffer` ring buffer class; `appendLog`, `getDiagnosticsSnapshot`, `getPluginLoadError` methods
- `electron/ipc/channels.ts` + `electron/ipc/handlers/plugin.ts` + `electron/ipc/handlers/plugin.preload.ts` — `plugin:get-diagnostics` channel
- `shared/types/ipc/generated-api.ts` + `generated.ts` — updated IPC surface
- `src/components/ErrorBoundary/buildReportIssueUrl.ts` — plugin section with `<details>` budget gate
- `src/components/ErrorBoundary/ErrorBoundary.tsx` — best-effort diagnostics fetch with `DIAGNOSTICS_TIMEOUT_MS` constant

## Testing

Unit test coverage added for: scrub patterns (bearer, `sk-*`, `gh[pousr]_*`, nested objects), ring-buffer wraparound at exactly 500 and 501 entries, per-message 2KB cap with truncation marker, load-error scrub path, `buildReportIssueUrl` budget-gating, and NaN-timestamp guard in the scrub helper. `npm run check` and targeted vitest pass.

### DIFF
--- a/eager-import-baseline.json
+++ b/eager-import-baseline.json
@@ -1,6 +1,6 @@
 {
-  "count": 1258,
-  "moduleCount": 1258,
+  "count": 1259,
+  "moduleCount": 1259,
   "allowlist": [
     "electron/ipc/errorHandlers.ts",
     "electron/ipc/handlers/agentCapabilities.ts",
@@ -1046,7 +1046,7 @@
     },
     {
       "file": "electron/services/PluginService.ts",
-      "line": 325,
+      "line": 392,
       "pattern": "sync-store-get"
     },
     {
--- a/electron/__tests__/clipboard.test.ts
+++ b/electron/__tests__/clipboard.test.ts
@@ -8,6 +8,10 @@ vi.mock("electron", () => ({
   clipboard: { readImage: vi.fn(), writeImage: vi.fn(), writeText: vi.fn(), readText: vi.fn() },
   nativeImage: { createFromBuffer: vi.fn(), createFromPath: vi.fn() },
   ipcMain: { handle: vi.fn(), removeHandler: vi.fn() },
+  // The clipboard handler imports `projectStore`, whose module-level singleton
+  // calls `app.getPath("userData")` at construction — so the mock must provide
+  // `app` or the import throws before any test runs.
+  app: { getPath: vi.fn(() => "/tmp") },
 }));
 
 // clipboard.ts imports projectStore from ProjectStore.js which imports
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -786,6 +786,7 @@ export const CHANNELS = {
   PLUGIN_PANEL_KINDS_GET: "plugin:panel-kinds-get",
   PLUGIN_FORGE_PROVIDERS_GET: "plugin:forge-providers-get",
   PLUGIN_FILE_DECORATIONS_GET: "plugin:file-decorations-get",
+  PLUGIN_GET_DIAGNOSTICS: "plugin:get-diagnostics",
 
   // Config reload channels
   APP_RELOAD_CONFIG: "app:reload-config",
--- a/electron/ipc/handlers/__tests__/plugin.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/plugin.handlers.test.ts
@@ -7,6 +7,7 @@ const mockListPlugins = vi.fn();
 const mockListPluginActions = vi.fn();
 const mockRegisterPluginAction = vi.fn();
 const mockUnregisterPluginAction = vi.fn();
+const mockGetDiagnosticsSnapshot = vi.fn();
 
 vi.mock("../../../services/PluginService.js", () => ({
   pluginService: {
@@ -17,6 +18,7 @@ vi.mock("../../../services/PluginService.js", () => ({
     listPluginActions: (...args: unknown[]) => mockListPluginActions(...args),
     registerPluginAction: (...args: unknown[]) => mockRegisterPluginAction(...args),
     unregisterPluginAction: (...args: unknown[]) => mockUnregisterPluginAction(...args),
+    getDiagnosticsSnapshot: (...args: unknown[]) => mockGetDiagnosticsSnapshot(...args),
   },
 }));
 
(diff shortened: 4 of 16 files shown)


gregpriday:

PR #9422, opened 2026-05-27
feat(plugin): lazy-load plugins via activationEvents manifest field

### LINKED ISSUE #9270: Add activationEvents to manifest + defer dynamic import
Every plugin's `main` is dynamic-imported at scan time. `PluginService.ts:461-481` runs `await import(pathToFileURL(plugin.resolvedMain).href)` for every loaded plugin during `loadPlugin`, then immediately invokes its `activate()` inside `runActivate`. There is no triggering event — a plugin contributing only a single rarely-used menu item still loads and executes its main module on every app startup. As the curated plugin set grows, cold-start cost grows linearly with it.

The spec already documents the fix. `docs/plugins/manifest.md:162-168` describes `activationEvents` with `onStartupFinished` as the sole supported event in v1; every other activation is inferred from the contribution points. A plugin contributing a `commands[{ name: "foo" }]` entry activates implicitly when `foo` runs. A plugin contributing only a panel kind activates implicitly when its panel is first opened. A plugin needing background work with no user-triggered entry point opts in to `onStartupFinished`.

Concrete deliverables:

- Add `activationEvents: z.array(z.literal("onStartupFinished")).default([])` to `PluginManifestSchema` (`electron/schemas/plugin.ts:117-159`). Reject any other value loudly so plugin authors don't write `onCommand:foo` expecting it to work — per Codex, the explicit-event surface stays minimal in v1.
- Mirror on the type: `activationEvents?: ("onStartupFinished")[]` in `PluginManifest` (`shared/types/plugin.ts:89-108`).
- Refactor `loadPlugin` (`PluginService.ts:302-484`): the manifest read, validation, contribution-point registration, and `this.plugins.set(...)` must all happen at scan time, exactly as today. The `import(...)` at `PluginService.ts:461-481` moves out into a new `activatePlugin(pluginId)` method that runs once per plugin and is idempotent (early-return if already activated, queue the second caller behind the first promise).
- At end of startup, after the renderer signals ready, iterate plugins whose `activationEvents` includes `onStartupFinished` and call `activatePlugin` for each.
- Hook implicit activation: when a plugin-contributed panel kind is first instantiated, when a plugin-contributed menu item is invoked, when a `{pluginId}.{actionId}` action is dispatched (via F12), call `activatePlugin(pluginId)` first and await it before proceeding. Forge providers and file-decoration providers — which need their `register*Impl` to have run before the host queries them — also gate on activation; route the lookup through an `activate-then-resolve` path so the first call after startup pays the import cost and subsequent calls are direct.
- The `revoke()` window for `host.register*` calls (`PluginService.ts:486` onwards) still closes at the end of `activate()`. The contract that some methods are activation-only does not change.

## Evidence

- [`PluginService.ts:461-481`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L461-L481) — eager dynamic import + immediate activate
- [`PluginService.ts:273-300`](https://github.com/daintreehq/daintree/blob/develop/electron/services/PluginService.ts#L273-L300) — `loadFromDir`, runs `loadPlugin` for every directory in parallel
- [`docs/plugins/manifest.md:162-168`](https://github.com/daintreehq/daintree/blob/develop/docs/plugins/manifest.md#L162-L168) — `activationEvents` spec, `onStartupFinished` as the only v1 event
- [`shared/types/plugin.ts:264-266`](https://github.com/daintreehq/daintree/blob/develop/shared/types/plugin.ts#L264-L266) — `PluginActivate` signature
- [`electron/schemas/plugin.ts:117-159`](https://github.com/daintreehq/daintree/blob/develop/electron/schemas/plugin.ts#L117-L159) — `PluginManifestSchema`

## Constraints

- Scan-time still has to run the full manifest validation and contribution-point registration. Lazy activation only defers the JS import and the `activate()` call. The command palette, menus, and toolbars need to populate before any plugin code runs (per `docs/plugins/manifest.md:5`).
- Forge providers and file-decoration providers are special: they declare themselves via `contributes.*` but their *impl* binding happens during `activate()`. The first query against a not-yet-activated plugin's provider has to await activation transparently — a synchronous miss would cause a visible glitch.
- Activation errors must not poison the plugin record. A failing `activate()` should mark the plugin record `loadError` (see F4) but leave its declared contribution points routable on retry; a manual "Retry activation" from the Settings tab (F19) should work without restart.
- No `onLanguage`, `onView:*`, or `onCommand:*` explicit events in v1. Keep the surface narrow — extending it is non-breaking, narrowing it isn't.

## Depends on

- #9267 — settles whether activation may itself trigger a permission gate.
- #9268 — the field is added to the same schema; the renames land together.
- F4 — the new `loadError` slot on `LoadedPluginInfo` records activation failures.

Under epic #5651.

gregpriday:
## Summary
- Plugins can now declare `activationEvents` in their manifest. Only `onStartupFinished` is supported as an explicit event in v1 — every other activation is inferred from the contribution points.
- Plugin JS is lazy-loaded on first use instead of at scan time. `loadPlugin` runs manifest validation and contribution-point registration at scan time as before; `activatePlugin` defers the `import()` and `activate()` call until the plugin is actually needed.
- Implicit activation fires transparently on the first command invocation, panel open, forge provider query, file-decoration lookup, or action dispatch.

Resolves #9270

## Changes
- Schema: `activationEvents` field added to `PluginManifestSchema` (`electron/schemas/plugin.ts`) and `PluginManifest` type (`shared/types/plugin.ts`)
- `PluginService`: split `loadPlugin` (scan-time: manifest + contributions) from `activatePlugin` (lazy: JS import + `activate()`) with idempotency and queueing
- `forgeProviderRegistry` and `fileDecorationRegistry`: gate provider resolution on plugin activation — first query awaits activation transparently, subsequent calls hit the resolved cache
- Tests: unit + integration coverage for lazy activation, implicit trigger paths, and idempotency

## Testing
- Unit tests: `PluginService.test.ts` — activation idempotency, queueing, error marking
- Integration tests: `PluginService.integration.test.ts` — full scan + activate lifecycle, `onStartupFinished` trigger
- Registry tests: `fileDecorationRegistry.test.ts` — activation gating on decoration queries
- Typecheck, lint, and format all pass

### DIFF
--- a/electron/schemas/index.ts
+++ b/electron/schemas/index.ts
@@ -80,9 +80,10 @@ export {
 } from "./external.js";
 
 export {
-  PluginManifestSchema,
+  getPluginManifestSchema,
   PanelContributionSchema,
   PluginCapabilitySchema,
+  SCOPED_PLUGIN_NAME_PATTERN,
   type PluginManifest,
   type PanelContribution,
   type PluginCapability,
--- a/electron/schemas/plugin.ts
+++ b/electron/schemas/plugin.ts
@@ -145,53 +145,64 @@ export const PluginToastOptionsSchema = z
   })
   .strict();
 
-export const PluginManifestSchema = z
-  .strictObject({
-    name: z.string().min(1).max(64).regex(SCOPED_PLUGIN_NAME_PATTERN, {
-      error: 'Plugin name must be in publisher.name format (e.g. "acme.linear-context")',
-    }),
-    version: z.string().min(1),
-    displayName: z.string().optional(),
-    description: z.string().optional(),
-    main: z.string().optional(),
-    engines: z
-      .object({
-        daintree: z
-          .string()
-          .trim()
-          .min(1)
-          .refine((val) => semver.validRange(val) !== null, {
-            message: "engines.daintree must be a valid semver range",
-          })
-          .optional(),
-      })
-      .optional(),
-    capabilities: z.array(PluginCapabilitySchema).default([]),
-    contributes: z
-      .strictObject({
-        panels: z.array(PanelContributionSchema).default([]),
-        toolbarButtons: z.array(ToolbarButtonContributionSchema).default([]),
-        menuItems: z.array(MenuItemContributionSchema).default([]),
-        keybindings: z.array(KeybindingContributionSchema).default([]),
-        contextMenus: z.array(ContextMenuContributionSchema).default([]),
-        experimental_views: z.array(ViewContributionSchema).default([]),
-        experimental_mcpServers: z.array(McpServerContributionSchema).default([]),
-        forgeProviders: z.array(ForgeProviderContributionSchema).default([]),
-        fileDecorationProviders: z.array(FileDecorationContributionSchema).default([]),
-      })
-      .default({
-        panels: [],
-        toolbarButtons: [],
-        menuItems: [],
-        keybindings: [],
-        contextMenus: [],
-        experimental_views: [],
-        experimental_mcpServers: [],
-        forgeProviders: [],
-        fileDecorationProviders: [],
+export function getPluginManifestSchema(isBuiltin: boolean) {
+  return z
+    .strictObject({
+      name: z.string().min(1).max(64).regex(SCOPED_PLUGIN_NAME_PATTERN, {
+        error: 'Plugin name must be in publisher.name format (e.g. "acme.linear-context")',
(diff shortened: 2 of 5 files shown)


gregpriday:
Reopening — Wave 1 PR closed accidentally during merge sequence.

Issue #9196, opened 2026-05-27
VoiceOver loses live-region announcements while modals are open


gregpriday:
Per [Chromium Issue 354736464](https://issues.chromium.org/issues/354736464) (filed July 22 2024, still open), macOS VoiceOver ignores `aria-live` region announcements that sit outside an open `aria-modal="true"` dialog's DOM subtree.

Daintree's [`AccessibilityAnnouncer`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Accessibility/AccessibilityAnnouncer.tsx#L70-L75) renders its `sr-only` aria-live regions at the root of [`App.tsx`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/App.tsx#L744), so any call to `useAnnouncerStore.announce()` made while a modal dialog is open is silenced on macOS. Plenty of dialogs declare `aria-modal="true"` — `AppPaletteDialog`, `AppDialog`, `ErrorFallback`, `WorktreeOverviewModal`, `ReviewHub`, `ProjectSwitcherPalette`, `WebviewDialog`, `ThemeBrowser`.

A partial mitigation already exists in five destructive-action paths, where the dialog is closed before `announce()` fires:

- [`WaitingContainer.tsx#L147-L153`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Layout/WaitingContainer.tsx#L147-L153)
- [`BackgroundContainer.tsx#L225-L230`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Layout/BackgroundContainer.tsx#L225-L230)
- [`TerminalDestructiveActionConfirmDialog.tsx#L157-L163`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Terminal/TerminalDestructiveActionConfirmDialog.tsx#L157-L163)
- [`TerminalContextMenu.tsx#L385-L391`](https://github.com/daintreehq/daintree/blob/73402a609b85888f4fd9741ba56160f803d3e8bd/src/components/Terminal/TerminalContextMenu.tsx#L385-L391)
[...]

PR #9469, opened 2026-05-28
fix(a11y): simplify announcer tests and trim comments in ThemeBrowser and WebviewDialog

gregpriday:
## Summary

- Simplifies the structural accessibility tests in `ThemeBrowser` and `WebviewDialog` — the tests that shipped with #9434 were a bit more complicated than they needed to be; the new ones assert the same contract more cleanly
- Removes the verbose inline comments added alongside those tests in `ThemeBrowser.tsx` and `WebviewDialog.tsx`
- Moves the `AccessibilityAnnouncer` import in `ThemeBrowser.tsx` to be consistent with the other imports

## Changes

- `ThemeBrowser.tsx` — remove comment block, reorder import
- `WebviewDialog.tsx` — remove comment block
- `ThemeBrowser.test.tsx` — replace the detailed multi-assertion test with a simpler focused check on the assertive region (unique to the shared announcer)
- `WebviewDialog.test.tsx` — consolidate duplicate test, drop the non-focusable region loop

## Testing

64 unit tests pass across the 3 affected test files. `npm run check` is clean.

### DIFF
--- a/src/components/Browser/WebviewDialog.tsx
+++ b/src/components/Browser/WebviewDialog.tsx
@@ -1,4 +1,5 @@
 import { useState, useEffect, useRef, useCallback, useId } from "react";
+import { AccessibilityAnnouncer } from "@/components/Accessibility/AccessibilityAnnouncer";
 
 const TABBABLE_SELECTOR =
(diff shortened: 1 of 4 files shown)


gregpriday:
Closing as redundant. Issue #9196 was already resolved by #9434, which landed the same fix: co-locating `<AccessibilityAnnouncer />` inside the `ThemeBrowser` and `WebviewDialog` `aria-modal` subtrees, plus matching structural contract tests. This branch was cut before #9434 merged, so it just duplicates work that's already on `develop`. Nothing to add on top.

PR #9471, opened 2026-05-28
fix(error-boundary): surface scrubbed stack in production fallback

### LINKED ISSUE #9142: Production fallback hides the stack from crash reporters
When a user hits the production build's React error-boundary fallback, the `<details>Technical details</details>` block in `ErrorFallback.tsx` does not render. It is gated on `import.meta.env.DEV`. The user sees the error message, a copyable Error ID, and the `Report issue` button, but no stack at all. When they file a bug they have nothing to quote, and the Error ID alone often is not enough for first-pass triage.

Industry precedent here is well-established: VS Code's `Report Issue` and GitHub Desktop's React error boundary both keep the equivalent collapsed-by-default disclosure in production, with redacted file paths. Daintree's current behavior is strictly more conservative than either.

The disclosure needs to ship redacted: `<home>` and `<workspace>` placeholders for paths, known secret patterns scrubbed (the existing main-process scrubber in `electron/utils/secretScrubber.ts` already covers this). The `<details>` stays collapsed by default so non-technical users self-select away, and the summary label stays neutral (`Technical details`).

There is a related microcopy fix in the same flow. When the issue-body URL doesn't fit the budget and the full payload falls back to the clipboard, the overflow toast title currently reads `Error details too long`. The project microcopy rule for toast titles is verb-led past-tense or noun-phrase (`Crash report copied`, `Crash report didn't fit`), not the present-tense adjective form. The toast is action-free in this branch; the inline stub URL paste is itself the recovery surface.

Current entry points:

- [`<details>Technical details</details>` guard in `ErrorFallback.tsx`](https://github.com/daintreehq/daintree/blob/c4119817b9a54074c8daab8002c4121116b33c6d/src/components/ErrorBoundary/ErrorFallback.tsx#L189-L200)
- [overflow-clipboard `notify` call in `ErrorBoundary.tsx`](https://github.com/daintreehq/daintree/blob/c4119817b9a54074c8daab8002c4121116b33c6d/src/components/ErrorBoundary/ErrorBoundary.tsx#L196-L202)

Depends on #9140. The production stack disclosure needs a renderer-callable redaction surface to land first; the toast microcopy fix is independent and can ship alongside or earlier.

gregpriday:
## Summary

Follow-up refinement to the `ErrorFallback` and `ErrorBoundary` production stack fix.

- Narrows the disclosure condition to gate on `errorInfo.componentStack` only (was `error.stack || errorInfo.componentStack`) — if there's a component stack, there's a JavaScript stack too, so the prior OR was unnecessary
- Removes the `import.meta.env.DEV` conditional inside the `<pre>` block entirely; both `error.stack` and `errorInfo.componentStack` are always piped through `scrubReportText` before display
- Fixes the overflow-clipboard toast title from `"Report not copied"` to `"Crash report didn't fit"` (verb-led, past-tense microcopy)

## Changes

- `ErrorFallback.tsx` — simplified disclosure block, no DEV branch, always-scrubbed output
- `ErrorBoundary.tsx` — updated toast title
- `ErrorFallback.test.tsx` — consolidated tests: collapsed-by-default assertion, unique stack-frame assertion (`src/Test.tsx:10:5`), path-scrubbing regression for both stack and component stack
- `ErrorBoundary.test.tsx` — updated three toast-title assertions

## Testing

All 78 `ErrorBoundary` unit tests pass. `npm run check` (typecheck + lint ratchet + format + build) green.

Closes #9142

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -608,7 +608,6 @@ jobs:
           COLLECTED=$(find "$FALLBACK_DIR" -type f 2>/dev/null | wc -l)
           echo "[e2e] Fallback crash dump collection: $COLLECTED files"
 
-
       - name: Upload test results
         if: always()
         continue-on-error: true
--- a/src/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary/ErrorBoundary.tsx
@@ -194,7 +194,7 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
         } else {
           notify({
             type: "info",
-            title: "Report not copied",
+            title: "Crash report didn't fit",
             message:
               "Couldn't copy the full report. Quote the Error ID shown above when filing the issue.",
             inboxMessage: "Couldn't copy crash report to clipboard.",
--- a/src/components/ErrorBoundary/ErrorFallback.tsx
+++ b/src/components/ErrorBoundary/ErrorFallback.tsx
@@ -188,25 +188,15 @@ export function ErrorFallback({
           )}
         </div>
 
-        {(error.stack || errorInfo?.componentStack) && variant !== "component" && (
+        {errorInfo?.componentStack && variant !== "component" && (
           <details className="w-full mt-4">
             <summary className="cursor-pointer text-xs text-daintree-text/60 hover:text-daintree-text/80">
               Technical details
             </summary>
-            {/* Production stacks are scrubbed before display so crash reporters
-                never expose user paths or secrets; dev keeps the raw stack. */}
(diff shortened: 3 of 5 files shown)


gregpriday:
Closing this — #9142 was already fixed and merged via #9427 earlier today, so this PR is a duplicate.

I rebased onto develop to compare, and develop's version is the stronger one: it shows the Technical details disclosure when *either* error.stack or the component stack is present (this branch narrowed it back to component-stack-only), it keeps the raw stack in dev and only scrubs in production, and it ships broader test coverage (error.stack-only, both-stacks-absent, Windows paths). Merging this branch would regress all of that, so there's nothing worth pulling into a follow-up.

No action needed — #9427 covers the issue.

PR #9474, opened 2026-05-28
Add weekly stale-quarantine auto-issue cron

### LINKED ISSUE #9121: Add weekly stale-quarantine auto-issue cron mirroring nightly-failure pattern
We should have a weekly scheduled GitHub Actions workflow that lists structured `quarantine`-type annotations from the Playwright suite, parses the `YYYY-MM-DD` date from each description, and opens or comments on a per-spec tracking issue for any annotation older than 30 days.

Today there's no signal when a quarantined test rots in place. The blanket `test.skip(() => true, ...)` in `e2e/full/panels/core-browser-panel.spec.ts` for the find-in-page webview crash has been disabled indefinitely with no follow-up — the kind of thing this cron would surface every Monday until someone either fixes the spec or formally retires it.

## Label naming

The label should be `test-quarantine`, not `quarantine`. The word `quarantine` is already in use for project-state corruption cleanup in [`electron/services/projectQuarantineCleanup.ts`](https://github.com/daintreehq/daintree/blob/50ddea69628e1b7fddc0eef2c8270fa11a9df763/electron/services/projectQuarantineCleanup.ts) and overloading it would muddy both surfaces.

## Pattern to mirror

[`.github/workflows/nightly.yml`](https://github.com/daintreehq/daintree/blob/50ddea69628e1b7fddc0eef2c8270fa11a9df763/.github/workflows/nightly.yml#L611-L647) (lines 611-647) already contains the canonical `actions/github-script` `listForRepo` + `createComment`-or-`create` shape this workflow should reuse — same dedup-aware pattern, scoped per spec instead of per failure date. The `test-quarantine` label needs to exist before the first run (workflows shouldn't bootstrap labels safely).

## Constraints

- Don't apply `human-review` by default — flake triage is agent-doable per repo convention; the label is for spec-design decisions that genuinely need a developer.
- Issue title format should be sentence case with no terminal period, per the project's microcopy rules.

Depends on #9120 — the cron has nothing to parse until the structured-annotations migration lands.

gregpriday:
## Summary

- Adds a weekly Monday cron (`.github/workflows/stale-quarantine.yml`) that scans Playwright E2E specs for `type: "quarantine"` annotations, parses the leading `YYYY-MM-DD` date, and opens or comments on a per-spec tracking issue for any annotation older than 30 days — mirroring the nightly-failure `listForRepo` + `createComment`-or-`create` dedup shape (`nightly.yml` lines 611-647), scoped per spec.
- Parsing and staleness logic lives in a pure, unit-tested lib (`scripts/stale-quarantine-lib.mjs`, 28 Vitest tests in `scripts/stale-quarantine-lib.test.mjs`); the workflow dynamically imports it from `github-script` so the parsing is testable in isolation.
- Uses the `test-quarantine` label (created out-of-band on the repo; the workflow intentionally does not bootstrap labels). Issue titles are sentence case with no terminal period. Does not apply `human-review`.

## Notes

The 4 current quarantine annotations are all dated 2026-05-27, so the cron is wired correctly but stays quiet until they age past 30 days.

Hardening beyond the core issue: URL/`//`-safe comment stripping (so a `://` or bare `//` inside a description isn't truncated), rejection of calendar-overflow dates (e.g. `2026-02-31`), and future-dated annotations are surfaced as stale rather than silently passing.

## Test plan

- [x] `npx vitest run scripts/stale-quarantine-lib.test.mjs` (28 passing)
- [x] `npm run check` (typecheck, lint-ratchet, format) green
- [ ] First scheduled run (or manual `workflow_dispatch`) confirms issue creation against the `test-quarantine` label

Closes #9121.

### DIFF
--- a/.github/workflows/stale-quarantine.yml
+++ b/.github/workflows/stale-quarantine.yml
@@ -1,0 +1,95 @@
+name: Stale Quarantine
+
+on:
+  schedule:
+    # https://crontab.guru/#0_9_*_*_1 — every Monday 09:00 UTC
+    - cron: "0 9 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  scan:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Open or update stale-quarantine issues
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const path = require("node:path");
+            const { pathToFileURL } = require("node:url");
+            const { readFile } = require("node:fs/promises");
+
+            const lib = await import(
+              pathToFileURL(path.resolve("scripts/stale-quarantine-lib.mjs")).href
+            );
+            const { collectStaleQuarantines, buildIssueTitle, buildIssueBody } = lib;
+
+            const label = "test-quarantine";
+            const workspace = process.cwd();
+
+            const globber = await glob.create("e2e/**/*.spec.ts");
+            const absPaths = await globber.glob();
+
+            const files = [];
+            for (const absPath of absPaths) {
+              const source = await readFile(absPath, "utf8");
+              // Repo-relative POSIX path so issue titles stay stable across runners.
+              const specPath = path.relative(workspace, absPath).split(path.sep).join("/");
+              files.push({ path: specPath, source });
+            }
+
+            const stale = collectStaleQuarantines(files, new Date());
+            if (stale.length === 0) {
+              core.info("No stale quarantined tests found.");
+              return;
+            }
+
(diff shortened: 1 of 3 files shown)


gregpriday:
Closing as a duplicate of #9436, which already merged and resolved #9121 earlier today. No need for two stale-quarantine crons.

For the record, this branch took a different tack on two points that may be worth folding into the live workflow as a follow-up:

- It doesn't bootstrap the `test-quarantine` label. #9121 explicitly asked that the workflow not create labels ("workflows shouldn't bootstrap labels safely"); the merged version does it inline. The label already exists on the repo now, so the bootstrap step in #9436 is a no-op going forward either way.
- It extracts the annotation parsing / staleness logic into a pure, unit-tested module (`scripts/stale-quarantine-lib.mjs`, 28 Vitest cases) instead of inline ripgrep in YAML — matching the `scripts/*-lib.mjs` + test convention we use elsewhere (test-ratio, the budget checks, etc.).

Both are minor and the merged workflow is functional, so this is just a pointer, not a request to change anything.

PR #9692, opened 2026-06-03
fix(projectview): stop agent terminals flashing stale WebGL blocks on reactivation

### LINKED ISSUE #9679: Agent terminal flashes stale WebGL blocks on project reactivation
When a user switches back to a project that's been backgrounded for a while, the agent terminal briefly renders as solid blocks roughly where the content should be, then corrects itself a fraction of a second later. The Claude Code mascot collapses into a black block, text rows become bands of blocks. The geometry is right — the blocks land where real content is — but the glyphs are wrong.

This is a stale WebGL texture atlas. Backgrounded project views get frozen via the Chromium web lifecycle `frozen` state (`Page.setWebLifecycleState`) plus CDP CPU throttling, which leaves xterm's WebGL renderer holding an out-of-date texture atlas. A freeze is not a `webglcontextlost` event, so nothing marks the atlas dirty — the addon stays pooled and is assumed healthy.

On reactivation the reveal path reconciles grid geometry (which is why the blocks sit in the right places), but it schedules the WebGL restore behind a 100ms debounce, and that restore skips terminals whose addon is still pooled. So the first painted frame draws glyph cells from the stale atlas. Recovery happens incidentally — a later write, a texture-atlas merge, or a real context-loss event — which is why it clears so quickly.

Only agent terminals load the WebGL addon, so only they show this. Plain shells on the DOM renderer are unaffected.

## Desired behavior

Switching back to a backgrounded project shows correct terminal content on the first visible frame — no block flash.

Two parts:

**Part A — repair the atlas before reveal.** On reactivation from a frozen view, proactively clear or reacquire the WebGL texture atlas before the view is composited. It must not be gated behind the 100ms restore debounce, and it must not skip terminals whose addon is still pooled.

**Part B — extend the paint gate to warm reactivations.** The anti-flash paint gate today only covers cold-start switches; warm cached reactivation reveals the view immediately. Extend the gate so a warm reactivation stays hidden until one clean (post-atlas-clear) frame has painted.

Part A should fix the common case on its own. Part B guarantees no flash survives even on a slow paint, and covers any stale-frame source the atlas clear misses.

gregpriday:
Closes #9679.

When a cached project view is re-activated, Chromium can composite its pre-freeze GPU surface for a frame before the renderer's wake fan-out repairs each WebGL atlas. The result is a flash of stale glyph blocks in every agent terminal for that view.

This is a two-part fix.

## Part A: Proactive atlas repair

`TerminalWebGLManager` gains `repairAtlasForReactivation()`, which runs a resize-like `_clearModel` reset followed by a `refresh` on every live pooled terminal. This gives the WebGL state a clean slate before the async `wakeAndRestore` IPC fires. No shared-atlas clear (that would disturb co-owning tiled terminals), no context-loss breaker tick, no reacquire — just the minimal reset the existing resize path already uses.

## Part B: Warm anti-flash paint gate

Mirrors the proven cold-start z-order bridge. Rather than bringing the cached view straight to the front:

1. Reattach the cached view *behind* the outgoing view — occluded but unfrozen, so its rAF loop and wake sequence run immediately.
2. Wait for the renderer to signal a clean post-repair frame via the new re-fireable `APP_VIEW_WARM_PAINTED` IPC channel.
3. Only then detach the outgoing view.

A 1500ms hard-timeout fallback fires if the renderer stalls, so a wedged renderer can't block the switch indefinitely. `PaintGate` gains a `releaseChannel` discriminator so cold (`APP_VIEW_PAINTED`) and warm (`APP_VIEW_WARM_PAINTED`) signals can't cross-release each other.

### Why not `setBackgroundColor`?

`setBackgroundColor` only shows through transparent content — it can't cover an opaque stale page surface, so it doesn't solve the problem.

## Testing

- `TerminalWebGLManager` repair suite
- `fullWake` ordering (repair runs before async IPC)
- Warm paint gate: arm, release, hard-timeout, mismatch, cross-channel
- Z-order invariant (outgoing stays on top until gate releases)
- Wake finally-signal
- `warmReactivationGate` double-rAF unit test

Full typecheck, lint, IPC/channel guards green.

Note: not runtime-verified in the app — the WebGL freeze-resume flash is GPU/compositor timing-dependent and can't be reliably triggered in a controlled test run. Covered statically and with unit tests.


gregpriday:
The red `ci-ok` check is unrelated to this PR. The only failing shard is test (ubuntu 2/4).

Both failures are in `src/components/Sidebar/__tests__/SidebarContent.grid.test.ts` at lines 433 and 458, asserting cursor-ring CSS. This branch changes no sidebar files.

Commit 04fe30acd on `develop` ("fix(sidebar): remove focus/cursor rings from worktree rows") removed the cursor-ring CSS but didn't update that test, and `develop`'s own latest CI run fails on the same commit.

So this failure is pre-existing on `develop`, not introduced here.

Issue #8402, opened 2026-05-19
Eliminate the dynamic-import microtask in worktreeStore.ts


gregpriday:
There's a class of structural bugs in the worktree selection store that all come down to the same thing: stale microtasks.

Every transition in `src/store/worktreeStore.ts` (`setActiveWorktree`, `selectWorktree`, `enterFleetScope`, `exitFleetScope`) bumps a monotonic `_policyGeneration` integer and then fans out into async continuations through a single helper `loadTerminalStoreModule()` that dynamically imports `@/store/panelStore`. The dynamic import adds a microtask boundary even after the module is cached, so each continuation runs on a later tick and has to revalidate state via inline `get()._policyGeneration === generation` checks. Six such callsites in the file.

We've been fixing stale-microtask wrong-worktree bugs one at a time. The workarounds are documented in code comments at [worktreeStore.ts:605-608](https://github.com/daintreehq/daintree/blob/a50072be1c03f918d733b3a6c60a38efb2c94558/src/store/worktreeStore.ts#L605-L608) and [worktreeStore.ts:649-658](https://github.com/daintreehq/daintree/blob/a50072be1c03f918d733b3a6c60a38efb2c94558/src/store/worktreeStore.ts#L649-L658). The bug class only exists because the microtask boundary forces every continuation to revalidate.

The cyclic-import bind cited as the original reason for the dynamic import is not a hard module-load failure. [`panelStore.ts:32`](https://github.com/daintreehq/daintree/blob/a50072be1c03f918d733b3a6c60a38efb2c94558/src/store/panelStore.ts#L32) already imports `useWorktreeSelectionStore` at the top level and uses it at line 159 with no helper indirection.

## Two viable mechanisms

1. Extend the existing accessor surface in `src/store/storeAccessors.ts` to cover `usePanelStore.getState()` and `usePanelStore.setState(...)`, registered alongside `setPanelStoreAccessor` in `src/store/rendererStoreOrchestrator.ts`.
2. Take a direct top-level import of `usePanelStore`, matching the reverse-direction edge that already exists.

The pick depends on a regression test added under `src/store/__tests__/` that imports both stores in either order from a cold module graph using `vi.resetModules()`. Ship that test first as the precondition gate — it answers whether option 2 is safe, and if it is, the smaller change is the right one.

## Once the microtask is gone

[...]

PR #10141, opened 2026-06-06
docs: document sanctioned panel↔worktree direct-import exception in store-init-order

### LINKED ISSUE #9890: store-init-order.md forbids the panel↔worktree cycle the code ships on purpose
`docs/architecture/store-init-order.md` is the load-bearing reference for how renderer Zustand stores read each other without ESM cycles. It tells maintainers the exact opposite of what the code does, and it mis-states the invariant that actually keeps the code safe.

The doc says direct top-level imports between `panelStore`, `projectStore`, `worktreeStore`, and `fleetArmingStore` form an ESM cycle, and that the accessor module exists so **all** cross-store references route through a dependency-free leaf instead of importing each other:

> Direct top-level imports between any pair of these stores form an ESM cycle. […] The accessor-module pattern eliminates both classes by routing all cross-store references through a leaf module that imports no other store module.

The "When This Breaks" section names the crash and prescribes the cure:

> `ReferenceError: Cannot access 'usePanelStore' before initialization`
> Caused by directly importing `usePanelStore` at module level in a cyclic dependency graph. Fix by reading through `storeAccessors.ts`.

But the shipped code does precisely that, deliberately. `worktreeStore.ts` statically imports `usePanelStore` and `panelStore.ts` statically imports `useWorktreeSelectionStore` — a mutual module-level cycle introduced by #8402, proven safe and gated by a dedicated regression test:

```typescript
// src/store/worktreeStore.ts:8
import { usePanelStore } from "@/store/panelStore";

// src/store/panelStore.ts:34
import { useWorktreeSelectionStore } from "./worktreeStore";
```

The doc never mentions this sanctioned exception. Worse, it mis-states *why* the rest of the graph is safe. The real invariant is that neither store dereferences the partner's binding at module-**evaluation** time — every use sits inside a function called later (`usePanelStore.getState()` inside `applyWorktreeTerminalPolicy`, `getActiveWorktreeId` inside panelStore, etc.). The doc frames the rule as "no direct top-level imports between these stores," which is both contradicted by the code and dangerously incomplete: a maintainer who needs panelStore to read worktree state could add an eval-time / slice-construction-time `getState()` call, reasoning that the existing direct import already proves direct edges are fine, and reintroduce the exact TDZ the doc warns about. Nothing in the doc flags eval-time dereference as the actual landmine, and nothing points to the test that guards it.

Evidence:
- [`docs/architecture/store-init-order.md` L9-L14](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/docs/architecture/store-init-order.md#L9-L14) — "routing all cross-store references through a leaf module"
- [`docs/architecture/store-init-order.md` L106-L112](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/docs/architecture/store-init-order.md#L106-L112) — "Fix by reading through `storeAccessors.ts`"
- [`src/store/worktreeStore.ts` L8](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/src/store/worktreeStore.ts#L8)
- [`src/store/panelStore.ts` L34](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/src/store/panelStore.ts#L34)
- [`src/store/__tests__/worktreeStore.circularInit.test.ts`](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/src/store/__tests__/worktreeStore.circularInit.test.ts) — blesses the cycle (#8402) and gates both cold evaluation orders for init-safety
- #8402 introduced the blessed static-import cycle

The doc should document the panel↔worktree static-import cycle as a sanctioned exception, restate the real safety condition as "no partner-store binding may be dereferenced at module-evaluation time" rather than "no direct top-level imports," and reference `worktreeStore.circularInit.test.ts` as the regression gate. The "When This Breaks" guidance needs to separate an eval-time dereference (the real crash) from a benign closure-only direct import.

gregpriday:
## Summary

- `docs/architecture/store-init-order.md` had a blanket ban on direct top-level imports between renderer stores, but that contradicted code we deliberately shipped in #8402 — `panelStore` and `worktreeStore` have a mutual static import of each other.
- Corrected the doc: the precise invariant is that a direct static import is safe as long as neither store dereferences the partner's binding at module-evaluation time (all reads happen inside later-running function bodies). Adds a "Sanctioned exception" section for the `panelStore` ↔ `worktreeStore` pair, citing `src/store/__tests__/worktreeStore.circularInit.test.ts` as the cold-graph regression gate.
- Reconciled `docs/architecture/state-management.md` (hydration-ordering note and new-store checklist) so it no longer contradicts the authority doc. Fleet/project cross-refs still route through `storeAccessors` (`getFleetArmedIds`) — the carve-out is scoped to this one pair.

Resolves #9890

## Changes

- `docs/architecture/store-init-order.md` — replace blanket import ban with the mechanistic invariant; add sanctioned-exception section; reframe "When This Breaks" and "Red Flags" around eval-time dereference as the actual TDZ cause.
- `docs/architecture/state-management.md` — align hydration note and new-store checklist.

## Testing

Docs-only change. `prettier --check` passes on both files. No code touched.


gregpriday:
The red check is the shard 2 unit test job, failing on a single test in `BrowserToolbar.test.tsx` (the screenshot capture test looking for a `Capture screenshot` button).

That same test is failing on the current `develop` HEAD (`ad8eace68`), so it's pre-existing and unrelated to this docs-only change. Everything else is green. Something to look at separately when this gets merged.

gregpriday:
Update after rebasing onto `develop` HEAD (`d3e5d66e1`): the earlier `BrowserToolbar` shard-2 failure is resolved, but `develop` now has a separate regression that breaks all four unit-test shards: `TypeError: app.getPath is not a function` at `electron/setup/environment.ts:54`, hit at module-eval time when `electron/window/ProjectViewManager.ts` is imported in tests without a mocked Electron `app`.

It's not from this PR (the only diff vs `develop` is two `.md` files) and it reproduces on the `develop` HEAD CI run itself. Still something to fix on `develop` separately before this can go green.

PR #10175, opened 2026-06-06
refactor(worktree): route the attach-issue flow through the forge provider

### LINKED ISSUE #9993: Route the worktree attach-issue flow through the forge provider
The attach-an-issue-to-a-worktree flow is hardwired to GitHub end to end, even though the forge abstraction now covers everything it uses. On a non-GitHub forge (GitLab is tracked in #5167) the "Attach Issue" dialog can't work at all.

Three couplings, plus dead data:

- `IssuePickerDialog` fetches through `githubClient.listIssues` and renders the `GitHubIssue` shape (`issue.state === "OPEN"`). The picker only consumes `number`, `title`, `state`, and `url` — all present on the normalized forge `Issue`, and `forgeClient.listIssues(cwd, opts)` accepts the same `state` filter and a `search` option.
- The attach IPC contract (`AttachIssuePayload` / `IssueAssociation` in `shared/types/ipc/worktree.ts`) validates GitHub's raw uppercase enum — the handler rejects anything but `"OPEN" | "CLOSED"` — where the forge-normalized state is `"open" | "closed"`.
- `issueState` and `issueUrl` are persisted into `worktreeIssueMap` but never read back anywhere. The renderer mirror (`ManualIssueAssociation`) and the hydration consumer both carry only `{ issueNumber, issueTitle? }`. Dead fields in a persisted contract.
- The singular `worktree.getIssueAssociation` IPC op has zero callers — only `getAllIssueAssociations` is used (hydration in `WorktreeStoreContext`).

The picker should route through `forgeClient.listIssues` and the dead fields should come out of the attach contract (or be normalized if they're about to earn a reader).

One caveat to handle as part of the change: the built-in GitHub forge provider currently ignores `opts.search` — `listIssuesImpl` issues the unfiltered repository query, and its comment notes that wiring search means routing to `SEARCH_QUERY` and adding search back to the cache key together. The picker depends on debounced search, so the provider needs search wired through; without it the migration would silently degrade the dialog on the one provider that works today.

Evidence:

- [Picker on `githubClient` + `GitHubIssue` shape](https://github.com/daintreehq/daintree/blob/a9aa4a5d2f3865922bf7aff835b8482db240c0cc/src/components/Worktree/IssuePickerDialog.tsx#L95-L122)
- [GitHub enum in the attach contract](https://github.com/daintreehq/daintree/blob/a9aa4a5d2f3865922bf7aff835b8482db240c0cc/shared/types/ipc/worktree.ts#L27-L41)
- [Handler validating `"OPEN" | "CLOSED"`](https://github.com/daintreehq/daintree/blob/a9aa4a5d2f3865922bf7aff835b8482db240c0cc/electron/ipc/handlers/worktree/pull-requests.ts#L28-L60)
- [Renderer mirror carries only number/title](https://github.com/daintreehq/daintree/blob/a9aa4a5d2f3865922bf7aff835b8482db240c0cc/src/store/createWorktreeStore.ts#L60-L70)
- [Provider ignores `opts.search`](https://github.com/daintreehq/daintree/blob/a9aa4a5d2f3865922bf7aff835b8482db240c0cc/plugins/builtin/github/main/forgeProvider.ts#L466-L481)

Companion to #9959 (workflow actions' issue/PR lookup) and #9963 (current-user identity for self-assign) — this covers the remaining GitHub-bound leg of the issue→worktree flow, the manual attach dialog and its persistence contract.

gregpriday:
Closes #9993.

## What this does

Decouples the worktree attach-issue flow from GitHub-specific types and APIs, routing it through the provider-neutral forge layer.

- **`IssuePickerDialog`** now calls `forgeClient.listIssues` with the normalized `Issue` type (lowercase `open`/`closed` state) instead of `githubClient`/`GitHubIssue`. The `no-restricted-imports` disable is gone and the dialog is removed from the `githubClient` allowlist in `githubActions.adversarial.test.ts`.
- **`opts.search` is now honored by the GitHub forge provider.** A new `searchIssuesImpl` routes text search through `SEARCH_QUERY` (`repo:{owner}/{repo} is:issue [state:x] sort:created|updated-desc {term}`, truncated to GitHub's 256-char query cap). Search results use a `search:`-prefixed in-flight dedupe key and never write into `forgeIssueListCache`, so typed-input ephemera can't be served later as the background-poll list.
- **Dead fields removed from the IPC contract.** `issueState` (uppercase `OPEN`/`CLOSED`) and `issueUrl` are gone from `IssueAssociation`/`AttachIssuePayload`, the handler validation, and the `WorktreeCard` payload — nothing ever read them. No store migration needed: old persisted entries' extra keys are harmlessly ignored.
- **Dead `worktree:get-issue-association` IPC op removed end-to-end** (channel, handler, preload, api types, `worktreeClient`, regenerated `generated.ts`).
- **Coalesce-key fix (caught in review):** `listOptionsKey` in `forgeData.ts` now includes `opts.search`, so two debounced picker keystrokes can't share an in-flight slot and hand the second query the first query's results.

## Tests

- New `listIssues search` suite in `forgeProvider.test.ts`: query shape, state/sort qualifier mapping, 256-char truncation, list-cache isolation, identical-call coalescing, distinct-term isolation, whitespace fallback to the list path.
- IPC-layer regression test in `forgeData.handlers.test.ts` asserting distinct search terms get distinct coalesce slots (+ per-test rate-limit budget reset).
- Updated the outbox test payload factory for the slimmed `AttachIssuePayload`.

`npm run check` is fully green (typecheck, lint ratchet −1 warning, format, all IPC/channel/confirm-wiring guards).

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -15,7 +15,6 @@ export const CHANNELS = {
   WORKTREE_DELETE: "worktree:delete",
   WORKTREE_ATTACH_ISSUE: "worktree:attach-issue",
   WORKTREE_DETACH_ISSUE: "worktree:detach-issue",
-  WORKTREE_GET_ISSUE_ASSOCIATION: "worktree:get-issue-association",
   WORKTREE_GET_ALL_ISSUE_ASSOCIATIONS: "worktree:get-all-issue-associations",
   WORKTREE_HOST_DISCONNECTED: "worktree:host-disconnected",
   WORKTREE_RESTART_SERVICE: "worktree:restart-service",
--- a/electron/ipc/handlers/__tests__/forgeData.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/forgeData.handlers.test.ts
@@ -50,6 +50,7 @@ const storeMock = vi.hoisted(() => {
 vi.mock("../../../store.js", () => ({ store: storeMock }));
 
 import { registerForgeDataHandlers } from "../forgeData.js";
+import { _resetRateLimitQueuesForTest } from "../../utils.js";
 import { forgeAuditService } from "../../../services/forge/forgeAuditService.js";
 
 function findHandler(channel: string): (...args: unknown[]) => unknown {
@@ -91,6 +92,9 @@ const makePR = (n: number): PR => ({
 describe("registerForgeDataHandlers", () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    // The handlers share a per-channel rate-limit budget (10 calls / 10s); a
+    // suite-cumulative count would make later tests fail on call volume alone.
+    _resetRateLimitQueuesForTest();
     resolveForCwdMock.mockResolvedValue({
       namespaceId: "fake.provider",
       repoRef,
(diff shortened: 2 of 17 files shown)


gregpriday:
CI note: the test-shard failures here (BrowserToolbar, PtyClient adversarial, CrashRecoveryService) are pre-existing on develop at the merge-base (ad8eace68) — verified by running the same suites on a clean develop checkout. Nothing in this diff touches those areas. Tracked in #10183; once develop is green a re-run should clear ci-ok.

PR #10306, opened 2026-06-07
fix(replay): align cast harness with production ActivityMonitor options

### LINKED ISSUE #9870: Replay harness validates a detection path production agents never run
The cast replay harness (`electron/services/__tests__/replay/castReplayHarness.ts`) replays asciinema recordings through a real `@xterm/headless` terminal and the real `ActivityMonitor`, which is exactly the right shape for regression-guarding agent-state detection. But it diverges from production in ways that defeat the point:

- It pins `simpleOutputState: false` ([castReplayHarness.ts#L314-L325](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/electron/services/__tests__/replay/castReplayHarness.ts#L314-L325)) while every production agent monitor runs `true` via `buildActivityMonitorOptions` ([terminalActivityPatterns.ts#L240](https://github.com/daintreehq/daintree/blob/a93c9d06f1e4a00643bbfae2685fa59f33fc3369/electron/services/pty/terminalActivityPatterns.ts#L240)).
- It pins `idleDebounceMs` and `promptFastPathMinQuietMs` to the legacy 6000ms while the production floor is 8000ms.
- It never wires `getVisibleContentSnapshot`, which production always passes, so the `AgentActivityTemperature` cell-snapshot path is unexercised.
- It never routes OSC 9;4 progress sequences into `onOscProgressWorking` / `onOscProgressIdle`, even though real agent recordings carry them and production consumes them via `Osc94Parser` (Claude Code emits them constantly per #8701).

Net effect: the suite validates a code path production agents never run, and changes to the path they do run cannot fail any replay test. Replaying the existing fixtures under unmodified production options produces a completely different transition sequence than what CI asserts.

What this needs:

- [ ] Replay fixtures under unmodified `buildActivityMonitorOptions` output (production flags and timing floors), and recalibrate the five existing fixtures against that
- [ ] Wire the cell-snapshot accessor the way `TerminalProcess` does
- [ ] Route OSC 9;4 sequences from cast output into the monitor's progress callbacks
- [ ] Extend the replay through `AgentStateService` and the agent FSM so canonical `agent:state-changed` sequences are asserted, not just local monitor states. PR #6681 prototyped exactly this and was reverted to let the core detection changes settle (#6664); the core settled in May and it never re-landed
- [ ] Decide what happens to the legacy-path fixtures based on the direction chosen in #9869

The coarse-grained e2e spec (`e2e/full/terminal/terminal-agent-state-status.spec.ts`) does drive the full chain with a fake agent binary, so the chain isn't entirely unguarded, but it can't pin precise transition sequences from real recordings the way the replay harness can.

Related: #9869.

gregpriday:
## Summary
- The cast replay harness was validating a code path production agents never take. It pinned `simpleOutputState: false` while production runs `true` via `buildActivityMonitorOptions`, pinned legacy 6000ms timing floors while production uses 8000ms, never wired `getVisibleContentSnapshot`, and never routed OSC 9;4 progress sequences into the progress callbacks.
- Replays the existing fixtures under unmodified `buildActivityMonitorOptions`, wires the cell-snapshot accessor the way `TerminalProcess` does, and routes OSC 9;4 sequences into `onOscProgressWorking` / `onOscProgressIdle`. The harness now also drives the replay through `AgentStateService` and the agent FSM, so canonical `agent:state-changed` sequences are asserted, not just local monitor states.
- Adds a new `claude-osc94-working` fixture and a `claude-respawn` cast that exercises the full chain end-to-end. Addresses review findings on cleanup leak, confidence tolerance, and respawn coverage.

Resolves #9870

## Changes
- `electron/services/__tests__/replay/castReplayHarness.ts`: drop the option pins, route OSC 9;4 into the progress callbacks, wire `getVisibleContentSnapshot`, extend through `AgentStateService`, and add a `translateToString` shim on the local `CursorBufferLine` type.
- Fixtures: recalibrate all five existing `expected.json` files under the new production-equivalent options, add `claude-osc94-working.cast` and `claude-respawn.cast` plus their expected outputs, and add `claude-watchdog-timeout` for the agent-state-machine suite.
- `electron/services/__tests__/ActivityMonitor.replay.test.ts` and `AgentStateMachine.replay.test.ts`: new coverage pinning the canonical transition sequences.

## Testing
- Local CI gate passed: full static tier, 32,850 tests, build, and smoke on the gated SHA `9370b278fc1538d409d6cf782e4086d9d9472cb0`.
- Two parallel Codex adversarial review passes were run on the harness diff; all findings (cleanup leak in `IBuffer` disposal, over-tight confidence tolerance, missing respawn coverage) were fixed in the follow-up commits on the same branch.


gregpriday:
Pausing this one for now. PR #10301 merged into `develop` today and rewrote the same files this PR touches: the cast replay harness (`castReplayHarness.ts`) and the ActivityMonitor expected-JSON fixtures. The two took opposite stances on the production model.

#10301 has Claude and Codex turns emitting `completed` then `idle` with `trigger: "pattern"`. This PR re-asserts the older calibration: `simpleOutputState: true` and a single `idle` at timeout. Rebasing onto develop fails 7 of the 14 tests.

Reconciling would mean discarding the entire premise of this PR, which is a real decision about whether production fires `completed` at all, not just a textual merge. Marking it `human.review` so the author can decide whether #10301 supersedes this, or whether it should be reworked on top of develop's new harness and fixture calibration.

gregpriday:
Closing this. PR #10301 merged into `develop` today and rewrote the same files this PR touches with the opposite calibration, so it no longer rebases cleanly.

Rather than reconcile two opposite stances on the production model, I'm closing this and reopening #9870 for a fresh rework on top of current `develop`. The work needs redoing against the new harness and fixtures, including the OSC 9;4 routing that #10301 deferred to #9870.

PR #10358, opened 2026-06-11
fix(plugins): use the exported border-interactive token in catalog cards

gregpriday:
## Summary

Develop's CI has been red since `7b9183a4b` (plugin catalog view): the `colorSystem.contract` test fails because `PluginCatalog.tsx` uses `hover:border-daintree-border-interactive`, a utility with no matching `--color-daintree-border-interactive` export in `src/index.css`. The exported token is `--color-border-interactive`, so the correct utility is `hover:border-border-interactive` (same convention as `border-border-default` used across the app).

One-line fix; the previous class silently produced no CSS, so this also makes the intended hover border actually render.

## Verification

- `npx vitest run src/config/__tests__/colorSystem.contract.test.ts` — 15/15 pass with this change (failed before).

This unblocks `ci-ok` for every PR based on current develop, including #10355.


gregpriday:
Superseded — develop picked up the equivalent fix in 8e039ee73, so this is no longer needed.

PR #10375, opened 2026-06-11
perf(core): last-mile micro-optimizations across main and renderer

gregpriday:
This is a batch of last-mile performance work across the app, both real speedups and changes that just make it feel faster. It came out of a wide multi-agent audit of the codebase looking for small wins, and this covers the ones worth shipping now.

The terminal-data hot paths are in here too. Those changes feed agent-state detection and the highest-throughput code in the app, so they got a careful pass: pending output always flushes before any agent state transition, and each change carries its own tests.

### What's in this PR

- **Cool the terminal-data hot paths.** A 64KB cap on `SemanticBufferManager`'s pending-data accumulation, throttling `ActivityMonitor`'s full-buffer ANSI strip and pattern scan to a 30ms cadence with a trailing-edge run so the last chunk before quiet still gets scanned, coalescing the per-chunk `agent:output` IPC into 50ms batches that flush synchronously before any agent state transition (and on exit), and skipping the reflow keep-alive jitter when xterm's renderer is provably unpaused.
- **Batch the stage/unstage IPC in ReviewHub.** `handleStageFiltered`/`handleUnstageFiltered` were firing one IPC call and one git spawn per file, which also tripped the 30-call rate limiter on large changesets. They now use the existing `git.stageFiles`/`unstageFiles` batch channels, so N spawns collapse to one. Also added the missing `invalidateStagingDiffStatCache` to the batch handlers so the follow-up refresh doesn't serve stale stats.
- **Cache the branch list per repo.** The New Worktree dialog showed a full-body skeleton on every open while `listBranches` resolved. There's now a per-repo cache that seeds the form synchronously and revalidates in the background, so the dialog opens instantly after the first time.
- **Deduplicate git spawns in the workspace host.** Sibling worktrees of the same repo were each running their own `git fetch origin`. Added a recency window so they share a recent fetch instead. Dropped the redundant `rev-parse --verify` on every status pass and cached the divergence result so working-tree-only changes don't recompute it.
- **Get disk writes and eviction off the project-switch path.** Outgoing-state persistence and LRU eviction were running inline on every switch. Both are deferred now, with the persist still awaited before the IPC resolves so the contract holds.
- **Pre-warm the workspace host on hover.** Hover prefetch was priming hydrate data but not the host, so the click still paid the cold fork. It now pre-warms the host during the hover dwell, so the switch lands as a near-hit.
- **Fold presets and theme config into the boot payload.** Agent presets and theme config were each a separate post-mount IPC plus a disk read. They ride along the existing hydrate/boot payload now, the same way `tabGroups` and `terminalSizes` already do — and `releaseBootPayload` nulls both fields after hydration, so the cached payload doesn't retain them for the renderer's lifetime.
- **Full value cache on `agentSettingsClient`.** Burst paths (bulk restart, duplicate, recipe spawns) now read a cached snapshot instead of refetching. Writes invalidate it at call time and on settle, and `agentSettingsStore.refresh()` drops it first so external changes (cross-window writes, config reloads) always hit main.
- **Narrow Zustand subscriptions.** `useCloudSyncWarning`, the toolbar `AgentButton`, and `NewWorktreeDialog` were subscribed broadly enough to re-render on unrelated store ticks. Narrowed to primitive/`useShallow` selectors so worktree status churn stops touching them.
- **Optimistic settings toggles.** The GeneralTab toggles were visually frozen during the IPC round-trip. They flip immediately now and revert on failure, matching what PrivacyDataTab already does.
- **Unblock the main process.** Converted the `execFileSync` calls in `system:check-command` and the CLI availability probe to async, and parallelized the per-terminal validation probes with per-command dedup.

Plus a handful of smaller ones: debounced URL-history persistence, cached `getTmpDir`, an event-inspector serialization cache, an audio-device listener leak fix, idle-time chunk preloads, and a non-blocking panel-limit dialog.

Tests were updated alongside each change. CI is the gate.

### DIFF
--- a/electron/ipc/__tests__/app.state.test.ts
+++ b/electron/ipc/__tests__/app.state.test.ts
@@ -56,6 +56,7 @@ vi.mock("../../services/ProjectStore.js", () => ({
     getProjectStateWithRecovery: vi.fn(),
     saveProjectState: vi.fn(),
     enqueueProjectStateUpdate: vi.fn(async () => undefined),
+    readInRepoPresets: vi.fn(async () => ({})),
   },
 }));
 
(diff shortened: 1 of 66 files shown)


gregpriday:
A few things didn't make it into this PR. Leaving a note so whoever picks it up next has the full picture.

1. **The four terminal-data hot-path optimizations.** Deliberately deferred to a separate pass, since they all feed agent-state detection and I'd rather not break that on a perf change. These are the `SemanticBufferManager` accumulation cap, throttling the `ActivityMonitor` ANSI strip, coalescing the per-chunk `agent:output` IPC, and gating the reflow keep-alive on xterm's paused state.

2. **`releaseBootPayload` in `bootPromise.ts`** should null out the `appTheme` and `projectPresets` fields, so the cached boot payload doesn't retain them for the lifetime of the renderer.

3. **`agentSettingsClient` shipped as singleflight-only**, no value cache, because the invalidation hook in `agentSettingsStore.refresh` was outside the scope of this change. A full value cache is still on the table.

gregpriday:
Everything from my earlier note is done now, and I've updated the description to match.

The four terminal hot-path optimizations are in with tests: the `SemanticBufferManager` accumulation cap, the throttled `ActivityMonitor` ANSI strip, the coalesced `agent:output` IPC, and the paused-state gate on the reflow keep-alive. `releaseBootPayload` now also nulls `appTheme` and `projectPresets`. And `agentSettingsClient` has the full value cache, with writes and `agentSettingsStore.refresh()` invalidating it so external changes still hit main.

PR #10631, opened 2026-06-17
fix(terminal): prevent duplicate agent spawn on Enter at trust prompt

### LINKED ISSUE #10541: bug(terminal): Enter on empty HybridInputBar launches new agent instead of sending \r to PTY
## Summary

When a Claude agent terminal opens and shows the Claude Code workspace-trust prompt ("Accessing workspace / 1. Yes, I trust this folder / 2. No, exit"), pressing Enter does not confirm the prompt. Instead, a **new Claude terminal panel is spawned**.

## Steps to reproduce

1. Open Daintree with no project loaded (or open a new Claude terminal panel).
2. Wait for the Claude CLI to display the workspace-trust prompt:
   ```
   Accessing workspace:
   C:\Users\<user>

   Quick safety check: Is this a project you created or one you trust?
   > 1. Yes, I trust this folder
     2. No, exit

   Enter to confirm · Esc to cancel
   ```
3. Press **Enter**.

## Expected behavior

The keypress is forwarded to the PTY as a raw carriage return (`\r`), confirming option 1. Claude proceeds normally.

## Actual behavior

A second Claude terminal panel opens. The original panel remains stuck on the trust prompt.

## Root cause

The bug is a focus + routing mismatch in three steps:

### 1. Agent terminals focus the HybridInputBar, not xterm

`TerminalPane.tsx:957-982` — when a panel with an agent identity is focused, `getTerminalFocusTarget` resolves to `"hybridInput"` because `shouldShowHybridInputBar` returns `true` for any panel with an `effectiveAgentId`. The xterm buffer never receives keyboard focus.

### 2. Enter on an empty HybridInputBar calls `onSendKey("enter")`

`src/components/Terminal/hooks/useEditorKeymap.ts:120-128` — when the CodeMirror editor is empty (always true on first open) and Enter is pressed, this branch fires:

```ts
const text = editorViewRef.current?.state.doc.toString() ?? latest.value;
if (text.trim().length === 0) {
  if (latest.onSendKey) latest.onSendKey("enter");
  return true;
}
```

### 3. `"enter"` is interpreted as agent-launch, not a raw PTY keystroke

`TerminalPane.tsx:1529-1534` wires `onSendKey` to `terminalClient.sendKey(id, key)`. The string `"enter"` is treated by the agent lifecycle as an "activate/launch" signal, not a raw `\r` to the PTY — so a new panel spawns instead of the keypress reaching the running CLI process.

## Proposed fix

In `src/components/Terminal/hooks/useEditorKeymap.ts`, change the empty-input Enter path:

```diff
- if (latest.onSendKey) latest.onSendKey("enter");
+ if (latest.onSendKey) latest.onSendKey("\r");
```

Sending `"\r"` instead of `"enter"` bypasses the agent-launch interpretation and writes a raw carriage return to the PTY, where the running CLI (Claude Code trust prompt, or any other interactive CLI) can consume it normally.

## Affected files

- `src/components/Terminal/hooks/useEditorKeymap.ts` — `onEnter` handler, lines 120–128
- `src/components/Terminal/TerminalPane.tsx` — `onSendKey` prop, lines 1529–1534 (may need to handle `"\r"` if `terminalClient.sendKey` doesn't already pass raw strings through)

## Verification

1. Open a Claude agent terminal from cold state.
2. Wait for the workspace-trust prompt.
3. Press Enter — should confirm option 1 without opening a new panel.
4. Confirm that pressing Enter on an empty HybridInputBar during normal idle state (no CLI prompt active) still behaves correctly (no-op or raw `\r` to PTY — both acceptable).

dev-en-m:
## Summary

Fixes #10541 — launching a Claude agent from the toolbar, then pressing **Enter** at Claude's workspace-trust prompt, spawned a **second** agent panel instead of confirming the prompt.

## Root cause

A focus-steal race, not the keymap path the issue originally suspected (`sendKey` already writes `\r` correctly — the proposed `"\r"` change would have *broken* Enter, per the thread discussion):

1. `AgentButton.handleClick` dispatches `agent.launch` fire-and-forget with **no DOM focus handoff** — the toolbar `<button>` keeps focus.
2. The new `TerminalPane` mounts focused and schedules a one-shot RAF to focus the input bar. But `HybridInputBar` is `React.lazy()` inside `<Suspense>`; on a cold chunk load the RAF fires while `inputBarRef.current` is still `null`, so it silently no-ops.
3. DOM focus never leaves the button. At the trust prompt, the focused native `<button>` synthesizes a `click` on Enter (HTML spec) → `agent.launch` again → second panel.

The existing per-agentId reentrancy guard in `useAgentLauncher` doesn't help: it releases when the first launch completes, long before the trust-prompt Enter.

## Fix (defense-in-depth)

- **`HybridInputBar`** self-claims focus on mount when its panel is focused, so focus lands correctly even when the focus RAF lost the race. The `EditorView` is created in `useEditorFactory`'s `useLayoutEffect`, which flushes before this `useEffect`, so `focusEditor()` always finds a live view.
- **`AgentButton`** blurs the button on launch so a stranded-focus Enter can never bounce back into a duplicate launch — closing the same class of bug from any other focus-stranding path.

## Tests

Adds a `full-platform` E2E regression test (`core-hybrid-input-bar.spec.ts`): launch via the primary Start button, press Enter at the trust prompt **without** clicking the editor, and assert (1) focus landed in the input bar, (2) the carriage return reached the PTY (fake CLI flips to `FAKE_CLAUDE_READY`), and (3) exactly one agent panel exists.

## Verification

- ✅ `npm run check` — typecheck, all IPC/channel/confirm-wiring/plugin guards, lint ratchet (warnings decreased), Prettier — all green.
- ✅ New spec typechecks and lints clean; renderer build emits the fix.
- ⚠️ Runtime E2E could not be run locally (the author's Windows machine blocks the unsigned dev `electron.exe` via WDAC/Device Guard — an OS policy unrelated to this change). The regression test will run in CI's `full-platform` bucket.

### DIFF
--- a/e2e/full/platform/core-hybrid-input-bar.spec.ts
+++ b/e2e/full/platform/core-hybrid-input-bar.spec.ts
@@ -333,8 +333,10 @@ test.describe.serial("Core: HybridInputBar trust-prompt focus (#10541)", () => {
     // only after it receives a carriage return.
     await waitForTerminalText(regPanel, "FAKE_CLAUDE_READY", T_LONG);
 
-    // Invariant 3: exactly one agent panel exists — no duplicate spawn.
-    const idsAfter = (await getGridPanelIds(window)).filter((id) => !beforeIds.has(id));
-    expect(idsAfter).toEqual([panelId]);
+    // Invariant 3: exactly one agent panel exists — no duplicate spawn. Re-check
+    // after a settle window since the buggy second spawn was an async addPanel.
+    expect((await getGridPanelIds(window)).filter((id) => !beforeIds.has(id))).toEqual([panelId]);
+    await window.waitForTimeout(T_SETTLE);
+    expect((await getGridPanelIds(window)).filter((id) => !beforeIds.has(id))).toEqual([panelId]);
   });
 });
--- a/e2e/full/platform/core-hybrid-input-bar.spec.ts
+++ b/e2e/full/platform/core-hybrid-input-bar.spec.ts
@@ -270,3 +270,71 @@ test.describe.serial("Core: HybridInputBar", () => {
     await expect(restoredEditor).toHaveText(new RegExp(draftText), { timeout: T_MEDIUM });
   });
 });
+
+// Regression for issue #10541: launching Claude from the primary toolbar button
+// left DOM focus stranded on that button (the lazy HybridInputBar chunk hadn't
+// resolved when TerminalPane's focus RAF fired, so the RAF no-opped). Pressing
+// Enter at the trust prompt then re-fired the focused button's click handler,
+// spawning a SECOND agent panel instead of confirming the prompt. The fix has
+// HybridInputBar self-claim focus on mount and AgentButton blur on launch.
+test.describe.serial("Core: HybridInputBar trust-prompt focus (#10541)", () => {
+  let regCtx: AppContext;
+  let regPanel: Locator;
+
+  test.beforeAll(async () => {
+    // Fresh app instance so the HybridInputBar chunk is cold on first launch —
+    // that cold load is the exact window the focus race lived in.
+    prepareFixture();
+    regCtx = await launchApp({
+      env: {
+        PATH: `${fakeBinDir}${path.delimiter}${process.env.PATH ?? ""}`,
+        DAINTREE_CLI_PATH_PREPEND: fakeBinDir,
+        ANTHROPIC_API_KEY: "e2e-fake-key",
+      },
+    });
+    regCtx.window = await openAndOnboardProject(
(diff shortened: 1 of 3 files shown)


gregpriday (approved):
Thanks for tracking down the root cause here — the focus-race between `AgentButton` losing focus and `HybridInputBar` not claiming it on mount is exactly the kind of thing that looks like a keymap bug until you actually dig in.

I trimmed a few of the verbose inline comments and tightened the test settle logic on top of your changes. Nothing structural — just keeping the signal-to-noise ratio up.

Good find.

PR #10633, opened 2026-06-17
test(pty): cover untested Windows kill/shell paths and write-error throttle

dev-en-m:
﻿## Summary

Closes three test-coverage gaps in the Windows node-pty surface. Tests only — no production changes.

### 1. ProcessTreeKiller — Windows `taskkill` path (false-green fix)

Every test in `ProcessTreeKiller.test.ts` began with `if (process.platform === "win32") { return; }`, so on Windows the suite passed **vacuously** — the actual Windows kill path (`taskkill /T /F /PID`, which tears down the entire agent process tree) was never asserted.

- Wrapped the POSIX SIGTERM/SIGCONT/SIGKILL cascade suite in `describe.skipIf(win32)` so it now reports as **skipped** on Windows instead of a misleading pass.
- Added a `Windows taskkill` suite that exercises the win32 branch under a mocked platform + mocked `spawnSync` (no real process killed): exact args/options, post-kill `ptyProcess.kill()`, absence of POSIX signals, error swallowing for thrown `taskkill`/`kill`, and the `shellPid <= 0` short-circuit.

### 2. terminalShell — Windows shell discovery

`findWindowsShell()` and `getDefaultShell()`'s win32 delegation were untested (only `getDefaultShellArgs` had coverage). Added tests for the fallback chain `pwsh.exe → powershell.exe → COMSPEC → cmd.exe` via mocked `execFileSync`.

### 3. TerminalProcess — write-error throttle

`logWriteError`'s 5s throttle window, the "Suppressed N…" summary, the counter reset, and `traceId` passthrough had no coverage. Added a dedicated suite driving the public `write()` fast path with a throwing pty.

## Verification

- Full PTY suite: 1213 passed / 17 skipped (no regressions)
- `ProcessTreeKiller.test.ts`: 6 Windows tests pass, 9 Unix tests skipped on Windows
- `npm run check` (typecheck + IPC/channel/confirm-wiring guards + lint ratchet + format): exit 0
- Anti-vacuous check: dropping `/T` from the expected taskkill args makes the test fail, confirming the assertion is load-bearing

### DIFF
--- a/electron/services/pty/__tests__/ProcessTreeKiller.test.ts
+++ b/electron/services/pty/__tests__/ProcessTreeKiller.test.ts
@@ -1,7 +1,27 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { spawnSync } from "child_process";
 import type { IPty } from "node-pty";
 import type { ProcessTreeCache } from "../../ProcessTreeCache.js";
 import { ProcessTreeKiller } from "../ProcessTreeKiller.js";
+
+vi.mock("child_process", () => ({ spawnSync: vi.fn() }));
+
+const spawnSyncMock = vi.mocked(spawnSync);
+
+const originalPlatformDescriptor = Object.getOwnPropertyDescriptor(process, "platform");
+
+function setPlatform(platform: NodeJS.Platform): void {
+  Object.defineProperty(process, "platform", {
+    value: platform,
+    configurable: true,
+  });
+}
+
+function restorePlatform(): void {
+  if (originalPlatformDescriptor) {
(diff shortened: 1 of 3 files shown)


gregpriday (approved):
Good find on these gaps. The `ProcessTreeKiller` `taskkill` path was passing vacuously — no assertions, just silent skips — and `findWindowsShell` and the write-error throttle had no coverage at all.

I've pushed a couple of additional suites on top: `undefined`/negative `shellPid` short-circuit, PowerShell probe order, and tighter kill-ordering assertions to make sure the sequence is load-bearing.

Everything looks solid. Approving.

Issue #10745, opened 2026-06-23
trigger not found for some reason


mai1015:
<img width="1737" height="1092" alt="Image" src="https://github.com/user-attachments/assets/27b0a641-9177-4e78-86ba-fc38ef508f9f" />

gregpriday:
Hi again, and thanks for another one. Good to see you back.

That screen is Daintree itself, not anything you did wrong. The "Not Found" you're seeing is the app failing to load one of its own internal pages, so the window gets stuck on a bare error screen instead of opening your project. That's a bug on our side.

My best guess is it happens when a project view reloads while Daintree is under memory pressure, which lines up with your earlier reports about memory use (#10498) and switching between projects (#10660). When a reload like that doesn't come back cleanly, the app should recover on its own and re-open the view, not strand you on this screen. Right now it doesn't, and that's the part I want to fix.

A few questions to help me pin it down:

- What were you doing right before it showed up? Switching between projects, opening one, or had you just launched Daintree?
- Does reloading the window, or quitting and restarting Daintree, bring it back to normal?
- Does it happen often, or was this a one-off?

Thanks again for the report. These are genuinely helpful.

mai1015:
switching between project cause this. i have to restart as non of the button or menu works.

gregpriday:
I've been doing quite a lot of work on project switching after your previous reports. Basically making sure that Daintree can handle more projects. So let's see if the next update resolves some of these issues. There'll be a 0.20 update this week.

PR #10802, opened 2026-06-26
Guard project view loads against missing bootstrap

dev-en-m:
## Summary
- Reject project view loads when the renderer reaches `did-finish-load` without exposing the expected `window.__DAINTREE_INITIAL_PROJECT__.id` bootstrap value.
- Route that failure through the existing project-switch rollback path so the previous working project view remains active instead of accepting an internal wrong page.
- Add a regression test for the wrong-page / missing-bootstrap load case.

## Related Issue
- Addresses daintreehq/daintree#10745. The issue reports Daintree getting stuck on an internal `Not Found` screen while switching projects; the maintainer confirmed this is Daintree failing to load one of its own internal pages.

## Platform
- Validated locally on Windows.

## Tests
- `CI=true npm run test -- electron/window/__tests__/ProjectViewManager.adversarial.test.ts electron/window/__tests__/ProjectViewManager.backgroundResize.test.ts electron/window/__tests__/ProjectViewManager.eviction.test.ts electron/window/__tests__/ProjectViewManager.focusIntent.test.ts electron/window/__tests__/ProjectViewManager.freeze.test.ts electron/window/__tests__/ProjectViewManager.gc.test.ts electron/window/__tests__/ProjectViewManager.paintGate.test.ts electron/window/__tests__/ProjectViewManager.preload.test.ts electron/window/__tests__/ProjectViewManager.recovery.test.ts electron/window/__tests__/ProjectViewManager.switchFailure.test.ts --maxWorkers=1`
- `npm run typecheck`

### DIFF
--- a/electron/window/ProjectViewManager.ts
+++ b/electron/window/ProjectViewManager.ts
@@ -1598,7 +1598,12 @@ export class ProjectViewManager {
         settle(() => reject(new Error("View load timed out")));
       }, LOAD_TIMEOUT_MS);
 
-      const onFinish = () => settle(() => resolve());
+      const onFinish = () => {
+        void this.verifyProjectBootstrap(wc, projectId).then(
+          () => settle(() => resolve()),
+          (error) => settle(() => reject(error))
+        );
+      };
       const onFail = (_event: Electron.Event, errorCode: number, errorDescription: string) =>
         settle(() => reject(new Error(`View load failed: ${errorDescription} (${errorCode})`)));
       const onPreloadError = (_event: Electron.Event, _preloadPath: string, error: Error) =>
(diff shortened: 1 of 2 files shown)


gregpriday (approved):
Nice work tracking down the root cause on #10745. The guard in `ProjectViewManager` is exactly the right place for it: if `did-finish-load` fires without `window.__DAINTREE_INITIAL_PROJECT__.id`, we reject and route through the existing switch rollback path rather than accepting a broken view.

I added two test commits to lock in the edge cases: wrong-project bootstrap mismatch and a mid-verify probe rejection. Both feel like the kind of thing that could silently regress, so good to have them covered.

Looks good to ship.

gregpriday:
Thanks for this, @dev-en-m — the bootstrap guard is a clean fix for #10745.

Your fork looks like it's been deleted, so this PR's branch can't be updated or merged from here anymore. I've re-pushed your branch to the main repo and continued the work in #10804, which keeps your original commit (`37f2110fd`) intact and adds a couple of regression tests (wrong-project bootstrap + bootstrap-probe rejection mid-verify). Closing this one in favor of that — full credit to you for the fix.

PR #10948, opened 2026-07-04
feat(pty-host): lossless terminal memory diet with state-aware retention tiers

gregpriday:
This PR puts host-side terminal memory on an explicit budget. Since we removed hibernation in #10811, every background terminal stays fully live, which means the pty-host's real memory cost is in the analysis buffers we keep per terminal: the headless mirror scrollback, the semantic buffer, the forensics ring, the agent-output mirror, and preserved exit snapshots. Until now those were all sized by incidental constants. Now they're bounded by policy, keyed to what each terminal is actually doing.

The invariant that matters: this never touches the renderer. The live xterm stays the visual source of truth, no PTY output is paused or dropped for memory savings, and every trim is recorded as retention trimming in diagnostics, clearly separate from the `data-loss` path. If a user scrolls back, they see the renderer's own scrollback exactly as before.

## Retention tiers

Four tiers, computed from observable state in `shared/config/terminalRetention.ts`:

- **foreground:** focused terminals, visible working agents, and visible plain shells. Budgets match the historical fixed caps exactly, so a terminal that never leaves the foreground behaves identically to before (pinned by a cross-module contract test).
- **working:** background working/directing agents. Mirror capped at 5000 lines, which matches the renderer's agent scrollback ceiling, so nothing user-visible is lost across a view-eviction restore.
- **settled:** waiting/idle/completed/exited terminals. Mirror at 2000 lines, matching the renderer's plain-terminal ceiling.
- **archived:** trashed terminals and preserved exit snapshots. Minimal budgets, with preserved snapshots LRU-pruned (20 steady-state, 5 under pressure).

A new `TerminalRetentionCoordinator` sweeps every 5 seconds and stamps tiers from agent state, focus, activity tier, and trash/preserve status. Terminals are born stamped `foreground` so the governor never sees an untiered, unprotected terminal before the first sweep. Re-applying an unchanged tier is a no-op, which means a pressure trim below the tier cap survives until a genuine tier transition instead of being quietly undone.

## State-aware pressure trimming

The `ResourceGovernor`'s pre-pause reclaim is now tier-aware. It prunes preserved snapshots first (pure retained memory, safest thing to drop), then trims the heaviest contributors to their per-tier floors. Foreground terminals are exempt until critical pressure at 95% raw utilization, where an emergency floor-trim runs after the immediate global pause, because at that point an OOM'd pty-host is strictly worse than lost mirror history. I also removed the uniform-flatten fallback when the targeted path throws; degrading to a trim that flattens foreground terminals defeats the whole exemption, and the pause backstop engages a tick later anyway.

## Trim races

Two staleness guards keep trims from corrupting snapshots. Every destructive mirror trim bumps a per-terminal `trimEpoch`, and `SessionSnapshotter` discards any serialize whose await overlapped a trim rather than persisting possibly mid-truncation state (the next flush captures the clean post-trim buffer). In worker mode, `set-scrollback` messages carry the slot generation so a stale trim from a superseded backend can't shrink a rebuilt session's mirror, and a trim that never reached a live worker no longer records phantom bookkeeping.

## Diagnostics

`get-flow-control-snapshot` now reports each terminal's retention tier, per-category retained-byte estimates (mirror, semantic, forensics, agent output, preserved snapshot), and trim history (`trimCount`, `lastTrimAt`, `lastTrimReason`). A support bundle can finally say which terminals and which buffers dominate host memory, not just that memory is high.

## Tests

- Tier selection across all agent states (idle, working, waiting, directing, completed, exited) plus focus, visibility, trash, and preserve combinations.
- Governor ordering: heaviest safe contributors first, per-tier floors, preserved-snapshot pruning, and the foreground exemption holding until critical.
- State detection still working after mirror trims (viewport reads and the waiting-to-busy promotion path).
- `SessionSnapshotter` race guards for both the debounced and event-driven flush paths.
- Renderer profile-change protections: background terminals untouched mid-stream, no mid-stream rewrap, evicting shrinks deferred for scrolled-back and alt-screen terminals.
- A stress suite running 10 real terminal processes under heavy interleaved output with tier churn and pressure trims mid-stream, asserting byte-exact renderer delivery and fully settled scheduler accounting.
- Zero-cap boundary tests for the semantic and forensics buffers (the `slice(-0)` hazard).

I considered a nightly E2E for this but held off; the stress suite exercises the same invariants deterministically, and the nightly memory-leak suite already watches the host footprint.

Builds on #10811 (hibernation removal), #10933 (per-terminal drop tallies), and #10919 (mirror scheduling). Docs updated in `docs/architecture/resource-governance.md`.


gregpriday:
Reverted this from `develop` because in practice it made things slower, not faster.

The goal was to cut pty-host terminal memory with state-aware buffer retention tiers. What actually happened: with a few agent terminals running under memory pressure, the app started stalling. The `ResourceGovernor`'s pre-pause reclaim exempts foreground terminals, but the pty-host runs on a fixed 512MB heap budget (`HEAP_BUDGET_MB`), so the visible active terminals filling that budget were exactly the buffers it refused to reclaim. It freed almost nothing, utilization kept climbing, and it escalated straight to the global PTY pause, which suspends terminal output for up to 10 seconds. That pause is the slowdown people were feeling.

One thing worth flagging: that 512MB budget isn't scaled to system RAM, so a machine with plenty of free memory hits this the same as an 8GB one. Took me a minute to trust the diagnosis on a 64GB box.

I confirmed it the boring way. Reverted, rebuilt, and the stalls went away.

I went with a full revert rather than a patch. The problem is the core design choice of exempting foreground from reclaim, so fixing it in place means reversing that premise while keeping all the machinery built on top of it. The older, simpler reclaim path already kept memory bounded, and it did it by trimming the headless analysis mirror (which nobody sees) instead of pausing output (which everybody sees). Not worth the added complexity for a memory win we didn't really need.

Revert is on `develop` now.

PR #10983, opened 2026-07-07
perf(terminal): resize/reflow benchmarks (PERF-110..112) + xterm buffer-core fastpath

gregpriday:
This PR adds three new performance benchmark scenarios, PERF-110 to PERF-112, covering the terminal resize and reflow pipeline, then uses them to land a significant optimization.

## Benchmarks

Resize was the highest-frequency geometry operation in the app with zero benchmark coverage. `terminal.resize()` parses xterm's entire pending write buffer in one synchronous main-thread task before re-wrapping the whole retained scrollback, so cost scales with backlog bytes x retained lines x terminal count.

* `PERF-110`: single-terminal reflow cost at 10k, 50k and 100k retained rows, on a seeded agent-output mix (~30% wrapped lines).
* `PERF-111`: an eight-terminal splitter-drag resize storm, 30 uncoalesced width steps across the grid. `maxStepMs` is the guarded signal: what one fired resize actually costs. Production debouncing swallows most steps, so `totalBlockedMs` deliberately overstates what a user feels during a drag; the scenario description spells this out.
* `PERF-112`: a regression guard reconstructing the 2026-07-06 incident, where a reveal-time resize on a hidden terminal with a large output backlog locked the renderer for 45 to 60 seconds. The over-budget arm must stay flat with respect to backlog size. Bypassing `RESIZE_FLUSH_SYNC_BUDGET_BYTES` blows the budget immediately (verified manually: 40.8ms against the 30ms ceiling).

## Optimization

`shared/utils/xtermReflowFastpath.ts` is a runtime shim that patches xterm's buffer internals with faster equivalents: bulk typed-array copies, capacity-stable views so a width change no longer allocates a view per line, hoisted ring access in the reflow drivers, and an O(1) per-line overflow check on shrink. It applies at every terminal creation site (renderer, pty-host mirrors, parse authority worker, analysis sessions, perf harness), so every process benefits and the benchmarks measure exactly what production runs.

Results on my machine (ci mode):

| Metric | Before | After | |
|---|---|---|---|
| PERF-110 100k-row reflow (mean) | 22.8ms | 12.8ms | 44% faster |
| PERF-111 worst drag step across the grid | 41.5ms | 22.9ms | 45% faster, now under the 32ms two-frame jank threshold |
| PERF-112 reveal blocking on a 4MiB backlog | 1.72ms | 0.75ms | 56% faster |

## Safety

Every fast path falls back to the original method when it encounters combined characters, extended attributes, protected cells, or any shape drift in a future xterm version. An equivalence test suite replays adversarial content and resize sequences (CJK, emoji, ZWJ, combining accents, curly underlines, DECSCA-protected cells, alt buffer round trips, markers) against patched and unpatched terminals and requires identical buffer snapshots, so an xterm bump that drifts the mirrored internals fails tests before it ships. There's also a kill switch, `DAINTREE_DISABLE_XTERM_REFLOW_FASTPATH=1`, which doubles as the A/B lever for the benchmarks.

A Codex review pass hardened the `copyCellsFrom` guard (overlap direction and logical bounds), tied the PERF-112 fixture to the production flush decision via a shared `exceedsResizeFlushSyncBudget()` predicate, and pinned the seeded content checksum as a golden value.

## Next steps

Run the `full-terminal` E2E bucket before merge. Baseline p95 entries are hand-seeded from this machine and will tighten via the daily perf baseline workflow.


gregpriday:
Rebased onto develop to clear the conflict with #10975. The only collision was the scenario-matrix count in `scenarioMatrix.test.ts` (PERF-035 landed there first), which is now 44 with both blocks registered. Reran typecheck, the equivalence suite and the matrix tests on the rebased branch, plus a quick PERF-110/112 sanity run: numbers hold (100k reflow ~12.9ms, backlog checksum unchanged).

PR #10986, opened 2026-07-07
feat(panels): allow file viewer panels to dock

### LINKED ISSUE #10985: Support file viewer panels in the dock
We now have a file viewer panel (#10982), but we can't move it to the dock. The move-to-dock button is hidden for these panels because the dock only supports terminal panels: `ContentDock` and `DockPanelOffscreenContainer` narrow their membership to `PtyPanelData` via `isPtyPanel`, and the `hasPty` gate in `PanelHeader` exists precisely so a non-terminal panel can't be moved somewhere it would sit stranded and invisible.

The use case is strong though. You're directing agents in the grid, but you have a spec or notes file you want to keep close by. Tuck it in the dock, pull it up when you need it, and it stays out of the way the rest of the time.

This means widening the dock's membership model beyond terminal panels: the dock chip list, tab groups, the offscreen container, and persistence all assume `PtyPanelData` today. Reading surfaces like the file panel are the first candidates; the same widening would presumably unlock review and browser panels later, but file panels are the ask here.

gregpriday:
This adds dock support for the new file viewer panels. Closes #10985. Until now, the dock only supported terminal panels.

This adds a `dockable` capability flag to the panel kind registry, and widens dock membership with a new `DockPanelData` type. It also adds a new dock chip component for file panels (`DockedFilePanelItem`), which reuses the existing offscreen wrapper relocation so the panel stays mounted as the dock popover is opened and closed: scroll position and view mode survive.

The move to dock button, the keyboard action (`terminal.moveToDock`), the dock toggle action, bulk move, and the context menu item all check the new capability before offering to move a panel.

In the process, review caught two pre-existing bugs that are fixed here:

* `terminal.toggleDock` and `bulkMoveToDock` had no kind gate at all, so they could strand a browser, dev preview, or review panel invisibly in a dock that never renders it. The shared context menu offered the same trap.
* `getNarrowPanel` silently dropped file panels: it enumerates the built-in kinds explicitly and never learned about the `file` kind, so any consumer of the narrowed read path lost them.

Tab groups in the dock stay terminal only. If you move a tab group that contains a file panel, the file panel renders as a standalone chip. Browser, dev preview, and review panels remain non-dockable for now; the same widening would unlock them later.

## Validation

* `npm run check` passes
* ~8,400 unit tests pass, including new coverage for the capability, dock render items, and the dock/grid round trip
* 8 end to end tests pass, including a new test that moves a file panel to the dock, previews it from the chip, and restores it to the grid

Stacked on #10982 (`feat/file-viewer-panel`).


gregpriday:
Superseded by #10988 — GitHub closed this one when its base branch was deleted after #10982 merged, and closed PRs can't be retargeted. Same commits, rebased onto develop.

PR #11062, opened 2026-07-11
feat(forge): add GitLab support alongside GitHub

### LINKED ISSUE #5167: feat(forge): add GitLab support alongside GitHub
## Problem

Canopy's repository integration is hardwired to GitHub.com — hardcoded GraphQL queries via Octokit, GitHub-specific URL parsing, and no abstraction layer for other git hosts. Users whose projects live on GitLab (cloud or self-hosted) cannot access any of the PR/issue workflow features: token auth, MR-to-worktree linking, issue panels, or bulk worktree creation from MRs.

## User Story

As a developer whose codebase lives on GitLab, I want Canopy to detect my GitLab remote and surface open MRs linked to my worktrees, so that I get the same workflow acceleration that GitHub users have today.

## Proposed Approach

Introduce a thin `IForgeService` interface with two concrete implementations (`GitHubForgeService`, `GitLabForgeService`), selected at project load time via remote URL detection. GitLab uses REST v4 (PAT auth); GitHub keeps its existing GraphQL/Octokit path unchanged.

## Acceptance Criteria

- [ ] Given a project whose remote URL is gitlab.com or a self-hosted GitLab instance, the forge is detected and a `ForgeService` backed by the GitLab REST v4 API is instantiated
- [ ] Given a valid GitLab PAT with `api` scope is saved in settings, token validation succeeds and stores user info (username, avatar)
- [ ] Given a worktree on a branch matching an open MR's source branch, the MR number, title, and state badge appear in the worktree panel — identical in structure to the GitHub PR display
- [ ] Given a user opens the issues/MR panel for a GitLab project, a paginated list of open issues and MRs is fetched and displayed with search/filter parity to the GitHub panel
- [ ] Given a user selects MRs from the list, the bulk "Create Worktrees" action creates a worktree per selected MR branch
- [ ] Given a project with a github.com remote, all existing GitHub behaviour is unchanged — no regression in auth, PR detection, tooltips, or actions

## Decomposition

1. Define `IForgeService` interface in `electron/services/forge/` covering: auth, listPRs/MRs, listIssues, getTooltip, openURL, parseRemoteUrl — confirm coverage of all 19 current IPC channels
2. Refactor `GitHubService` to implement `IForgeService`; add `detectForge(remoteUrl)` factory that returns the right implementation
3. Implement `GitLabForgeService` using GitLab REST v4: PAT auth, MR list, issue list, repo stats, branch-to-MR matching, self-hosted base URL support
4. Generalize IPC handler (`github.ts` → `forge.ts`) to delegate to the active `IForgeService` — update all four IPC layers: `channels.ts`, `handlers/`, `preload.cts`, `src/types/electron.d.ts`
5. Add GitLab token input to settings UI; update or replace `githubConfigStore` with a forge-aware equivalent
6. Rename/generalize UI components (`GitHubResourceList`, `GitHubListItem`) to be forge-agnostic in place; update action definitions in `githubActions.ts`
7. Write unit tests for `detectForge()`, `GitLabForgeService` (mock REST responses), and branch-to-MR matching; regression test confirming GitHub path is unaffected

## Out of Scope

- GitHub Enterprise Server
- Bitbucket, Azure DevOps, Gitea
- OAuth / browser-based auth flows (PAT only for MVP)
- GitLab CI / pipeline status integration
- Issue or MR creation from within Canopy
- Self-hosted GitLab certificate pinning or SSO

## Notes

- The `IForgeService` interface definition is the critical design gate — get that agreed on before the implementation fan-out
- GitLab REST v4 covers every query currently made via GitHub GraphQL; no capability gaps for MVP scope
- Branch naming: `feat/forge-gitlab-support`

gregpriday:
Resolves #5167

This adds GitLab as a built-in forge provider plugin at `plugins/builtin/gitlab`, sitting alongside the GitHub plugin. It's built on the forge provider abstraction that shipped after this issue was filed, and the abstraction held up well: the host needed zero changes for the provider itself. The registry, IPC surface, credential storage, and the Code Forge preferences UI all picked the new provider up automatically.

## What's implemented

The plugin implements the full `ForgeProviderImpl` contract over the GitLab REST v4 API:

- Issues and merge requests: paginated lists with search/label/assignee filters, single fetches, and the full mutation set (create, close/reopen, edit, comment, labels, assign/unassign, merge, draft toggling).
- Branch to MR detection, so worktree cards get their MR badges. Batched lookups go through GitLab's GraphQL `mergeRequests(sourceBranches:)` in chunks of 20 branches, with the REST `source_branch` filter as the per-branch fallback.
- CI status from the MR's head pipeline, mapped into the normalized vocabulary.
- Repo metadata, releases, hover tooltips, viewer identity, commit-author avatars via `/avatar`, toolbar repo stats (open issue/MR counts from `x-total` headers), token health, and rate limit reporting.

MRs are numbered by `iid` throughout, and nested subgroups are handled (the whole `group/subgroup/project` path URL-encodes as the project id).

## Auth and self-hosted

Auth is a personal access token (`api` scope, or `read_api` for read-only), validated against `/user` with scope and expiry introspection via `/personal_access_tokens/self`. Storage rides the host's existing forge credential store.

Self-hosted support works through an instance URL setting on the plugin's settings tab. Custom schemes, ports, and path prefixes are preserved on API calls, and the token is strictly scoped to the configured instance: it's never attached to a request against any other host, and it fails closed if the settings read breaks. gitlab.com plus a handful of well-known public instances (`invent.kde.org`, `salsa.debian.org`, `gitlab.freedesktop.org`, `gitlab.gnome.org`, `gitlab.haskell.org`) match by hostname; anything else routes via the existing per-project provider override. Switching the instance URL clears the stored token rather than silently re-scoping it.

## Host integration fixes

A second review pass traced the real host call paths end to end (activation, credential replay, the workspace-host RPC bridge, the toolbar, settings IPC) and fixed four seams a second provider exposed:

- Credential save and token test now implicitly activate a lazy provider before the impl lookup. Previously, connecting GitLab on a fresh session failed with "Provider not activated" while you were standing in the Settings screen that was supposed to activate it.
- The toolbar issue/MR pills route clicks to the forge's own web pages when the provider contributes no stats-dropdown view, instead of toggling an empty popover. Hover prefetch is skipped in that case too.
- A provider without the `projectHealth` capability now reports `unsupported`, and the pulse card stays quiet instead of telling you to "Connect a git remote" you already have.
- The New Worktree dialog hides the "Link Issue" section when the provider contributes no issue-selector view, rather than rendering an orphaned label.

## Review

Both Codex review passes are folded in. The first caught a token scoping leak where a deactivation race could send a self-hosted token to the gitlab.com fallback, and a GraphQL truncation bug where branches could be reported as having no MR when the contract requires omitting them so the host falls back per-branch. The second pass audited every request against the real GitLab API (REST params, GraphQL field and enum names, Bearer PAT auth, pagination headers, `%2F` project-id encoding all check out) and produced the host integration fixes above.

## Deliberate follow-ups

- Stats dropdown, bulk create worktrees, and issue selector views. Counts and web click-through work today; the in-app list views are the next chunk of renderer work.
- Merge request terminology in the host UI. "PR" is hardcoded across host-owned components and needs a terminology field on the provider contract.
- Multi-account support (one token per provider today, so gitlab.com plus a self-hosted instance means picking one).
- Authenticated cloning. I dropped this on purpose: embedding the token in the clone URL persists it into `.git/config` via `remote.origin.url`. It needs an askpass-style flow like the GitHub plugin's `gh repo clone` path.
- Two pre-existing host gaps the review surfaced that affect GitHub too: provider resolution reads the `origin` remote only (a configured `forgeRemote` is honored by workspace polling but not by stats/tooltips/actions), and the renderer treats public repos as token-required for tooltips and badge clicks even though unauthenticated reads work.

## Testing

80 plugin unit tests cover remote parsing, mappers, the provider contract, and activation, plus updated host handler and toolbar tests for the integration fixes. `npm run check` is green. This is a draft because it still needs real-API testing against gitlab.com and a self-hosted instance before it's ready to merge.

### DIFF
--- a/electron/ipc/handlers/__tests__/forge.rateLimit.test.ts
+++ b/electron/ipc/handlers/__tests__/forge.rateLimit.test.ts
@@ -93,6 +93,7 @@ vi.mock("../../../utils/openExternal.js", () => ({
 vi.mock("../forgeResolution.js", () => ({
   resolveForCwd: resolveForCwdMock,
   getImplForNamespace: () => fakeImpl,
+  getImplForNamespaceActivating: () => Promise.resolve(fakeImpl),
 }));
 
 vi.mock("../../../store.js", () => ({
--- a/electron/ipc/handlers/__tests__/forge.rateLimit.test.ts
+++ b/electron/ipc/handlers/__tests__/forge.rateLimit.test.ts
@@ -96,6 +96,7 @@ vi.mock("../../../utils/openExternal.js", () => ({
 vi.mock("../forgeResolution.js", () => ({
   resolveForCwd: resolveForCwdMock,
   getImplForNamespace: () => fakeImpl,
+  getImplForNamespaceActivating: () => Promise.resolve(fakeImpl),
 }));
 
 vi.mock("../../../store.js", () => ({
--- a/electron/ipc/handlers/__tests__/forgeSettings.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/forgeSettings.handlers.test.ts
@@ -63,10 +63,12 @@ vi.mock("../../../services/GitServiceCache.js", () => ({
 }));
 
 // The handlers lazy-import PluginService to gate registry reads behind
-// startup load + activation (the #9285 init-race guard); stub the singleton
-// so tests never construct the real service.
+// startup load + activation (the #9285 init-race guard) and to implicitly
+// activate a lazy provider before the credential impl lookup; stub the
+// singleton so tests never construct the real service.
 const pluginServiceMock = vi.hoisted(() => ({
   waitForInit: vi.fn(() => Promise.resolve()),
+  activatePluginForForgeProvider: vi.fn(() => Promise.resolve()),
 }));
 
 vi.mock("../../../services/PluginService.js", () => ({
@@ -541,7 +543,32 @@ describe("registerForgeSettingsHandlers", () => {
     expect(workspaceClientMock.updateForgeCredentials).not.toHaveBeenCalled();
   });
 
-  it("setCredential returns a not-activated error when no impl is registered", async () => {
+  it("setCredential activates a lazy provider before the impl lookup", async () => {
+    registerGiteaProvider();
+    const validateToken = vi.fn().mockResolvedValue({ valid: true });
+    // Impl absent on first lookup; activation binds it for the re-read —
+    // modeling a fresh session where Settings is the provider's first touch.
+    registryMock.getForgeProviderImpl.mockReturnValueOnce(undefined);
+    // `mockImplementationOnce`: `clearAllMocks` in beforeEach does not reset
+    // implementations, so a persistent one would leak the re-bound impl into
+    // later tests.
+    pluginServiceMock.activatePluginForForgeProvider.mockImplementationOnce(() => {
+      registryMock.getForgeProviderImpl.mockReturnValue({ validateToken });
+      return Promise.resolve();
+    });
+    registerForgeSettingsHandlers();
+    const setCredential = findHandler("forge:set-credential");
+
+    const result = (await setCredential(null, "acme.gitea", { token: "x" })) as {
+      valid: boolean;
+    };
+
+    expect(pluginServiceMock.activatePluginForForgeProvider).toHaveBeenCalledWith("acme.gitea");
+    expect(result.valid).toBe(true);
+    expect(validateToken).toHaveBeenCalledWith("x");
+  });
+
+  it("setCredential returns an unavailable error when activation binds no impl", async () => {
     registerGiteaProvider();
     registryMock.getForgeProviderImpl.mockReturnValue(undefined);
     registerForgeSettingsHandlers();
@@ -552,8 +579,9 @@ describe("registerForgeSettingsHandlers", () => {
       error?: string;
     };
 
+    expect(pluginServiceMock.activatePluginForForgeProvider).toHaveBeenCalledWith("acme.gitea");
     expect(result.valid).toBe(false);
-    expect(result.error).toMatch(/not activated/i);
+    expect(result.error).toMatch(/isn't available/i);
     expect(storeMock.set).not.toHaveBeenCalledWith("forgeCredentials", expect.anything());
   });
 
--- a/electron/ipc/handlers/__tests__/forgeSettings.handlers.test.ts
+++ b/electron/ipc/handlers/__tests__/forgeSettings.handlers.test.ts
@@ -74,10 +74,12 @@ vi.mock("../../../services/GitServiceCache.js", () => ({
 }));
 
 // The handlers lazy-import PluginService to gate registry reads behind
-// startup load + activation (the #9285 init-race guard); stub the singleton
-// so tests never construct the real service.
+// startup load + activation (the #9285 init-race guard) and to implicitly
+// activate a lazy provider before the credential impl lookup; stub the
+// singleton so tests never construct the real service.
 const pluginServiceMock = vi.hoisted(() => ({
   waitForInit: vi.fn(() => Promise.resolve()),
+  activatePluginForForgeProvider: vi.fn(() => Promise.resolve()),
 }));
 
 vi.mock("../../../services/PluginService.js", () => ({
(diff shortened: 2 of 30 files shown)


gregpriday:
Rebased this onto `develop` and worked it over properly. It was 3249 commits behind, so a lot had moved underneath it.

## Rebase

Two files conflicted, `ProjectPulseCard.tsx` and `IssueSelectorView.tsx`, and both had been reworked on `develop` in the meantime. The pulse card now builds its health section into a `healthSection` variable, so I reapplied the `unsupported` handling on top of that structure. `IssueSelectorView` already had the same `if (!IssueSelector) return null;` guard I was adding, so I took develop's side.

Eight type errors came out the other side, all from `ForgeProviderImpl` changing while this branch sat: eight mutations now return the mutated entity instead of `Promise<void>`, and cache invalidation moved from the host to the provider.

## Review

Three rounds of Codex review on top of that, plus two rounds of research against GitLab's docs. The substantial things:

**Credentials are now bound to the instance a validation actually reached**, recorded as an instance URL plus a sha256 digest in plugin storage. The host's credential store holds a bare token with no instance attached, so on every activation the provider was handed a token it had no way to place. Repointing `instanceUrl` (which the generic plugin settings form does without clearing credentials) could then hand instance A's token to instance B. Three weaker versions of this leaked in review before the current one held: adopting the current setting, lazy adoption on an unknown binding, and a single slot that a background probe could clobber mid save.

**Destination and authorization now come from one settings read.** They were two, so a failed second read could send the token to a fallback origin that dropped the configured port and deployment path.

**Three places were wrong against current GitLab**, and I only found these because I checked the docs rather than trusting the code:

* The `WIP:` prefix was removed in GitLab 14.8. Treating it as a draft made "convert to draft" a silent no-op on merge requests GitLab considers open.
* The assignee filter used a parameter shape Grape silently drops, which returns an unfiltered list rather than an error. It sends `assignee_username[]` now.
* `merge_commit_sha` is preferred over `squash_commit_sha`. Squashing into a merge commit project populates both, and the merge commit is the one that lands on the target branch.

Also fixed: a 200 carrying an object became an empty page ("this project has no issues"), an HTML sign in page certified the token as healthy, and a merge acknowledgement with no state reported a merge. Self hosted deployment prefixes were being folded into the project namespace, and my first fix for that broke SSH remotes.

Plugin suite is at 131 tests, up from 80. I mutation tested every new guard, so each one fails when its fix is reverted.

## Tested against a real GitLab

Everything above was mocked `fetch` and documentation, which is not the same as knowing it works. So I stood up GitLab CE 19.3.1 in Docker (Colima with Rosetta, since the official image is still amd64 only) and wrote three harnesses that import the real provider modules rather than reimplementations. 77 checks, all passing.

What the live run confirmed that mocks could not:

* The assignee filter genuinely filters. This is the one a mock can never catch, because a wrong shape returns unfiltered results instead of an error.
* Draft toggling round trips, and the server agrees afterwards with no `Draft:` prefix stranded in the title.
* `mergePR` returns a commit sha that actually exists in the repo. I fetched the commit back to check.
* Pagination cursors advance with zero overlap between pages.
* Nested subgroups work three levels deep (`g/team/squad/app`).
* Credential isolation holds live: repointing `instanceUrl` withholds the token, provenance still names the proven instance, and it authenticates again at its own instance.

Two findings worth knowing:

1. **`RateLimit-*` headers are absent on a self-managed instance by default.** The snapshot comes back all null. We handle that correctly (no limit recorded, no false gate), but it means the rate limit surface is really only meaningful on gitlab.com.
2. **GitLab 19 moved the canonical issue `web_url` to `/-/work_items/N`** as part of the work items migration. This is fine for us because the mappers read the server's own `web_url`, so links follow automatically. I checked the Rails route table and `/-/issues/N` is still live (62 routes), so `buildIssueUrl` and the comment anchors are still valid. If that route ever goes, `buildIssueUrl` is the one place to change.

One correction on myself: three of the first run failures were bugs in my harness, not the product. `AuthValidation` does not carry `username` or `credentialRejected`, and `findPRsByBranches` returns a `Map` which I was indexing like an object, so `JSON.stringify` printed `{}` and it looked like a real failure. Fixed the assertions and reran clean rather than leave a passing but broken harness.

## Known limitation

`WorkspaceService.fetchPRBranch` fetches `pull/<n>/head`, which is GitHub's refspec. GitLab serves `merge-requests/<n>/head`, so fork merge requests and un-fetched branches cannot be checked out. Filed as #12324 rather than grown into this PR, since it means adding to the `ForgeProviderImpl` contract and threading it through the workspace host IPC. The dialog degrades gracefully (warning banner, submit disabled, `git fetch origin` workaround); the agent path throws a raw git error, which is the half worth fixing.

## Verification

Every step from `ci.yml` run locally: `npm run check`, the full vitest suite (57,577 passing), `npm run build`, `check:preload-backdoors`, and the launch smoke test. Lint ratchet sits 3 below baseline. I left the baseline alone, since lowering it on a branch collides with parallel PRs.

`check:sample-views` fails as a known false positive in a worktree with symlinked `node_modules` (comment only path diff), so I ran the four guards it skips separately.

PR #11206, opened 2026-07-16
feat(agents): default grok and opencode to inline mode

gregpriday:
Makes inline rendering the default for the `grok` and `opencode` agent CLIs. Their output now flows into our own WebGL scrollback and stays selectable with our native text-selection layer, instead of getting locked inside the agent's full-screen alternate-screen TUI.

The global default was already inline for any agent with an inline flag, so this was a small change.

## Grok

Grok had an explicit opt-out, `defaultInlineMode: false`. I removed it, so grok now follows the global switch like everything else. Its `--no-alt-screen` flag was already wired.

## OpenCode

OpenCode is a Bubble Tea TUI, so the renderer-side `blockAltScreen` path would garble it (same constraint as Crush). Instead I wired OpenCode's own `--mini` flag as its inline flag. I verified with a PTY probe that `opencode --mini` emits no alternate-screen enter (`\x1b[?1049h`) and no mouse-reporting, while plain `opencode` emits both.

OpenCode's headless `run` subcommand rejects `--mini` (`--mini must be used without the run subcommand`), so I gated the inline-flag injection in `generateAgentCommand` on interactive mode. `buildAgentLaunchFlags`, the persistent interactive-spawn path, still injects it unconditionally since that path is always interactive.

Both agents stay overridable globally and per-agent.

## Tests

Updated the affected tests. Grok now defaults inline, and the curated-default precedence tier is exercised via a synthetic registered agent since no shipped agent pins `defaultInlineMode` anymore. 143 pass, typecheck and lint clean.

## Follow-up

OpenCode's working-state detection patterns (the braille spinner and `esc to interrupt`) should be validated live under `--mini`. The idle, prompt and boot markers (`Ask anything`, the prompt glyph, `OpenCode`) are already confirmed present in `--mini` output.

### DIFF
--- a/shared/config/agents/grok.ts
+++ b/shared/config/agents/grok.ts
@@ -73,13 +73,14 @@ export const config: AgentConfig = {
   capabilities: {
     scrollback: 10000,
     resizeStrategy: "default",
-    // Rust TUI that takes over the alternate screen. We keep it ON the alt
-    // screen by default: forcing it inline (--no-alt-screen) renders every frame
-    // into the normal-buffer scrollback, so scrolling up shows garbled overdrawn
-    // history and the overlay scrollbar can't be auto-hidden. The flag stays
-    // available as an opt-in via the inline-mode toggle.
+    // Rust TUI that would otherwise take over the alternate screen. We default
+    // it inline (`--no-alt-screen`, grok's scrollback-native mode) so output
+    // flows into Daintree's own WebGL scrollback and stays selectable with our
+    // native text-selection layer instead of grok's mouse-grabbing full-screen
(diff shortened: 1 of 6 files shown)


gregpriday:
The main reason for this is to unlock our own WebGL scrollback, which is significantly faster than the scrolling the agents implement themselves. It also lets us own advanced mouse behaviour.

The longer-term goal is a full selection system: select content in one terminal, then send it to another. To make any of that work we need to take control of scrollback, selection and mouse handling ourselves. Alternate-screen rendering hands all of that to the agent and gives us none of it.

PR #11582, opened 2026-07-31
feat(mcp): ship a small core tool set and defer the rest

### LINKED ISSUE #11540: Ship a small core tool set and defer the rest
The progressive disclosure machinery is already built and almost unused.

`shouldExposeTool` drops any entry marked `hidden` or `discoverable` from `tools/list`. `isTierPermitted`, which gates `tools/call`, never looks at `mcpVisibility` at all. So a `discoverable` action stays fully callable, findable through `actions.search` and inspectable through `actions.getSchema`, at zero eager cost.

Of 480 registered actions, 5 are marked `core`, 10 `discoverable`, 1 `hidden`, and the rest are unclassified. The external tier advertises 96 allowlisted ids and emits 87 of them eagerly. The system tier emits 144, which measures at roughly 128 KB.

Marking most of the surface `discoverable` and keeping a core of 15 to 25 would cut the eager envelope by around 80 percent while leaving every tool reachable.

Two things should shape which ids stay in the core. Prefer what only Daintree can do, terminal and agent orchestration, worktrees, recipes, live IDE context, over what a client already has its own tools for. And keep the entry points an agent needs before it knows what to search for.

Worth being explicit that this is a listing change, not an access change. The tier allowlist stays exactly as it is.

Two things to check before landing: the readiness probe asserts `actions.list` appears in `tools/list`, and the tier snapshot tests assert surface membership.

Depends on #11525. Search has to return only what the caller can dispatch before search becomes the primary way to find a tool.

Evidence: `electron/services/mcp-server/tierAuth.ts`, `shared/types/actions.ts`, `shared/config/helpAssistantTierAllowlists.ts`, `electron/services/mcp-server/shared.ts`.

gregpriday:
## What

`tools/list` used to advertise everything a tier permitted. Since almost no action set `mcpVisibility`, that meant 144 tools at the system tier (~128 KB) and 87 externally — most of which a client already has better tools for.

Eager listing is now an explicit opt-in. `shouldExposeTool` advertises an entry only when it declares `mcpVisibility: "core"`; unset, `discoverable`, and `hidden` are all deferred. Deferred tools stay fully dispatchable and reachable through `actions.search` / `actions.getSchema` — this is a listing change, not an access change, and `TIER_ALLOWLISTS` is untouched.

Flipping the default rather than annotating ~150 ids `discoverable` is the point: with eager-by-default, every action later added to a tier allowlist silently re-bloats the surface. Now the core is an explicit, greppable opt-in and the rest is deferred by construction.

**Result:** system tier 144 → 22 eager tools, external 87 → 20. Workbench 11, action 22.

## The core set (22)

Discovery entry points — `actions.list`, `actions.getContext`, `actions.search`, `actions.getSchema`, plus `help.displayImage`.

Daintree-unique orchestration — `terminal.list`, `terminal.getStatus`, `terminal.getOutput`, `terminal.sendCommand`, `terminal.new`, `terminal.inject`, `terminal.close`, `terminal.waitUntilIdle`, `terminal.waitUntilIdleBatch`, `agent.launch`, `worktree.list`, `worktree.getCurrent`, `worktree.setActive`, `worktree.createWithRecipe`, `recipe.list`, `recipe.run`, `fleet.getRunStatus`.

Deliberately deferred: forge, git plumbing, and file tools (a client has its own), `project.*` reads (`actions.getContext` already returns the active project), and `agent.getState` (`terminal.getStatus` is the canonical terminal-id-keyed path). `terminal.close` is core while `kill`/`closeAll`/`restart` are not — close is the recoverable trash path, the others are confirmation-gated and lose process state.

## Guarding the surface

`actionDefinitions.quality.test.ts` derives the core cohort from the real registry — never restating the 22 ids — and asserts a 15–25 count budget, a ceiling on total core description bytes (count alone can't see 22 tools with tripled descriptions), that every core id is dispatchable at some tier, that none is `restricted`, and that the discovery entry points stay core. Without those last three, a caller can't search for the search tool.

## Assistant prompts

The bundled prompts told the model `ListTools` was the exact surface and that its tier could be read off it. Both are now false, so they're corrected — including one real trap: out-of-tier tools are dropped from `actions.search` and `getSchema` returns `NOT_FOUND`, so an assistant could never provoke the `TIER_NOT_PERMITTED` the old wording told it to wait for. Search-absence is now the documented signal.

## Worth a second opinion before merge

There's no generic dispatcher tool — `actions.*` is list/getContext/persistedStores/search/getSchema, all read-only. So while the server happily accepts `tools/call` for an unlisted name, a *model* can generally only emit calls for tools its client registered from `tools/list`. That's equally true of the 10 actions already shipping as `discoverable`, and `docs/architecture/mcp-server.md` states those are "deliberately callable once an agent finds it via the meta-tools" — so this PR follows the existing design rather than changing it. Still worth a real check with Claude and Codex help sessions: search for a deferred tool, then call it.

Separately, and pre-existing: `setSessionTier` (`httpLifecycle.ts:1522`) sends no `tools/list_changed` on elevation — only decay notifies. The visible core differs by tier (11 → 22), so a client caching `tools/list` won't learn about newly permitted core tools after an elevation. Not touched here.

## Testing

1,696 tests green across every module this branch touches — `tierAuth`, `sessionServer`, `readinessProbe`, `rendererBridge`, all five `McpServerService` suites, the action-definition suites, and the quality gate. `check:help` verifies the regenerated prompts match their sources. eslint clean, prettier clean.

Resolves #11540


gregpriday:
Not merging this. It rests on an assumption that does not hold: that tools dropped from `tools/list` stay callable through `actions.search` and `actions.getSchema`.

They do not. There is no generic dispatch tool. `actions.*` is five read-only introspection tools (`shared/config/actionIds.ts:341-345`) and none of them executes an arbitrary action id. MCP clients only invoke names they received from `tools/list`, so the deferred remainder becomes unreachable rather than lazily loaded.

The server half of the path genuinely works, which is what makes this easy to miss. `tools/call` never reads `mcpVisibility`, it only gates on `isTierPermitted` (`sessionServer.ts:494`), so it would dispatch a deferred action quite happily. But nothing ever sends that request. Claude Code builds its registry from `tools/list` and fails an unlisted name client side before it becomes a `tools/call`, and its tool search defers schemas, never existence. I checked the other escape hatches too and none of them rescue it: we return no `nextCursor` (`sessionServer.ts:416`), re-listing after `notifications/tools/list_changed` just reapplies the same core-only filter, and there is no standardised per-tool dynamic registration.

The added tests do not catch this because neither one closes the loop. `sessionServer.test.ts` asserts deferred ids stay tier-permitted, and `introspectionActions.test.ts` stops after retrieving a schema without calling the action.

Same gap already applies to the 10 actions marked `discoverable` today, so this is not a new bug so much as scaling an existing one from 10 actions to the whole surface. Nothing in `help/` even mentions the meta-tools. The instructions we ship say the opposite: "Discover the exact tool surface at runtime via `ListTools`".

Two things this PR gets right and I want to carry forward. `tierAuth.ts` is not a security regression, it is strictly tighter: the `!== "core"` early return subsumes the old hidden and discoverable rejects, the `restricted` floor is untouched, and `tools/call` still ends up at `isTierPermitted`. And the underlying goal is real and more urgent than the issue framed it. 87 tools on the external tier is already past what Cursor will accept (it caps at 40 and silently truncates).

Follow-up in #11585 with what the actual fix looks like.

gregpriday:
Closing in favour of #11585, which takes the same goal (a smaller eager tool surface) but as a real allowlist cut rather than a listing filter. Reasoning in the review comment above.

Keeping the branch rather than deleting it. The `tierAuth.ts` tightening here is good on its own and the core-set selection is the input to the allowlist cut.

PR #11663, opened 2026-08-04
fix(perf): repair the always-failing Performance workflow

gregpriday:
The scheduled Performance workflow has failed every run since 2026-05-14 — 60 consecutive failures. It gates nothing (`ci-ok` is the sole required check), which is why it rotted unnoticed. Three independent breakages, plus a fourth found while verifying.

## What was broken

**1. The perf store bundle stopped linking.** `worktreeSidebarFixture.ts` esbuild-bundles the real `createWorktreeStore` with four renderer-only leaves stubbed. `40352d377` (2026-07-23) added two more imports to the store, and esbuild links every static import regardless of reachability:

```
No matching export in "daintree-stub:@/components/Worktree/worktreeDeleteHelper" for import "captureWorktreeTerminalSnapshot"
```

That killed the smoke, ci, and nightly matrices outright.

**2. `perf-update-baselines` could not produce a nightly baseline.** The job ran `npm ci` and then the four regen commands — no build, no `electron-builder`, no xvfb. PERF-004 launches a real packaged binary, so the nightly regen threw `packaged binary not found under release/` and aborted before writing anything. `baseline.nightly.json` could therefore never gain the 9 scenarios the `perf-nightly` gate requires. The job that exists to heal a stale baseline was structurally incapable of producing one. All four regen steps are `continue-on-error: true`, so they showed green ticks while exiting 1.

**3. The PR step is blocked by org policy.** `peter-evans/create-pull-request` fails with `GitHub Actions is not permitted to create or approve pull requests`. This is a **daintreehq org** setting, not a repo one — the repo-level API returns HTTP 409. See "What you still need to do".

**4. (found while verifying) `run.ts` zeroes 24 gated scenarios.** `fccf058ee` (2026-04-19) changed the timing to `sample.durationMs >= 0 ? sample.durationMs : wallClockMs` for PERF-004's NDJSON path. But ~32 legacy scenarios hardcode `durationMs: 0` from when the harness wall-clocked everything unconditionally, and `0 >= 0`. Regenerating writes a literal-zero p95 for them, which sinks each one under `gate.ts`'s `MIN_REGRESSION_BASELINE_MS` floor and silently disables its regression gate. This was invisible because no baseline has regenerated since 2026-02-11 — before the change. Had the other fixes landed alone, the first healing PR would have quietly gutted the gate for 24 scenarios.

## What this changes

- Stub fix, plus `panelIds` on the panelStore stub (`cleanupOrphanedTerminals` reads it)
- `perf-update-baselines` gains build → package → verify → runtime-deps → xvfb, mirroring `perf-nightly`; only the nightly regen needs xvfb, since PERF-004 is nightly-only
- `run.ts` now treats only a **positive** `durationMs` as self-timed
- New `npm run perf verify-baselines`, run between the regens and the PR step: asserts all four baselines are fresh, mode-correct, complete against their budgets, finite, and non-zero. Without it the job stages the modes that succeeded next to the modes left stale — not theoretical, that's how `perf/update-baselines-automated` ended up carrying a soak-only commit
- `scripts/perf/__tests__/worktreeStoreBundle.test.ts` links the bundle from normal PR CI, so the next missing stub export fails where someone will see it
- `workflow_dispatch` allowed on develop; optional `PERF_BASELINE_TOKEN`

## Verification

Run locally against a real packaged binary:

| Step | Result |
|---|---|
| `npm run build` + `electron-builder --dir` + verify binary | exit 0 |
| `perf {smoke,ci,nightly,soak} --update-baseline` | all exit 0 |
| `perf verify-baselines` | all 4 verified |
| `perf soak` | exit 0 |
| perf unit tests (12 files, 128 tests) | exit 0 |
| typecheck / prettier | exit 0 |

The regenerated nightly baseline carries all 63 scenarios including the 9 that were missing, with **zero** literal-zero p95 values (54 and 24 respectively before these fixes). The bundle test was confirmed to fail against the original stub, reproducing the exact CI error, and to pass once fixed.

No baselines are committed here: these are Apple Silicon numbers and the gate runs on `ubuntu-22.04`. They have to be regenerated on the runner, which is what `perf-update-baselines` is for.

The `perf ci` and `perf nightly` **budget** gates flake on my machine — a different scenario trips each run (PERF-031/033/034, then PERF-130, then PERF-137, then PERF-042/052/080/100) at load average 11 with several Electron renderers at ~38% CPU. That is measurement noise, not drift. Worth knowing: the sentinel fix trades "silently disabled gate" for "gate that can flake on noisy hardware" for those ~32 scenarios, and their budgets were written against Feb-era numbers, so they may want a tune once real Linux-runner baselines land.

## What you still need to do

`perf-nightly` cannot go green until a complete baseline is committed, and that needs the PR step unblocked. Either:

- flip **org** Settings → Actions → "Allow GitHub Actions to create and approve pull requests", or
- add a `PERF_BASELINE_TOKEN` repo secret (fine-grained PAT, contents + pull_requests write) — no org change needed

Then dispatch Performance on develop and merge the baseline PR it opens.


gregpriday:
Dispatched this branch on real runners to check the budget gates on proper hardware rather than my laptop: [run 30907857061](https://github.com/daintreehq/daintree/actions/runs/30907857061).

**`perf-ci`: PASS — `scenarios=56 failed=0`.** First green since 2026-05-14, with zero budget failures. That settles the flakiness question from the description: it was load on my machine (average 11, several Electron renderers at ~38% CPU), not the gates themselves. No budget tuning looks necessary.

**`perf-nightly`: FAIL**, exactly as predicted and for the one remaining reason:

```
[perf:nightly] FAIL 9 budgeted scenario(s) missing from baseline — regression gate cannot run:
PERF-004, PERF-063, PERF-070, PERF-071, PERF-072, PERF-073, PERF-080, PERF-090, PERF-091
```

Worth noting what did pass in that job: **Build → Package Linux binary (dir target) → Verify packaged binary exists** all succeeded on Ubuntu. Those are the steps this PR adds to `perf-update-baselines`, so the mechanism that was missing is confirmed working on the runner.

`perf-update-baselines` correctly skipped (gated to `develop`).

So the chain is: merge this → unblock the PR step (org toggle or `PERF_BASELINE_TOKEN`) → dispatch Performance on develop → merge the baseline PR it opens → `perf-nightly` goes green. Nothing else is in the way.

gregpriday:
## Performance workflow is green

[Run 30915183221](https://github.com/daintreehq/daintree/actions/runs/30915183221):

```
perf-ci:               success
perf-nightly:          success   [perf:nightly] PASS scenarios=63 failed=0
                                 [perf:soak]    PASS scenarios=8  failed=0
perf-update-baselines: skipped   (develop-only, correctly gated)
```

First green run since 2026-05-14, and the first time `perf-nightly` has completed at all — it previously refused to start.

## Two changes since the original description

**The org permission is no longer on the critical path.** Rather than wait on the toggle, the baselines were regenerated by the workflow itself on `ubuntu-22.04` and harvested from the new artifact ([run 30908805945](https://github.com/daintreehq/daintree/actions/runs/30908805945): all four regens plus `verify-baselines` green, only the PR step failing). They're committed here, so `perf-nightly` is green on merge rather than after a toggle-plus-baseline-PR dance. The artifact upload is permanent — it means a blocked PR step no longer discards an hour of runner time.

**`ABSOLUTE_REGRESSION_MS_FLOOR` 1ms → 3ms.** With the gate finally running on real numbers, `perf-nightly` tripped one marginal budget per run, a different one each time:

| Run | Failure |
|---|---|
| 30911282063 | PERF-062, 31.3% vs 20% regression budget on an ~11ms scenario |
| 30912452991 | PERF-001, +1.212ms vs the 1ms absolute floor |

PERF-001 is `critical` and measures 1.096–2.932ms across runs — a spread wider than the 1ms tolerance itself, so pass/fail depended on where in that band the baseline happened to be captured (the 2026-02-11 baseline caught its high end and always passed; a regenerated 1.36ms caught the low end and failed next run). That's a coin flip, not a signal. 3ms is ~2x the widest observed delta; the other two critical scenarios sit at 0.3ms and 0.7ms, so a genuine regression in any of them is an order-of-magnitude blowup and still trips it.

PERF-062 passed on the subsequent runs — worth watching, since a 20% budget on an ~11ms scenario is tight, but not worth loosening on one sample.

Issue #11790, opened 2026-08-15
A frozen or evicted background project view strands MCP dispatch to a bound session


gregpriday:
Once an external MCP session can be bound to a workspace, the bound project's view is usually not the active one. That's the whole point: the agent drives project A in the background while the user works in project B. But a backgrounded project view is not merely idle, and dispatch into one does not reliably work.

Two separate problems.

**Frozen views can't service the dispatch IPC.** CPU throttling is fine on its own. `webContentsLifecycle.ts:13-16` notes that `Emulation.setCPUThrottlingRate` is per-target and leaves the event loop, IPC and MessagePort dispatch running, so a throttled view still answers, just slower. The freeze is the problem. Under the efficiency resource profile we also CDP freeze cached views via `Page.setWebLifecycleState` (`ProjectViewLifecycleController.ts:156-162`), and a frozen renderer cannot run JS, so `MCP_SERVER_DISPATCH_ACTION_REQUEST` queues in Mojo and nothing answers it. `ProjectViewManager.ts:841` has the note from the last time this bit us, when queued `agent:state-changed` events stranded a background dashboard on a stale state. Chromium does not auto-resume a frozen renderer on focus either (`webContentsLifecycle.ts:18-21`), so nothing rescues it.

The failure mode is nasty because it is conditional. A bound session works fine, then starts timing out once the machine drifts into Efficiency and the bound project has no active agent. Worse, it doesn't even fail cleanly: the renderer's own 28s confirm timer never runs, so main's 30s bridge timer fires instead (`rendererBridge.ts:242`) and the caller gets a generic `EXECUTION_ERROR` rather than anything that explains itself.

The precedent for the fix already exists. `unfreezeActiveAgentViews()` (`ProjectViewManager.ts:860`) thaws a cached background view without activating it, and `freezeAllCached()` (`:838`) skips any project with a live agent for exactly this reason. A live session binding wants to be another reason to stay thawed.

**Eviction destroys bound views.** `ProjectViewEvictionController.ts:275` hard-protects views with a live assistant backend and nothing else, so memory pressure will evict a bound project's view out from under an active session. Every later call then fails `SESSION_BINDING_GONE`, which is at least honest, but it makes a bound session unreliable for exactly the long-running background work it exists to serve.

[...]

Issue #11789, opened 2026-08-15
Bind external MCP sessions to a workspace, and split routing from session origin


gregpriday:
An external MCP session should be bindable to one workspace. Right now it resolves its target project by window focus order at call time, so two external agents driving two projects retarget each other: focus window B and agent A's next call lands in B. The `org.daintree/resolved-workspace` metadata added in #11564 reports where a call landed, but only after it ran, which is no protection for a mutation.

Reported in #11788 with a repro and a curl prototype.

The shape of the fix: read a workspace selector once, while creating a new session, resolve it to a project, store it on the session, and route every later call there or fail. No fallback to another window, ever. `handleStreamableHttpRequest()` already holds the raw request at `httpLifecycle.ts:1094` and the new-session branch starts at `:1120`, right next to where the `Authorization` header is read, so the hook is already there. Nothing reads a selector today.

Two constraints that shape the implementation.

**`sessionWebContentsMap` cannot simply be reused.** Presence in that map is also what makes a session eligible for renderer-driven privilege changes. `issueGrant()` (`httpLifecycle.ts:1618`) and `setSessionTier()` (`:1550`) both reject on `pinnedWcId === undefined`, and the comment above the latter says the intent plainly: only help-session bearers should be elevated through that surface. Drop an external session into the map and it becomes eligible. `setSessionTier` happens to be saved by the rank ordering at `:1559` (`external` sits top, so the move reads as a refused downgrade), but `issueGrant` has no such floor. Its only check is `minimumPermittingTier(toolId) !== null` (`:1631`), and the call gate honours a grant over failed tier membership. That is a path to an external client holding a grant for a tool outside `MCP_EXTERNAL_TIER_TOOLS`.

So the session record needs an explicit origin rather than an inferred one, something like `{ origin: "help" | "assistant-pane" | "external", webContentsId, workspaceId, context? }`. Routing reads `webContentsId`. Authorization, notifications and inventory read `origin`. `notifyTierMismatch` (`:1326`) and `notifyToolCallStarted` (`:1373`) need the same treatment: they currently no-op for external sessions by accident, and `HelpPanel` is mounted unconditionally in `AppLayout.tsx`, so those events would not land in a void.

[...]

Issue #11788, opened 2026-08-14
Bind external MCP sessions to an explicit workspace for concurrent multi-project orchestration


jobordu:
## Problem

External API-key MCP sessions are not bound to a Daintree workspace. Every renderer-backed tool call resolves the most-recently-focused live project window at call time.

That makes concurrent external agents across projects unsafe: changing focus for agent B can retarget agent A's next call into B's project. The `org.daintree/resolved-workspace` metadata reports where an action landed only **after** execution, so it cannot protect mutations.

Reproduced with Daintree 0.31.0 / current `develop`:

1. Open projects A and B in separate Daintree windows.
2. Connect an external API-key MCP client from each project.
3. Focus B.
4. Call `actions.getContext`, `terminal.list`, or `worktree.list` from A.
5. A receives B's state. Focus A and repeat from B; the target changes again.

The caller's cwd, project-local MCP config, MCP session ID, and `clientInfo` do not scope routing.

## Source evidence

- `rendererBridge.getActiveProjectWebContents()` selects `WindowRegistry.focusOrder()[0]`.
- `HttpLifecycle.resolvePinnedWebContentsId()` only pins help-session/assistant-pane bearers.
- `SessionStore.sessionWebContentsMap` explicitly leaves API-key sessions absent.
- `buildSessionServerDeps()` sends every unpinned external call through shared focus-based dispatch.
- `MCP_EXTERNAL_TIER_TOOLS` excludes project discovery/switching.
- #11564 intentionally retained unpinned external sessions and added focus-order routing plus post-dispatch observability.

## Required outcome

An external client can establish a session explicitly bound to one workspace, and preferably an initial worktree context. Multiple sessions bound to different projects can operate concurrently without focus changes retargeting either.

**The bound workspace must be driven in the background. Binding or dispatching must not visibly switch projects, focus a Daintree window, attach the background view, or change the user's active worktree.** The user must be free to watch or interact with any other project while agents continue.

A reasonable transport contract is a project-specific generated endpoint or header, for example:

```
http://127.0.0.1:<port>/mcp?workspaceId=<id>
```

or a static `X-Daintree-Workspace-Id` header. The binding can reuse the pinned `WebContents` and `ActionContext` machinery already used by help sessions, but must resolve the target without calling `project.switch`. If a background renderer must be created or temporarily unfrozen, that operation must remain headless and must not disturb visible focus.

This is a contract suggestion, not a requirement to use query parameters specifically.

[...]

Issue #11536, opened 2026-07-31
External tool calls route to the first registered window


gregpriday:
An unpinned external MCP session resolves its target renderer through `getActiveProjectWebContents()`, which iterates `registry.all()`. That is Map insertion order, not focus order.

So an API key client drives whichever window registered first, regardless of what the user is looking at.

`WindowRegistry.focusOrder()` already exists and has no call sites anywhere in `electron/`.

There is a second, related gap. Nothing scopes an external session to a project. A caller does a series of tool calls, the user switches project midway, and every un-scoped call after that silently retargets. For an agent doing long running work that is a correctness problem, not a UX one, because the results look normal.

Stamping the resolved project id and path onto the tool result envelope would at least make the drift observable.

Evidence: `electron/services/mcp-server/rendererBridge.ts`, `electron/services/mcp-server/httpLifecycle.ts`.

PR #11842, opened 2026-08-17
fix(mcp): capture the window registry when the server is enabled at runtime

jobordu:
## Summary

Found while writing the two-window E2E probe #11788 asks for. The binding work in #11789/#11790 holds up — every criterion in #11788 that a two-window test can check passes on `develop` unmodified. What doesn't hold up is the path running beside it.

`McpServerService.setEnabled(true)` hands `getWindowRegistry()` straight to `httpLifecycle.start()` and never assigns `this._registry`. Only `start()` does that, and the deferred `mcp-server` boot task (`globalServicesInit.ts:1024`) returns early when the server is disabled — which is the default. So turning MCP on in Settings, the normal first-setup flow, leaves `_registry` null for the rest of the process.

The renderer bridge reads that field through the `getRegistry()` callback it was constructed with (`McpServerService.ts:198`). A null one silently degrades every registry-aware path to the process-global `getProjectViewManager()` — whichever window was created last. No error, no log.

## What breaks

**Unbound sessions never follow focus.** `getActiveProjectWebContents()` skips its `focusOrder()` branch entirely and falls through to the global, so an unbound external session pins to the newest window regardless of what the user is looking at. #11564's focus-order fix is dead code under this state. Worse, when that window closes the session fails with `"No Daintree window is open, so the action surface is unavailable"` while other windows are open and working.

**The resolved-workspace stamp only works for the newest window.** `resolveDispatchedWorkspace()` can only identify views owned by the global manager, so `org.daintree/resolved-workspace` (#11536) is absent everywhere else — and so are `kind`/`workspacePath` in the `org.daintree/workspace-binding` initialize capability (#11789), the fields a client reads to confirm which project it bound to before issuing a mutation.

Not affected: single-window installs (the global points at the only window), any session after the Daintree Assistant has run (`HelpSessionService.ts:1623` calls `start()`), and workspace-bound sessions — those route through `webContentsRegistry.getWebContentsForProject()`, a different registry. That last one is why this survived #11789 and #11790: binding is the one path that was already correct.

## Evidence

Two windows, project A opened from the welcome screen, project B via `window.openNew(path)`, MCP enabled from Settings. An unbound "canary" session runs alongside two bound ones. OS focus verified via `BrowserWindow.getFocusedWindow()` at every step.

```
focused window 2 -> canary landed in B
focused window 1 -> canary landed in B     <- focus moved, routing didn't
focused window 2 -> canary landed in B
focused window 1 -> canary landed in B
```

Stamps, before and after opening a third window:

```
before window C:  A=MISSING  B=B
after  window C:  A=MISSING  B=MISSING  C=C
```

B loses its stamp the moment C exists — the signature of a process-global that the newest window overwrites, not of per-window state.

Discriminator, with window B closed and window A the only window left:

```json
{"code":"EXECUTION_ERROR","message":"No Daintree window is open, so the action surface is unavailable. Retry once a project window is open.","retriable":true}
```

A bound session answered from window A correctly one call later, so the view was alive the whole time.

Control arm — same spec, same two windows, only difference is `mcpServer.enabled: true` in `config.json` before boot so the deferred task runs `start(registry)`:

```
focused 2  canary->B  boundA->A  boundB->B   all stamped
focused 1  canary->A  boundA->A  boundB->B   all stamped
focused 2  canary->B  boundA->A  boundB->B   all stamped
focused 1  canary->A  boundA->A  boundB->B   all stamped
```

Both symptoms vanish when the registry is captured. That's the diagnosis, not an inference from reading the code.

## The fix

One assignment: store the registry instead of only passing it along. Reaching `start()` from here would also work; this keeps the existing lifecycle call shape.

## Tests

New `e2e/full/resilience/mcp-workspace-binding.spec.ts` — two windows, two API-key sessions, raw JSON-RPC against the live Streamable HTTP endpoint (no SDK client, so the wire contract is what's under test):

1. bound sessions stay on their own workspace while focus moves — 36 interleaved calls, four verified focus flips, identity read from the payload rather than the server's own label
2. driving a background workspace changes nothing the user can see — window focus, each window's attached project, each view's `aria-current` worktree row
3. confirm-gated tools withheld from a bound session — absent from `tools/list`, direct call refused with `CONFIRMATION_REQUIRED` / `confirmationChannel: "workspace-bound"`
4. handshake refuses unknown and mismatched selectors without minting a session
5. an unbound session follows focus (the documented pre-binding behaviour)
6. every bound result carries the resolved-workspace stamp
7. the stamp does not depend on which window was created last
8. losing the bound view fails that session closed and leaves the other running

Tests 5–7 fail on the parent commit and pass with the fix. 1–4 and 8 pass either way — they're the #11788 criteria, and they're the reason I'm confident the binding feature itself is sound.

Ran locally on macOS: 8/8 E2E green, `McpServerService.*` 160 tests and `electron/services/mcp-server/` 973 tests all passing, clean tsc / eslint / prettier.

**Not covered:** the #11790 freeze and eviction paths. Forcing the efficiency profile or real memory pressure needs a main-process seam the E2E harness doesn't expose today, so `_forceProfileForTesting` and the eviction lease are still untested end-to-end. Worth its own follow-up.

https://claude.ai/code/session_01DVcR1KZbgvPGW5qMEM4yUR

### DIFF
--- a/e2e/full/resilience/mcp-workspace-binding.spec.ts
+++ b/e2e/full/resilience/mcp-workspace-binding.spec.ts
@@ -66,10 +66,11 @@ async function post(
   extraHeaders: Record<string, string> = {},
   query = ""
 ): Promise<JsonRpcResponse> {
-  // A deadline of its own: a stalled call would otherwise outlive the
-  // `expect.poll` deadlines that wrap these, because a callback that never
-  // returns is never re-polled — the run hangs to the suite timeout instead of
-  // failing where the stall happened.
+  // A deadline of its own, so a call that never comes back fails as itself
+  // rather than as an orphaned request outliving the test that made it.
+  // Deliberately looser than the poll deadlines wrapping some of these: when a
+  // stall could trip either timer, the poll timeout names the assertion that
+  // was waiting, so it is the one that should fire first.
   const res = await fetch(`http://127.0.0.1:${endpoint.port}/mcp${query}`, {
     method: "POST",
     headers: {
@@ -79,7 +80,7 @@ async function post(
       ...extraHeaders,
     },
     body: JSON.stringify(body),
-    signal: AbortSignal.timeout(T_LONG),
+    signal: AbortSignal.timeout(T_LONG * 2),
   });
   const raw = await res.text();
   return {
@@ -410,20 +411,26 @@ test.describe.serial("MCP: external sessions bound to a workspace (#11788)", () 
     const b = await initializeSession(endpoint, { workspaceId: workspaceB, clientName: "bg-B" });
     const rounds = 5;
     const tools = ["actions.getContext", "worktree.list", "terminal.list"];
-    // `terminal.list` is the one identity-free payload (both projects are idle),
-    // so it is the only tool that contributes no landing evidence.
-    const identityBearingTools = tools.length - 1;
-    const landings: Array<string | null> = [];
+    const landings: Array<{ tool: string; landedIn: string | null }> = [];
     for (let i = 0; i < rounds; i++) {
       for (const tool of tools) {
         const landedIn = payloadWorkspaceId(await callTool(endpoint, b.sessionId, tool));
         expect(landedIn === null || landedIn === workspaceB).toBe(true);
-        landings.push(landedIn);
+        landings.push({ tool, landedIn });
       }
     }
     // Without this, a run where every call errored out would land `null` across
     // the board and sail through the assertion above having proved nothing.
-    expect(landings.filter((id) => id !== null).length).toBe(rounds * identityBearingTools);
+    // Per tool rather than a total: a tool that stopped reporting identity is
+    // caught either way, but one transient error no longer fails the run.
+    // `terminal.list` is excluded because its payload carries no identity at all
+    // while both projects are idle, which is the expected state here.
+    for (const tool of tools.filter((t) => t !== "terminal.list")) {
+      expect(
+        landings.filter((l) => l.tool === tool && l.landedIn !== null).length,
+        `${tool} never reported a workspace id in ${rounds} rounds`
+      ).toBeGreaterThan(0);
+    }
 
     const after = await captureVisibleState(ctx.app, { pageA, pageB });
     expect(after).toEqual(before);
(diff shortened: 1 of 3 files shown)


gregpriday (approved):
Thanks for the diagnosis and the fix here, @jobordu. The two-window repro in the PR description made this easy to verify. This is my part of the contribution.

I independently verified your production fix is correct and complete. It's a one-line change in `McpServerService.setEnabled`, delegating to `start()` so `this._registry` actually gets assigned. I audited every `_registry` call site and confirmed `start()` is the sole assigner and every path routes through it. I didn't touch any production code.

The gap I closed was test coverage. This PR shipped no PR-gating regression proof. The only existing unit test on this path, `McpServerService.lifecycleAndElicitation.test.ts` around line 471, is pre-existing on `develop` and only asserts `isRunning` and `currentPort > 0`. Both were already true before the fix, because the old code still called `httpLifecycle.start(registry)`, so that test passed with the bug live. The only regression proof in the PR was the E2E spec, and this repo's CI policy runs unit tests only on PRs, no E2E, so nothing in this PR's CI actually proved the fix or would catch a regression of it.

I added a falsifiable unit test in `McpServerService.authAndToolSurface.test.ts`. It enables the server at runtime with the registry focused on window A while the process-global fallback points at window B, then asserts routing follows focus, first A then B, on both the dispatch payload and the resolved-workspace stamp. It's mutation-verified: reverting the one-line fix collapses the trace to `B -> B` and inverts the first assertion. I kept the existing `start()`-path focus test alongside it rather than converting it, since it asserts the pre-binding contract through the binding code path.

I also hardened three real holes in the E2E spec without restructuring it:

- A request deadline on the `post()` helper, so a stalled call fails as itself instead of leaving an orphaned request.
- A non-null assertion on the third window's workspace id. Without it, the test silently drops the binding header and turns a bound case into an unbound one that passes for the wrong reason.
- A per-tool non-null landing guard in the background-drive test, which previously would have passed vacuously if every call errored and returned null.

Local verification: 48 tests pass across both MCP unit test files, ESLint and Prettier are clean on both changed files, and a trial merge against `develop` is conflict-free.

Approving. Thanks again for tracking this down.

PR #11848, opened 2026-08-17
test(mcp): cover the #11790 bound-view eviction protection end to end

### LINKED ISSUE #11847: Add E2E coverage for the #11790 eviction protection on a bound MCP view
## Summary

#11790 / PR #11796 shipped the eviction protection for a background view backing a live MCP session with no end-to-end coverage:

> No E2E was run locally. The existing LRU eviction spec has no MCP session to bind, so it exercises none of this.

Unlike the freeze half (#11846, which cannot be tested from Playwright at all), this one **is** testable and discriminating. I have a working test and would like to know whether you want it before I open a PR.

Measured against `c1e0f2f75`, macOS local.

## What it asserts

`ProjectViewEvictionController.ts:330` — `const boundMcpSession = mcp.liveBinding || mcp.unknown` — runs through `evictStaleViews`, whose observable is membership of the `pvm.views` map. That is a real effect Playwright cannot distort, which is precisely why this half works where the freeze half does not.

A within-run A/B on a single app instance, three projects in one window:

- **CONTROL** (no binding): A is LRU-oldest, A is evicted. Pure LRU, the pre-#11796 behaviour.
- **BOUND** (session bound to A): A is still LRU-oldest, but B dies and A survives. LRU inverted.

Also confirmed: a real MCP session bound over HTTP to a **cached** workspace flips `mcpActivityFor(id, wc).liveBinding` to `true` from the `initialize` handshake alone, with no tool call needed.

## Negative control — reproduced first-hand

Temporarily changing that line to `const boundMcpSession = false && (mcp.liveBinding || mcp.unknown);` and rebuilding:

```
guard disabled:  CONTROL ✓ (4.3s)   BOUND ✘ (4.5s)
guard restored:  CONTROL ✓ (4.2s)   BOUND ✓ (4.6s)
```

BOUND fails at exactly the intended assertion while CONTROL keeps passing, so the failure is specific to the protection rather than a broken harness. Source reverted, `git diff` clean.

## Stability

Two full runs of the pair: ~4.2–5.2s per test, **21s wall per run** including cold Electron launch. Five runs during development were 5/5 green. No retries, no flake observed on macOS local.

## Proposed placement

`e2e/full/resilience/` as a **new file** — it needs one window with three projects, which does not fit `mcp-workspace-binding.spec.ts`'s two-window arrangement.

Not `core`: this is an internal memory-policy interaction rather than a release-gating user flow, and `core` runs with `FAIL_ON_FLAKY_TESTS`. Not `e2e/nightly`: that bucket is memory-leak detection on a 600s budget. Worth stating plainly that PR CI runs no E2E, so this buys nightly regression detection only.

## Questions before I send anything

1. Do you want it?
2. Would you rather write it yourself? `core-lru-project-eviction.spec.ts` already has the `configurePvm` helper, so this may be faster for you than reviewing someone else's.
3. If you want it from me, is `e2e/full/resilience/` the placement you'd pick?

## Gaps in what I checked

- Covers `liveBinding` (last-tier deprioritization) only. The `dispatchLease` hard exclusion needs an in-flight request held open across an eviction pass; I have not established whether that is drivable.
- macOS local only. Timing on a loaded Ubuntu runner is unverified.

jobordu:
Closes #11847.

E2E coverage for the #11790 eviction protection: a background view backing a live MCP session must survive an eviction pass that pure LRU would have taken it in.

## What it asserts

`ProjectViewEvictionController.ts:330` (`boundMcpSession = mcp.liveBinding || mcp.unknown`) runs through `evictStaleViews`, whose observable is membership of the `pvm.views` map. A within-run A/B on one app instance, three projects in one window:

- **CONTROL** (nothing bound): A is LRU-oldest, A is evicted. Pure LRU, the pre-#11790 behaviour.
- **BOUND** (session bound to A): A is still LRU-oldest, but B dies and A survives. LRU inverted.

The binding is a real MCP session over the HTTP endpoint, not a stub — `liveBinding` flips from the `initialize` handshake alone, with no tool call needed.

## Both corrections from the issue are in

1. **Forced tier-2 reclaim is stubbed, not tolerated.** `setLowMemoryFreeThresholdMb(null)` closes the periodic sampler path but not the forced pass — `criticalPressure` is just `forcePressure` and never consults the policy (`:132`), so a forced pass would still set `effectiveMax = 1` and evict A. The spec overrides `reclaimCachedViewsUnderPressure()` on the PVM rather than loosening the assertion, since loosening it would mask the regression this test exists to catch.
2. **The threshold is re-asserted immediately before the eviction trigger**, because `ResourceProfileService.applyCurrentProfileTo()` re-arms the policy from a deferred task (`:455`) and can otherwise overwrite a `configurePvm` that landed first.

`dispatchLease` is deliberately out of scope, per the issue.

## Verification

Both directions, on this branch:

```
positive   run 1  CONTROL ✓ 4.3s   BOUND ✓ 4.5s   (21.0s wall)
           run 2  CONTROL ✓ 4.2s   BOUND ✓ 4.6s   (20.2s wall)
           run 3  CONTROL ✓ 4.3s   BOUND ✓ 4.5s   (20.3s wall)

negative   guard at :330 flipped to `false && (...)`, rebuilt:
           CONTROL ✓ 4.3s   BOUND ✘ 4.5s          (source reverted, git diff clean)
```

BOUND fails at exactly the intended assertion while CONTROL keeps passing, so the failure is specific to the protection rather than a broken harness.

`npm run check` clean — typecheck, lint ratchet (no new warnings against the 1091 baseline), format.

## Limits

macOS local only; timing on a loaded Ubuntu runner is unverified.

**Correction to the issue text:** I wrote there that this buys "nightly regression detection only". That is wrong — the test-nightly was retired (`docs/e2e-testing.md:183`), and `e2e.yml` is invoked by `release-macos.yml` / `release-linux.yml` / `release-windows.yml`, so `full-resilience` gates releases on all three platforms. This spec is therefore release-gating, which raises the flakiness bar on it rather than lowering it. Placement still looks right to me and matches your call, but you should weigh it knowing that.

### DIFF
--- a/e2e/full/resilience/mcp-bound-view-eviction.spec.ts
+++ b/e2e/full/resilience/mcp-bound-view-eviction.spec.ts
@@ -5,6 +5,7 @@ import { createFixtureRepos } from "../../helpers/fixtures";
 import { openAndOnboardProject } from "../../helpers/project";
 import { addAndSwitchToProject, selectExistingProjectAndRefresh } from "../../helpers/workflows";
 import { T_LONG } from "../../helpers/timeouts";
+import { ACTIVE_AGENT_STATES } from "../../../shared/types/agent";
 
 /**
  * #11790 moved a cached view backing a live MCP session binding into the LAST
@@ -47,9 +48,15 @@ const PROJECT_A = "project-A";
 const PROJECT_B = "project-B";
 const PROJECT_C = "project-C";
 
-/** Three cached views, then one, so exactly one eviction decision is observed. */
+/**
+ * Three managed views — A and B cached, C active, since the active view stays
+ * in the map — then a limit one lower, so exactly one eviction is observed.
+ */
 const CACHE_LIMIT_ARRANGED = 3;
 const CACHE_LIMIT_TRIGGER = 2;
+
+/** Passed into the main process: `hasActiveAgent`'s own set, not a copy of it. */
+const ACTIVE_STATES: string[] = Array.from(ACTIVE_AGENT_STATES);
 
 interface McpEndpoint {
   port: number;
@@ -156,6 +163,17 @@ interface PvmViewProbe {
   dispatchLease: boolean;
   /** The activity callback threw — deprioritizes like a binding, so it is a premise. */
   unknown: boolean;
+  /**
+   * `hasActiveAgent` for this project, recomputed from the manager's own
+   * terminal maps. It ranks a view behind the unprotected tier, so an active
+   * agent on A would spare A without the binding doing any of the work.
+   */
+  activeAgent: boolean;
+  /**
+   * A registered assistant backend — the one HARD floor, checked before the
+   * bound-session tier, so it would remove A from the candidate list outright.
+   */
+  assistantBackend: boolean;
 }
 
 interface PvmProbe {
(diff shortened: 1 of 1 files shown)


gregpriday (approved):
Thanks for this, @jobordu. Solid Playwright spec for the #11790 bound-view eviction protection: a real MCP session over HTTP rather than a stub, and a clean CONTROL/BOUND A/B structure.

I pushed two commits on top to tighten a few assertions: made the BOUND case's pass condition attributable specifically to the binding rather than other eviction protections that could have masked it, swapped a polling check for a direct synchronous before/after membership capture, dropped a tautological assertion, stubbed a second eviction path that could race with the test, and added `test.slow()` to match the sibling spec's timeout handling.

Ran lint, format, and typecheck clean, and the spec passes repeatedly locally. Approving.

Issue #11846, opened 2026-08-17
Playwright's focus emulation makes CDP freeze a no-op, so #11790's freeze guard cannot be E2E-tested


jobordu:
## Summary

#11790 / PR #11796 shipped with a noted coverage gap:

> No E2E was run locally. The existing LRU eviction spec has no MCP session to bind, so it exercises none of this.

I went to close it and found the freeze half **cannot** be covered from Playwright — not "flaky", not "hard", structurally inert. Filing the measurement so nobody else spends a day rediscovering it.

The eviction half *is* testable and is split out to #11847.

Measured against `c1e0f2f75`, macOS local.

## A Playwright-driven freeze test can never fail

Under Playwright, a cached view with `setEfficiencyFreeze(true)` past the debounce keeps running JavaScript. Counting `MessageChannel` self-post tasks over 3s inside the app:

| scenario | tasks executed |
|---|---|
| cached, not frozen (control) | 396,724 |
| cached, "frozen" | 430,697 |

`document.visibilityState` reads `visible` and `document.hasFocus()` is `true`, even though `view.getVisible() === false` and the view is detached from `contentView`.

## Root cause

Isolated in a standalone Electron app with no Playwright present, mirroring `deactivateEntry` (attach a `WebContentsView`, detach, `setVisible(false)`, CDP-freeze):

| phase | `visibilityState` | tasks in 3s |
|---|---|---|
| control, not frozen | `hidden` | 113,606 |
| **frozen** | `hidden` | **0** |
| recovered | `hidden` | 109,647 |
| after `Emulation.setFocusEmulationEnabled{enabled:true}` | **`visible`** | — |
| **frozen again, focus emulation on** | `visible` | **1,938,407** |

The only variable between "freeze works perfectly" and "freeze is completely inert" is `Emulation.setFocusEmulationEnabled`, which Playwright sends to every page target it attaches to. `Page.setWebLifecycleState(frozen)` calls `WasHidden()` internally, and a forced-visible page cannot be hidden — so the freeze silently no-ops while still returning `ok`.

**CDP freeze is not broken in production. It works flawlessly. It is unreachable from a Playwright test, in every project view, always.**

## The positive control is what settles it

In the exact scenario a test author would write — one window, project A bound to a live MCP session and cached, project B cached and unbound, efficiency freeze on:

- bound A: 262,222 tasks, `visible`
- unbound B (must freeze for any assertion to mean anything): 255,016 tasks, `visible`

[...]

Issue #11840, opened 2026-08-17
Moving an agent to another worktree leaves it in the old directory


gregpriday:
Dragging a running agent terminal onto a different worktree row re-files the panel, but the process keeps running in its original directory. Nothing in the UI says so.

Justin hit this on Codex. He moved a session out of the main repo onto a worktree row, kept working, and made 20 commits to `main` believing they were landing on a branch. In his words: "the worktree move had moved the panel, but didn't adjust the PWD, so I was merrily working on main the whole time."

## What actually happens

[`moveTerminalToWorktree`](https://github.com/daintreehq/daintree/blob/284b7c579e4860630f0c1370b362824c6e04f858/src/store/slices/panelRegistry/restart.ts#L923-L1005) rewrites `panel.worktreeId` and `panel.location`. That is all it does. It never touches `panel.cwd`, never restarts the PTY, and never signals the agent. The panel ends up filed under a worktree its process has never been in.

Three things make this worse than a stale label:

1. The move records itself in `agentLifecycleLedger` as `"explicit"` attribution [specifically so that later cwd-based inference can never overwrite it](https://github.com/daintreehq/daintree/blob/284b7c579e4860630f0c1370b362824c6e04f858/src/store/slices/panelRegistry/restart.ts#L956-L961). The one mechanism that could have self-corrected is deliberately switched off.
2. [Grouped panels take a different branch](https://github.com/daintreehq/daintree/blob/284b7c579e4860630f0c1370b362824c6e04f858/src/store/slices/panelRegistry/restart.ts#L934-L951) and `return` before that ledger write ever runs. One drag can divert several agents at once, and none of them get the attribution record.
3. [`terminal.moveToWorktree`](https://github.com/daintreehq/daintree/blob/284b7c579e4860630f0c1370b362824c6e04f858/src/services/actions/definitions/terminalSpawnActions.ts#L244-L270) is marked `danger: "safe"` with no confirmation.

The sibling feature gets it right. [`moveToNewWorktreeAndTransfer`](https://github.com/daintreehq/daintree/blob/284b7c579e4860630f0c1370b362824c6e04f858/src/store/slices/panelRegistry/restart.ts#L1007-L1085) (#4773) captures the terminal buffer, sets the new cwd, restarts the PTY there, and re-injects the buffer as a first prompt. Same gesture family, opposite semantics. Nobody wired the existing-worktree path to it.

## Why the obvious fixes don't work

[...]

PR #11849, opened 2026-08-17
test(projectview): Playwright-free harness for CDP renderer freeze

jobordu:
Refs #11846. Does not close it — that issue stays open for the `webContentsLifecycle.ts` comment.

The Playwright-free freeze harness you invited. It boots the real app and measures whether CDP freeze actually stops the renderer, which #11846 established Playwright structurally cannot do.

## Your four requirements

1. **Real app, not a mirror.** A `--daintree-freeze-harness` flag boots through the real `main.ts` / `bootstrap.ts` / `setupWindowServices`, then drives the production path: `ProjectViewManager.switchTo` → `deactivateEntry` → `setEfficiencyFreeze` → `freezeAllCached`. Nothing stubbed.
2. **No debugger client on the measured view** — satisfied, but by a different means than you suggested, and I'd like your call on it. You proposed an IPC channel; this uses `webContents.executeJavaScript`, which is Blink script execution rather than CDP, so it never calls `debugger.attach` and cannot collide with `ensureAttached`. Same shape as your suggestion — counter ticks in the renderer, main reads it — but it adds zero entries to `channels.ts`, zero `IpcEventMap` entries and zero preload surface, which seemed worth preserving given `check:preload-backdoors` and the recent tool-surface shrink. Say the word and I'll swap it for a real channel.
3. **Ratio, not a bound.** `MIN_FREEZE_RATIO=100`, `MIN_RECOVERY_RATIO=100`, `MIN_RECOVERY_FRACTION_OF_CONTROL=0.1`. Stall duration is logged, never asserted. The single absolute is `MIN_CONTROL_TICKS=1000`, a liveness floor on the positive control only — without it a dead probe reads 0/0/0 and every ratio passes vacuously.
4. **Recovery leg.** Two assertions: recovered-vs-frozen ratio, and recovered ≥ 10% of control so a trickle doesn't pass. A hung renderer fails separately via a 15s read timeout with its own message.

## Verified in both directions

```
freeze intact      control =  51,340   frozen = 0   recovered = 50,283   ratio 51,340x   PASS
freeze neutered    control =  54,334   frozen = 58,495                   ratio     0.9x  FAIL (exit 1)
```

Against a 100x floor that is three orders of magnitude of headroom on either side. `frozen=0` and `frozenInterior=0` on every green run; control rate varies 41k–59k between runs, which is exactly why the assertions are ratios rather than thresholds.

## One correction to my own issue text

I claimed in #11846 that this would catch the `disable-renderer-backgrounding` regression from lesson #4683. **It does not, because that regression no longer breaks freeze.** Appending both switches in `main.ts`:

```
without switches   control =  51,340   frozen = 0   PASS
with both          control = 446,953   frozen = 0   PASS
```

The switches took effect — control rate up 8.7x — and freeze still zeroed the renderer. Details and the implication for the audit comment at `webContentsLifecycle.ts:20-27` are in a comment on #11846. Flagging it here because it changes what this harness is worth: it verifies freeze semantics, it does not guard that particular regression.

## Placement

`scripts/run-freeze-harness.mjs` + `electron/services/freezeHarness.ts` + a flag in `runtimeFlags.ts` + `npm run test:freeze-harness`. This mirrors `--smoke-test`, the repo's only other Playwright-free real-app-boot harness, including `run-packaged-smoke.mjs`'s convention of exported pure functions behind an `invokedDirectly` guard so the adjudication logic is unit-testable.

**Deliberately not wired into CI.** `docs/e2e-testing.md:183` records the test-nightly as retired, and only `nightly-publish.yml` still runs on a cron, so there was no obvious home. Left as run-on-demand rather than guessed at — tell me where you want it and I'll wire it.

## Limits and known risks

- **macOS only.** Windows unverified, and as you predicted it is the likeliest to differ. Stated as a limit in the runner header and the docs rather than implied as coverage.
- `ResourceProfileService` can call `setEfficiencyFreeze` underneath the harness on a profile transition. Both directions fail *safe* — red, never a false green — but it is a genuine flake source on a memory-pressured box. Not observed across 13 runs.
- Cached views get a periodic CDP memory purge 20s after caching; a run completes in ~14s. Raising `DAINTREE_FREEZE_HARNESS_WINDOW_MS` past ~4,500 would cross that.
- `MIN_CONTROL_TICKS=1000` sits 40x below the slowest observed control, but is untested on a slow CI box.

## Checks

`npm run check` clean — typecheck, lint ratchet at the 1091 baseline, format. Full unit suite 49,754 passing. Mutation check: forcing the verdict to always-pass kills 6 tests.

### DIFF
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -224,6 +224,8 @@ npm run build && npm run test:freeze-harness
 FREEZE_HARNESS_RUNS=5 npm run test:freeze-harness   # variance
 ```
 
+The whole measurement has to finish before the cached view's first memory purge (`CACHED_VIEW_PURGE_DELAY_MS`, armed the moment the view is cached), or the purge perturbs the throughput being measured. That is checked against the elapsed clock just before the control leg, so widening `DAINTREE_FREEZE_HARNESS_WINDOW_MS` past what fits fails the run with the shortfall rather than reporting a number measured across a purge.
+
 Reference numbers (macOS, Electron 42, 3s windows): control ~54,000 ticks, frozen **0**, recovered ~52,000. With `freezeWebContents` neutered the same run reads control 54,026 / frozen 53,875 — a ratio of 1.0x against 54,000x, so the harness is discriminating by a wide margin.
 
 **Measured on macOS only.** The mechanism is Chromium/CDP semantics and should be platform-independent, but that is an inference; Windows is unverified and is the platform most likely to differ. The harness is not wired into any workflow yet — run it on demand.
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -209,6 +209,25 @@ Two distinct smoke checks run at different points in the pipeline:
 
 `npm run test:e2e:core` runs the 5 Playwright specs in `e2e/core/` against the Electron app. These are deterministic release-gate tests that gate every OS publish.
 
+## `test:freeze-harness` — why it can't be a Playwright spec
+
+`npm run test:freeze-harness` (`scripts/run-freeze-harness.mjs` + `electron/services/freezeHarness.ts`) measures whether a cached project view's renderer genuinely stops executing tasks when the efficiency-freeze path freezes it, and resumes when it is thawed.
+
+**Do not port this to Playwright.** Playwright sends `Emulation.setFocusEmulationEnabled` to every page target it attaches to. That handler takes out a `WebContents` capturer with `stay_hidden=false`, which permanently tells the renderer it is user-visible. `Page.setWebLifecycleState(frozen)` calls `WasHidden()` internally, so on a forced-visible page it no-ops — while still returning success. A freeze assertion written in Playwright passes whether or not freeze works, in every project view, always (#11846). A second CDP session can't undo it either: `capture_handle_` and `focus_emulation_enabled_` are per-session handler state.
+
+The harness therefore boots the real app with no Playwright and no debugger client of its own — attaching one would collide with the `ensureAttached` inside `freezeWebContents`, land in `EXPECTED_CDP_ERRORS`, and be swallowed. The renderer counts its own `MessageChannel` tasks into wall-clock buckets; `webContents.executeJavaScript` (Blink script execution, not CDP) reads the timeline out **after** the thaw, because a frozen renderer cannot answer while frozen. Three legs are measured at equal width — cached-unfrozen, frozen, thawed — and the assertions are **ratios**, not timing bounds, because a bound goes red on a loaded box.
+
+`MessageChannel` rather than a timer is deliberate: timers are subject to background throttling, which would confound "frozen" with "merely throttled".
+
+```bash
+npm run build && npm run test:freeze-harness
+FREEZE_HARNESS_RUNS=5 npm run test:freeze-harness   # variance
+```
+
+Reference numbers (macOS, Electron 42, 3s windows): control ~54,000 ticks, frozen **0**, recovered ~52,000. With `freezeWebContents` neutered the same run reads control 54,026 / frozen 53,875 — a ratio of 1.0x against 54,000x, so the harness is discriminating by a wide margin.
+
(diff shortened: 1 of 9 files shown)


gregpriday (approved):
Accepting this as is. The core design is right, and I checked the reasoning behind it myself rather than taking it on faith. Freeze suspends every freezable Blink task queue, MessagePort delivery included, so a `MessageChannel` is the correct tick source. A timer would have conflated "background-throttled" with actually frozen. And `executeJavaScript` against a frozen renderer hangs the promise indefinitely instead of rejecting, so reading the counter after thaw and the `withTimeout` wrapper around it aren't defensive padding. Both are load-bearing. Ratios over fixed bounds is the right call too, given how much the control rate varies from run to run.

On your two questions:

1. Keep `executeJavaScript`, don't move to an IPC channel. `check:preload-backdoors` scans for `DAINTREE_E2E_*` env vars and `__DAINTREE_E2E_*` globals, and `--daintree-freeze-harness` is a plain `process.argv` flag, so it was never in scope for that gate to begin with, same precedent as `--demo-mode`. A preload/IPC surface is a permanent cost on the product for a dev-only tool, and `executeJavaScript` is Blink script execution, not a second CDP session, which is exactly what keeps it from colliding with `ensureAttached` inside `freezeWebContents`.

2. Leave it unwired and run-on-demand for now. We retired the cron test runs on purpose, so if this ever does get wired in it belongs on the `stabilize.yml` on-demand surface, not `nightly-publish.yml`. Not a blocker either way.

One more thing before the diff: thanks for catching yourself on the `disable-renderer-backgrounding` regression from lesson #4683. Flagging that the harness doesn't actually catch it, against what the original issue claimed, is exactly the right instinct. That's worth more to me than a harness that quietly overclaims what it covers.

I pushed three commits on top of yours, all in the runner and the harness, nothing in app code:

- The runner reproduced, almost exactly, the pre-fix pattern behind the v0.32.0 Windows release-blocking hang in the sibling `run-packaged-smoke.mjs`: no explicit exit after flush on the success path, and a root-only kill. Ported `exitAfterFlush` plus a real process-tree teardown. I deliberately didn't port the `/im` image-name fallback that `run-packaged-smoke.mjs` uses. That runner is single-tenant against a packaged build. This one launches the unpackaged `electron.exe`, so killing by image name would take out VS Code and every other Electron app on a dev's machine. PID-scoped only.

- Bigger one: the runner was settling only on `close`. Electron's descendants inherit the runner's stdio pipes, so one survivor means `close` never fires. A run that printed every PASS marker and exited 0 would have sat until the 180s timeout and been reported as a failure. It now settles from `exit`, with a bounded drain for the log tail.

- `runFreezeHarness` could reject instead of returning: `mkdtemp` sat outside the `try`, and the final `rm` was unguarded. Either one would escape `setupWindowServices` with the window still up and nothing issuing an exit. `app.exit` now sits in a `finally` in `windowServices.ts`, so nothing between the harness returning and the exit can strand the process.

- The purge budget the `MEASURE_WINDOW_MS` docstring described was promised, not enforced. `CACHED_VIEW_PURGE_DELAY_MS` is now exported from `ProjectViewLifecycleController` instead of duplicated, and `evaluatePurgeBudget` checks the planned schedule before the control leg and the actual finish after the last window. `DAINTREE_FREEZE_HARNESS_WINDOW_MS=5000` used to run a measurement straight across a purge and just report the number.

- The harness now thaws explicitly before injecting the probe, so a run that starts under the efficiency profile measures a live control leg instead of injecting into an already-frozen renderer.

- Smaller stuff: `taskkill /f` only on escalation, `SIGINT`/`SIGTERM`/`SIGHUP` forwarded to the child's process group (detached spawning had put Electron outside the terminal's group), timeouts above Node's timer ceiling get rejected outright instead of silently clamped to ~1ms, and the child's captured output gets dumped on failure since only the marker lines are echoed live.

Two gaps I'm leaving alone rather than papering over. `runHarnessOnce`'s event interleavings still have no tests, covering them means dependency-injecting `spawn`, which felt disproportionate here. Worth it if this runner grows. And the harness's single-view shape structurally can't detect freeze bleeding into a sibling active view sharing a renderer process. Real follow-up, not something to retrofit into this PR.

Last thing: `develop` moved 7 commits since you opened this (the #11840 worktree-move work), but none of it overlaps and the merge comes out clean, so I left the branch un-rebased on purpose to keep this push a fast-forward.

PR #12672, opened 2026-09-23
Make the notification centre triageable at fleet volume

gregpriday:
## Summary

- The notification centre fell apart at realistic volume. With about forty entries across three projects, it couldn't tell you where a notification came from, where the new stuff ended, or why things had gone quiet.
- This is a design pass on the popover to fix that, plus a small follow-on that stops five other screens from showing worktree and project ids as names.

## What changed

**Where it came from.** Every row now has a quiet `project · worktree` line under the message. Grouped headers use the same names instead of a sha256 project id or a worktree path, keep the worktree visible when the project name is long, and stay pinned while you scroll a section. A project that's no longer registered reads "Another project", app-wide events (disk space, the pty host, plugin updates) read "Daintree", and nothing prints an id.

**What needs you.** The Needs attention rail is capped at three compact rows (title, source, recovery buttons, no body), so the first screen shows actual list rows instead of five pinned errors. An agent waiting for input now pins ahead of failures.

**What's new.** "New since you last looked" marked where new started and nothing marked where it stopped. There's now an "Earlier" boundary, and grouped headers show "N new". The Jump to new pill no longer sits on top of the divider it points at, retires when you scroll straight past it, and lands focus on the first new row instead of the label.

**Snooze.** Snoozed rows show when they come back as visible text. The picker shows the resolved wake time next to each option ("Until tomorrow" is 8:00 AM, and says so). `h` opens the durations directly with focus on the first one, and Escape puts you back on the row.

**Why it's quiet.** The quiet strip now shows real silences, not just mutes: kinds switched off that ship on, plus what this project or other projects in the inbox have silenced from a row's menu. It offers Manage. Completed ships off, so it doesn't count on its own, otherwise everyone would see the strip.

**Keyboard and a11y.** Marking a pinned row read with `u` no longer drops focus on `<body>`. Row controls are named for their row ("Dismiss Push rejected"). Grouped sections take their accessible name from the header. The last slash-alpha legacy tokens in the panel moved to semantic ones, and Resume, Manage and the divider's mark-read share one bordered button with a focus ring.

**Outside the panel.** ReEntrySummary now uses the same severity glyphs as the toaster and inbox. The dev server dashboard, the "Dev server stopped" toast, two sidebar announcements and the settings dialog no longer fall back to ids. The dashboard uses the Pilot's shortest-unique-path labels, now in `src/lib/worktreeLabels.ts`, so two checkouts called `main` still read apart. That's its own commit (`refactor(ui): stop showing worktree and project ids as names`) so it can come out on its own.

## Design review

Scored 6.2 → 8.0 → 8.6 → 9.5 over four rounds, against triage speed, information architecture, row craft, state coverage, keyboard and a11y, and consistency. The consistency survey found eight other places printing ids as names. Five were changed, and two were left on purpose (the recipe scope badge and a deleted worktree's record keep the full path).

There's a new opt-in capture harness, `e2e/screenshots/notification-center-volume-review.spec.ts`, that seeds a realistic multi-project inbox and captures 18 states.

## Testing

- `npm test`: everything passes except 12 tests in `packages/daintree-plugin`, which fail only in a worktree because it's missing that package's nested `commander` install. They pass 64/64 with it linked in, and this branch doesn't touch `packages/`.
- `npm run check`: passes, except `check:sample-views`, which reports a false stale in a worktree (two `//#region` path comments from the `node_modules` symlink). Every step after it passes when run individually.
- The new tests pin rules rather than values, and each one was checked by putting the bug back and watching it fail.
- The text-ramp manifest is regenerated for the six sites moved off the opacity ramp.


gregpriday:
The earlier red run was the vitest worker flake in shard 2: "Worker exited unexpectedly" after all 15,314 tests in the shard had passed, with only vitest internals in the stack. Every test file this branch touches passed in its shard. Reran the failed jobs with no code changes and everything's green, including `ci-ok`.

